PR #24, opened 2023-01-08
RTP API

### LINKED ISSUE #5: RTP level media API
The media API currently works only on "Sample" level, full H264/VP8/etc packets. The RTP level is an internal concern to str0m and there are built-in packetizer/depacketizer to go from sample level to RTP level. https://github.com/algesten/str0m/tree/main/packet

# The case against RTP level API

The WebRTC related RFC sometimes seem to consider SFU (Signal Forward Units) as something that just takes incoming RTP packets and routes them to other clients. This solves a quite narrow use case where not many peers participate in the same session.

RTP packets can't pass through an SFU unaltered because many RTP header fields depends on situation:

* Header extensions – negotiated in SDP, must be fulfilled specific to the peer-peer connection.
* Sequence numbers – must not be increased if you're not actively sending.
* TWCC (transport wide congestion control header) – must be set for entire peer.
* Bandwidth estimation / congestion control – must be handled for entire peer.
* PLI – if the sending peer has a bad connection, all downstream peers would shout PLI ("PLI storms" is a problem).
* Nacks and RTX resends – at the very least the original packets must be buffered to handle resends.

The above requires at least some level of rewriting the RTP header.

```
               Bad link                            ┌─────────────┐
                  │                                │             │
                  │                         ┌──────│  Sender 1   │
                  │                         │      │             │
┌─────────────┐   │    ┌─────────────┐      │      └─────────────┘
│             │   ▼    │             │      │                     
│  Receiver   │◀ ─ ─ ─ │     SFU     │◀─────┤                     
│             │        │             │      │                     
└─────────────┘        └─────────────┘      │      ┌─────────────┐
                                            │      │             │
                                            └──────│  Sender 2   │
                                                   │             │
                                                   └─────────────┘
```

The receiver here will miss different packets from Sender 1 and Sender 2. In a naive SFU, the NACK, PLI and FIR would be forwarded back to source. That means an increased pressure on bandwidth for all peers, not just the Receiver.

To make the SFU smarter, we can implement various levels of RTP buffering.

1. Buffer packets to handle NACK. This is very simple, just handle resends on the last leg (str0m does this internally).
2. Buffer packet sequences to handle some PLI/FIR without hitting the Sender. This requires a knowledge of what is a keyframe, which in turn requires us to reassemble RTP into Samples.

# Make an RTP level API

The above doesn't mean it's impossible to make an RTP level API. It just means the RTP level can't be thought of as "simply write this RTP packet to the wire". At the very least all headers must be fixed up to be correct given the SDP, TWCC etc.

I think we want a method to switch either an m-line, or the entire session into "RTP mode", which means instead of emitting `Event::MediaData(MediaData)`, we have another event such as `Event::RtpData(RtpData)`, and also  `MediaWriter::write_rtp` for the other direction.

algesten:
Close #5

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -177,8 +177,9 @@ struct NextPacket<'a> {
 #[derive(Debug)]
 struct ToPacketize {
     pt: Pt,
-    meta: PacketizedMeta,
     data: Vec<u8>,
+    meta: PacketizedMeta,
+    rtp_mode_header: Option<RtpHeader>,
 }
 
 enum NextPacketBody<'a> {
@@ -261,6 +262,7 @@ impl MediaInner {
         data: &[u8],
         rid: Option<Rid>,
         ext_vals: ExtensionValues,
+        rtp_mode_header: Option<RtpHeader>,
     ) -> Result<(), RtcError> {
         if !self.dir.is_sending() {
             return Err(RtcError::NotSendingDirection(self.dir));
@@ -269,6 +271,16 @@ impl MediaInner {
         let Some(spec) = self.params_by_pt(pt).map(|p| p.spec()) else {
             return Err(RtcError::UnknownPt(pt));
         };
+
+        // Implicitly if we have an rtp mode header, the codec must be Null.
+        let codec = if rtp_mode_header.is_some() {
+            Codec::Null
+        } else {
+            spec.codec
+        };
+
+        // Never want to check this for the null codec.
+        let is_audio = spec.codec.is_audio();
 
         // The SSRC is figured out given the simulcast level.
         let tx = get_source_tx(&mut self.sources_tx, rid, false).ok_or(RtcError::NoSenderSource)?;
@@ -279,8 +291,8 @@ impl MediaInner {
         // We don't actually want this buffer here, but it must be created before we
         // get to the do_packetize() (as part of handle_timeout).
         let _ = self.buffers_tx.entry(pt).or_insert_with(|| {
-            let max_retain = if spec.codec.is_audio() { 4096 } else { 2048 };
-            PacketizingBuffer::new(spec.codec.into(), max_retain)
+            let max_retain = if is_audio { 4096 } else { 2048 };
+            PacketizingBuffer::new(codec.into(), max_retain)
         });
 
         trace!(
@@ -298,8 +310,9 @@ impl MediaInner {
 
         self.to_packetize.push_back(ToPacketize {
             pt,
-            meta,
             data: data.to_vec(),
+            meta,
+            rtp_mode_header,
         });
 
         Ok(())
@@ -317,13 +330,52 @@ impl MediaInner {
                 .get_mut(&t.pt)
                 .expect("write() to create buffer");
 
-            if let Err(e) = buf.push_sample(now, &t.data, t.meta, MTU) {
-                return Err(RtcError::Packet(self.mid, t.pt, e));
-            };
+            if self.rtp_mode {
+                let rtp_header = t.rtp_mode_header.expect("rtp header in rtp mode");
+                buf.push_rtp_packet(now, t.data, t.meta, rtp_header);
+            } else {
+                if let Err(e) = buf.push_sample(now, &t.data, t.meta, MTU) {
+                    return Err(RtcError::Packet(self.mid, t.pt, e));
+                };
+            }
 
             // Invalidate cached queue_state.
             self.queue_state = None;
         }
+
+        Ok(())
+    }
+
+    pub(crate) fn write_rtp(
+        &mut self,
+        pt: Pt,
+        wallclock: Instant,
+        packet: &[u8],
+        exts: &ExtensionMap,
+    ) -> Result<(), RtcError> {
+        let Some(spec) = self.params_by_pt(pt).map(|p| p.spec()) else {
+            return Err(RtcError::UnknownPt(pt));
+        };
+
+        let header = RtpHeader::parse(packet, exts)
+            .ok_or_else(|| RtcError::Other("Failed to parse RTP header".into()))?;
+
+        // Remove header from packet.
+        let packet = &packet[header.header_len..];
+
+        let rid = header.ext_vals.rid;
+        // This doesn't need to be lengthened
+        let rtp_time = MediaTime::new(header.timestamp as i64, spec.clock_rate as i64);
+
+        self.write(
+            pt,
+            wallclock,
+            rtp_time,
+            packet,
+            rid,
+            header.ext_vals,
+            Some(header),
+        )?;
 
         Ok(())
     }
@@ -532,7 +584,7 @@ impl MediaInner {
             self.bytes_retransmitted.push(now, pkt.data.len() as u64);
         }
 
-        let seq_no = source.next_seq_no(now);
+        let seq_no = source.next_seq_no(now, None);
 
         // The resend ssrc. This would correspond to the RTX PT for video.
         let ssrc_rtx = source.ssrc();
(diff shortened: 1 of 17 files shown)


algesten:
This is a start of an API interacting on RTP level rather than Sample level.

There is some problem – we get video through, but something is not right because there are a lot of decompression artefacts. Like a bad packet loss.

There is a problem of the `marker` field in the RTP header is "lost" between the `NullDepacktizer` and `NullPacketizer`. Though even if we jerry righ that field through, it doesn't fix the problem – which is logical because it should only be an indication.

Setting the marker to a constant `false` makes there be no picture at all. `true` or correctly jerry rigged is equally bad result.

algesten:
This is ready for review now.

src/media/mod.rs
```diff
@@ -23,6 +23,9 @@ pub(crate) use inner::{MediaInner, Source};
 /// Half internal structures regarding RTP level.
 pub mod rtp {
     pub use crate::rtp::{Extension, ExtensionMap, ExtensionValues};
+
+    // Exposed for integration tests, not for general usage.
+    pub use crate::rtp::RtpHeader;
```
xnorpx, line 28: I think you should expose utilities for rtp packet creation and parsing etc. Because even that people want to forward rtp packets they still might want to repacketize/serialize in some other way. So having the utilities available for this would be good.

tests/rtp-mode.rs
```diff
@@ -0,0 +1,140 @@
...
+    r.last = max;
+
+    let params = l.media(mid).unwrap().payload_params()[0];
+    assert_eq!(params.spec().codec, Codec::Opus);
+    let pt = params.pt();
+
+    let mut exts = ExtensionMap::empty();
+    exts.set(3, Extension::AudioLevel);
+
+    let to_write = vec![
```
xnorpx, line 59: using RtpHeader here to construct the header would make the test easer to read and to create more tests.

src/media/mod.rs
algesten: Isn't that inviting people to the dark side? 😏

xnorpx: I mean, people are gonna do it anyway. Either they write their own header or take it from another crate :)

algesten: The question is where to stop. I mean I can expose `RtpHeader`, but what about packetizer/depacketizer?

Feels like potentially a whole other crate. rtp-tools or somethin.

tests/rtp-mode.rs
```diff
@@ -0,0 +1,140 @@
...
+        &[
+            //
+            144, 33, 183, 152, 0, 0, 39, 16, 0, 0, 0, 44, 190, 222, 0, 1, 48, 170, 0, 0,
+            // payload
+            0x1, 0x2, 0x3, 0x4,
+        ],
+        // 3
+        &[
+            //
+            144, 33, 183, 154, 0, 0, 54, 176, 0, 0, 0, 44, 190, 222, 0, 1, 48, 172, 0, 0,
```
xnorpx, line 70: 154 can be 155 instead to also exercise lost packet

algesten: Let's try it

algesten: Works

tests/rtp-mode.rs
```diff
@@ -0,0 +1,140 @@
...
+
+#[test]
+pub fn rtp_mode() -> Result<(), RtcError> {
+    init_log();
+
+    let rtc1 = Rtc::builder().set_rtp_mode(true).build();
+    let rtc2 = Rtc::builder()
+        .set_rtp_mode(true)
+        // release packet straight away
+        .set_reordering_size_audio(0)
```
xnorpx, line 23: if this is set to 4 will the test still pass?

xnorpx: when packet is lost

algesten: I think it will release the first two in sequence and not the last one.

xnorpx: I see I need more packets to get it to release the last one.

algesten: yes. it needs to "push out"

src/media/mod.rs
xnorpx: I see, if people want to packetize h264 for themself? 

Tbh, unless it can break things hard I don't see it as a bad thing. I would also let people test out their own packetizers easily, expand on the current ones.

But yes, once you make something public it's hard to go back.

Up to you, worst case people will just copy the code out and have a fork of it.

src/packet/null.rs
```diff
@@ -0,0 +1,37 @@
...
+#[derive(Debug)]
+pub struct NullDepacketizer;
+
+impl Packetizer for NullPacketizer {
+    fn packetize(&mut self, _mtu: usize, b: &[u8]) -> Result<Vec<Vec<u8>>, PacketError> {
+        Ok(vec![b.to_vec()])
+    }
+
+    fn is_marker(&mut self, data: &[u8], previous: Option<&[u8]>, last: bool) -> bool {
+        true
```
xnorpx, line 15: should this ever be hit?

algesten: No. It takes the marker from the incoming packet. I should make a code comment.

src/session.rs
```diff
@@ -508,7 +505,19 @@ impl Session {
         // here we have incoming and depacketized data before it may be dropped at buffer.push()
         let bytes_rx = data.len();
 
-        buf.push(meta, data);
+        // In RTP mode we want to retain the header. After srtp_unprotect, we need to
+        // recombine the header + the decrypted payload.
+        let data = if self.rtp_mode {
+            let hlen = meta.header.header_len;
```
xnorpx, line 511: you need temp allocation here or can you do it in data directly?

algesten: We could potentially allocate once and move the data around in the vec.

For sample based data is just the body, no header.

src/media/inner.rs
```diff
@@ -276,8 +291,8 @@ impl MediaInner {
         // We don't actually want this buffer here, but it must be created before we
         // get to the do_packetize() (as part of handle_timeout).
         let _ = self.buffers_tx.entry(pt).or_insert_with(|| {
-            let max_retain = if spec.codec.is_audio() { 4096 } else { 2048 };
-            PacketizingBuffer::new(spec.codec.into(), max_retain)
+            let max_retain = if is_audio { 4096 } else { 2048 };
```
xnorpx, line 294: unrelated to this pr, but why is audio 4096 and else is 2048?

algesten: NACK handling. Maybe a tad big :)

tests/rtp-mode.rs
```diff
@@ -0,0 +1,140 @@
...
+    assert_eq!(media.len(), 3);
+
+    // no change from standard above
+    let exts = ExtensionMap::standard();
+
+    let h0 = RtpHeader::parse(&media[0].data, &exts).unwrap();
+    let h1 = RtpHeader::parse(&media[1].data, &exts).unwrap();
+    let h2 = RtpHeader::parse(&media[2].data, &exts).unwrap();
+
+    assert_eq!(h0.sequence_number, 47000);
```
xnorpx, line 123: maybe add negative tests for things that should be different for each leg, ssrc etc?

xnorpx (approved):
LGTM! since it's under a flag it's should be safe and not affect the sample path.

algesten:
Weird that it fails on windows. Wtf?! Wonder if it's some timer resolution or something?

xnorpx:
Windows has a default time resolution at 15 ms, you can set it to 1ms resolution if you call some Windows media api. 

You can also use for better accuracy thread time sleep.
https://docs.rs/spin_sleep/latest/spin_sleep/
https://crates.io/crates/spin_sleep

xnorpx:
it passes locally on my windows laptop btw, so funky it doesn't pass in ci

algesten:
And rerunning it, it did pass. My favorite kind of test.

src/media/mod.rs
algesten: For now I'd like str0m to stay narrowly focused on what it does do rather than exposing a whole new range of API. 

People that reach for the rtp-mode API are aiming for a use case that while supported, but isn't the best experience of using str0m. It's advanced usage and you're a bit on your own. I'll make updates to the doc to this effect.

PR #33, opened 2023-01-16
rtc: media added event to provide rid

algesten:


algesten:
This is a breaking change for `Event::MediaAdded`.

PR #36, opened 2023-01-17
add media changed event

davibe:


str0m/src/media/mod.rs
```diff
@@ -977,6 +980,8 @@ impl Media {
                     self.mid, self.dir, new_dir
                 );
 
+                self.need_changed_event = true;
```
k0nserv, line 983: Note, the spec specified two direction fields:

1. `direction` which is the direction requested by the user.
2. `currentDirection` which is the actual direction as resolved based on negotiation.


webrtc-rs used to be implemented like str0m is here with only a single direction field, however we had reason to improve this in https://github.com/webrtc-rs/webrtc/pull/210. I'm not sure if we'll need this too

algesten: @k0nserv the field here would be equivalent of `currentDirection`. Str0m doesn't track the "desired state" in these structures. The desired changes are all kept in `ChangeSet` which is _things to apply to session after negotiation_. https://github.com/algesten/str0m/blob/main/str0m/src/change.rs#L15-L20

algesten: Or One can see it as this chain:

`ChangeSet` -> `Offer` -> `PendingChanges(ChangeSet).apply_answer(Answer)` -> Session

algesten (approved):
Perfect!

str0m/src/media/mod.rs
k0nserv: Ah okay then this is all good then

PR #37, opened 2023-01-17
Fix connection never timing out for ice-lite

### LINKED ISSUE #35: ICE agent in ice-lite never times out
If an ICE agent doesn't receive _any UDP traffic at all_. I..e. no candidate is ever nominated, the agent never times out at all.

That means we get forever hanging clients when there is no UDP connectivity.

algesten:
Close #35

### DIFF
--- a/ice/src/agent.rs
+++ b/ice/src/agent.rs
@@ -827,6 +827,16 @@ impl IceAgent {
         }
 
         if self.ice_lite {
+            // Remote binding request time is the timestamp in the CandidatePair that
+            // is used to decide whether something is timed out or not. We need all
+            // pairs to have this time set, so that pairs that don't receive any
+            // STUN binding requests eventually times out.
+            for p in &mut self.candidate_pairs {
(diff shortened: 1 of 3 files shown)


ice/src/agent.rs
```diff
@@ -827,6 +827,16 @@ impl IceAgent {
         }
 
         if self.ice_lite {
+            // Remote binding request time is the timestamp in the CandidatePair that
+            // is used to decide whether something is timed out or not. We need all
+            // pairs to have this time set, so that pairs that don't receive any
+            // STUN binding requests eventually times out.
+            for p in &mut self.candidate_pairs {
+                if !p.remote_binding_request_time().is_some() {
```
k0nserv, line 835: `p.remote_binding_request_time().is_none()`

k0nserv (approved):

ice/src/agent.rs
algesten: good point! :)

PR #39, opened 2023-01-19
sdp: support the absence of the a=setup attribute

davibe:


sdp/src/parser.rs
```diff
@@ -1131,6 +1131,34 @@ mod test {
...
+        m=audio 0 UDP/TLS/RTP/SAVPF 0\r\n\
+        c=IN IP4 0.0.0.0\r\n\
+        a=inactive\r\n\
+        a=mid:1\r\n\
+        a=rtpmap:0 PCMU/8000\r\n\
+        ";
+
+        let (sdp, _) = sdp_parser().parse(sdp).unwrap();
+
+        // Firefox 'a=setup' attribute can be missing
```
k0nserv, line 1154: Should we raise a bug for Firefox?

davibe: we would have to gather at least the version of Firefox that generated this. I cannot reproduce apparently.

algesten (approved):

PR #40, opened 2023-01-19
sdp: use easy_parse() for better error output

davibe:


sdp/src/lib.rs
```diff
@@ -19,8 +18,8 @@ mod parser;
 /// Errors from parsing and serializing SDP.
 #[derive(Debug, Error)]
 pub enum SdpError {
-    #[error("SDP parse: {0}")]
-    Parse(#[from] StringStreamError),
+    #[error("SDP inconsistent: {0}")]
+    ParseError(String),
```
algesten, line 22: Why "inconsistent" when it really is a _parse error_ quite literally. I mean we could potentially split the error into parsing and then our subsequent validation.

algesten (changes requested):
Looks great. Just a detail around some wording change.

sdp/src/lib.rs
davibe: just missed it when moving around, fixed now

algesten (approved):

PR #42, opened 2023-01-21
Switch sha1 calculations done by rust sha1 crate to openssl

algesten:


daxpedda:
I was reading on Discord that the reason for switching to the OpenSSL SHA-1 implementation is because of the issue of the RustCrypto crate not support ASM on some platforms, notably MSVC, which has a noticeable impact on performance.

Hopefully this can help: https://github.com/RustCrypto/hashes/pull/450.

algesten:
Thanks!

Truth to be told, I don't know if the OpenSSL SHA1 is making any performance difference on Windows. The inline ASM will certainly solve this. I was looking at how to implement the SHA1 in inline ASM but decided it was slightly beyond the scope of what I wanted to attempt right now.

PR #45, opened 2023-01-26
Add session level send pacer

k0nserv:
A pacer that can limit total egress bitrate in a given session. Will
form a part of the Bandwidth Estimation system.


packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+impl LeakyBucketPacer {
+    pub fn set_target_bitrate(&mut self, new_rate: u64) {
+        self.target_bitrate = new_rate;
+    }
+
+    pub fn next_interval(&self) -> Option<Instant> {
+        self.last_budget_increase.map(|l| l + self.pacing)
+    }
+
+    pub fn decide(&mut self, data: &[u8], now: Instant) -> Decision {
```
algesten, line 28: Can this take `len: usize`, so we do't need to pass the data itself?

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+        self.maybe_update_budget(now);
```
algesten, line 29: I'm thinking that rather than implicitly driving time, maybe the component should split the operation into two.

1.`handle_time(&mut self, Instant)`
2.`decide(&mut self, len: usize)`
3. `poll_timeout(&mut self) -> Option<Instant>`

This ties into how the rest of str0m operates, i.e. that driving time forward is an explicit action, while polling for output doesn't drive time.

The `now` that is available in poll_packet is more about registering send times than actually _driving_ time. At least that was the original idea.

Might be splitting hairs. What do you think?

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+        dbg!(data.len(), self.budget);
```
algesten, line 30: Leftover?

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+                "Budget increase, new budget: {}, elapsed: {:?}",
+                self.budget, elapsed
+            );
+        }
+    }
+
+    fn calculate_budget(bitrate: u64, duration: Duration) -> usize {
+        ((bitrate as f64 / 8000.0_f64) * (duration.as_millis() as f64)) as usize
+    }
+}
```
algesten, line 69: I think I'm missing some kind of `poll_timeout()`, that would happen at the end of `do_poll_output()` after there is nothing to send. Or am I misunderstanding?

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+        if elapsed >= self.pacing {
+            let budget = Self::calculate_budget(self.target_bitrate, elapsed);
+            self.budget = budget;
+            self.last_budget_increase = Some(now);
+            debug!(
+                "Budget increase, new budget: {}, elapsed: {:?}",
+                self.budget, elapsed
+            );
+        }
+    }
```
algesten, line 64: I think we should keep track of whether any of the budget actually was used at all. So that we don't produce a timeout if we don't need to.

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn calculate_budget(bitrate: u64, duration: Duration) -> usize {
+        ((bitrate as f64 / 8000.0_f64) * (duration.as_millis() as f64)) as usize
+    }
+}
+
+impl Decision {
+    pub fn emit(&self) -> bool {
+        matches!(self, Self::Emit)
+    }
+}
```
algesten, line 75: Isn't it easier to call `decide` something like `try_allocate_budget(&self, len: usize) -> bool` and return a bool? This enum. Does it clarify?

Maybe it's more clear that the function is a bit more precise in the name?

packet/src/pacer.rs
algesten: Oh it's called `next_interval()`. Hm hm.. Yeah. That works.

algesten (changes requested):
Some small things.

packet/src/pacer.rs
k0nserv: I suppose so

packet/src/pacer.rs
k0nserv: Yes, shall remove

packet/src/pacer.rs
algesten: 👍

packet/src/pacer.rs
algesten: So what I'm thinking is, imagine if we have only audio. It's quite likely we get "gaps", like below. I.e. between two pace, there is no data flowing.

I was thinking if we reach  a pace line and _NO_ budget was used during the previous interval, there is no point in scheduling any new pace, because we can just wait for the next packet and start pacing again then.

```
----- pace 5ms
packet1
packet2
packet3
.
---- pace 5ms
.
.
.
.
---- pace 5ms
.
.
.
packet 4
```

k0nserv: Okay, I'll think about it and see if I can figure out a way to achieve it. Key thing is we don't want to use the budget that was unused in a previous interval when this happens I think.

An early versions of this code had this property that the pacer could essentially "save up" budget, which is not accurate. You only get `X` bytes in a given time span, if you don't use it it's gone

algesten: Yeah. Not saving budget. Just stop pacing if previous budget was totally unused.

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
+use std::time::{Duration, Instant};
+
+pub struct LeakyBucketPacer {
+    /// Target bitrate in bits per second.
+    target_bitrate: u64,
```
pthatcher, line 5: You may want to be aware that eventually you will actually want 3 rates, not just 1.

1.  The target rate: the rate given by the congestion controller that basically means "don't sent more than this"
2.  The padding rate: a value less than or equal to the target rate.  You would not send padding more than this, but you would send non-padding more than this.  Because sometimes you have to send padding, but you don't want to send padding beyond what would be useful for the application.
3.  Some rate related to draining the queue when it gets too big.  For libwebrtc's pacer, I think they can have this be larger than the target rate.  For the pacer I wrote, I instead allocated a portion of the target rate for draining and then had the target rate used elsewhere lowered.

Here is the libwebrtc code:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/pacing/pacing_controller.h;l=223;drc=c14f4e9399fc4e23a6c4c68fc60086bdceec1c90;bpv=1;bpt=1

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+    last_budget_increase: Option<Instant>,
+    /// The current budget in bytes.
+    budget: usize,
+    /// The current pacing i.e. how frequently we refresh the budget.
+    pacing: Duration,
+}
+
+pub enum Decision {
+    Emit,
+    Hold,
```
pthatcher, line 16: Similar to my comment above, I think eventually you may have Padding be one of the decisions here, and it might have a size included.

packet/src/pacer.rs
pthatcher: I was thinking the same thing: that this should be like the other "Sans I/O" things with handle_timeout and poll_timeout and poll_transmission and that kind of thing.

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+            debug!(
+                "Budget increase, new budget: {}, elapsed: {:?}",
+                self.budget, elapsed
+            );
+        }
+    }
+
+    fn calculate_budget(bitrate: u64, duration: Duration) -> usize {
+        ((bitrate as f64 / 8000.0_f64) * (duration.as_millis() as f64)) as usize
+    }
```
pthatcher, line 68: FWIW, for the pacer I wrote, I created types for DataSize and DataRate and that simplified things like this quite a bit.

packet/src/pacer.rs
```diff
@@ -0,0 +1,131 @@
...
+
+pub struct LeakyBucketPacer {
+    /// Target bitrate in bits per second.
+    target_bitrate: u64,
+    /// The last time we refreshed the budget.
+    last_budget_increase: Option<Instant>,
+    /// The current budget in bytes.
+    budget: usize,
+    /// The current pacing i.e. how frequently we refresh the budget.
+    pacing: Duration,
```
pthatcher, line 11: I admit I don't understand why this is needed.   Why not just refresh the budget with every packet?  That's what the libwebrtc pacer does.  Both it and the pacer I wrote work in terms of "last packet sent" and "next time to send a packet", which I think would avoid any delays introduced by this value.

https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/pacing/pacing_controller.cc;l=306;drc=c14f4e9399fc4e23a6c4c68fc60086bdceec1c90;bpv=1;bpt=1

str0m/src/session.rs
```diff
@@ -59,6 +59,11 @@ pub(crate) struct Session {
 
     enable_twcc_feedback: bool,
 
+    /// A pacer for sending RTP at specific rate.
+    pacer: Option<LeakyBucketPacer>,
+    /// A SRTP packet that is being held because the pacer prevented it from being sent.
+    held_srtp_packet: Option<(Vec<u8>, u64)>,
```
pthatcher, line 65: Both the libwebrtc pacer and the one I wrote work in terms of the pacer being a queue.  You send things into it and then things come out of it.  I think that might be a better way to model this.  I guess you could put the queue here, but you probably don't want a queue that's only 1 packet.  That will lead to a lot of dropped packets with a burst such as can happen with a big key frame or screenshare where this is suddenly a lot of motion.

pthatcher (commented):
Sorry for showing up out of no where, but I'm familiar with implementations of similar things and thought a few things might be useful for you to know.

packet/src/pacer.rs
k0nserv: Sounds like we agree. I will split this

packet/src/pacer.rs
k0nserv: Interesting, this is based somewhat on Pion's implementation which the two alternate rates you mentioned.

Padding seems easy enough to add later.

I've been assuming that the system that pushes the data into the pacer will respect the current bitrate here. My thinking is that if the input rate exceeds the target rate we should drop the oldest packets in the queue(s), this is what happens currently. It sounds like you are saying that we should attempt to burst them out instead. 

I think the answer to this might have significant implications on the design

pthatcher: You definitely can *not* assume the rate coming into the pacer will be respected over short periods.  Traffic can be bursty.  And that's kind of the whole point of the pacer.

It's not really a good idea to just drop burst traffic.  It's often that the burst is really important.  It's better to queue and the figure out a better behavior, such as allocating part of the target send rate to draining the queue, or drain out higher priority packets before lower priority packets, or maybe drop the lower priority packets if the queue gets really too big. 

It can get somewhat complicated.  

And it doesn't need to be all done in this PR.  I just wanted to let you know kind of direction you might want to go down eventually.

str0m/src/session.rs
pthatcher: It seems like it would be easy to end up with the logic being spread around quite a bit.  At some point it's likely you'll want a tight coupling between knowing what is queued and deciding what to do about it (how quickly to pace out, what things to drop and not drop, how much to queue before taking action).

algesten: In light of today's discussions, I see the reason to have one. I would image putting some generic queue here with operations for reordering/dropping etc. My main concern previously was to make it the same component as Pacer.

Oh, and also I want to poll one packet out at a time rather than polling Vec<Packet>.

packet/src/pacer.rs
k0nserv: Awesome, this insight is super valuable.

I guess the pressing question is whether the design as outlined here will be problematic.

```
                                                                                                                            
                                        Actual buffers                                                                      
                                           in here                                                                          
                                                                                                                            
┌────────────────────────┐        ┌────────────────────────┐                                                                
│      Media::write      │───────▶│   PacketizingBuffer    ├─────────┐                                                      
└────────────────────────┘        └────────────────────────┘         │                                                      
                                                                     │          ┌────────────────────────┐                  
                                                                     ├─────────▶│         Pacer          │────poll_packet──▶
                                                                     │          └────────────────────────┘                  
┌────────────────────────┐        ┌────────────────────────┐         │                                                      
│      Media::write      │───────▶│   PacketizingBuffer    │─────────┘                 Store packets                        
└────────────────────────┘        └────────────────────────┘                          until next poll                       
                                                                                                                            
                                                                                                                            
                                        Actual buffers                                                                      
                                           in here                                                                          
                                                                                                                            ```

If choice around prioritisation need to be made I can see how this isn't great 🤔

Comment (k0nserv) [reply packet/src/pacer.rs:5]: If choices around prioritisation need to be made I can see how this isn't great 🤔

File: packet/src/pacer.rs (hunk 10 above)
Comment (k0nserv) [reply packet/src/pacer.rs:11]: The thinking here is that this will be used as follows

```rust
// On timeout
// Aim to call this every 5 ms
pacer.refresh_budget(now);
```

In [poll_packet](https://github.com/algesten/str0m/blob/6ebbc9776e8dc1b300092087d322b7f93b3f52c0/str0m/src/session.rs#L603)

```rust
        for m in only_media_mut(&mut self.media) {
            let twcc_seq = self.twcc;
            if let Some((header, seq_no)) = m.poll_packet(now, &self.exts, &mut self.twcc, buf) {
                trace!("Poll RTP: {:?}", header);

                let protected = srtp_tx.protect_rtp(buf, &header, *seq_no);
                let should_send = self
                    .pacer
                    .budget_allows_send(protected.len());

                if should_send {
                    self.twcc_tx_register.register_seq(twcc_seq.into(), now);
                    return Some(protected.into());
                } else {
                    self.held_srtp_packet = Some((protected, twcc_seq));

                    break;
                }
            }
        }

```

It's a bit different than if the queue lived within the pacer. The queued packets here are within each media line.

Str0m's contract is to call `poll_output` until a timeout is returned and here we'll return all packets allowed by the budget before doing so.

pthatcher: I think the problem is just that 5ms is not high resolution enough, so it seems like it's better to just call poll_timeout() and have that return the time when the next packet should be sent, at which point you would call handle_timeout/poll_output.  That's why I think it's easier to think of this not as whether or not to send or hold a given packet, and instead to think of this as there's a time you want to send a packet.  Or, to use the term of the name of the struct, there is a time to leak the next drop out of the bucket.   And, yes, that's easier to do when you have the contents of the bucket.  

But if you didn't want to hold the queue in the pacer, I suppose what you could do is pass info about the queue (perhaps even an &mut) every time you call poll_output.  So it doesn't require holding the queue, but it doesn't require knowing something about the queue at the time poll_output is called.

packet/src/pacer.rs
pthatcher: If I recall correctly, libwebrtc actually has multiple buffers that it round-robins through.  But that's done more as an implementation detail of the pacer.  I suppose it just depends on where you draw the line of where the "pacer" is.   You could view the "PacketizingBuffers+Pacer" in your diagram as the "whole Pacer" and the "Pacer" in your diagram as just one part of the "whole Pacer".

At the end of the day, regardless of what you call the pieces, you need:

1.  Something that cares a lot about time and decides _when_ to send a packet (controls when to leak).
2.  A place (or places) to store packets that are waiting for their turn to go (the bucket, or buckets).
3.  Something to decide which packet comes out when it's time to send one (controls what to leak)
4.  Something to decide what happens when the bucket(s) is (are) getting too full.
5.  A way to get or create padding packets when there is nothing stored (what to send when the bucket is empty).

It sounds like you'll do (5) later.

Your diagram seems like it could work for (1) and (2).  I guess you just need to determine how (3) and (4) happen.

k0nserv: I was just looking at that code in libWebRTC that does prioritization. Seems easy enough to add later.

I think I'm gonna aim for 1, 2, and 4 initially, leaving 3 and 5 for later. For 4 this is what you were hinting at with a drain bitrate, right? 

For example, if the target bitrate is ~1Mbit/s(625 bytes per 5ms interval) and we find at the next send interval that we actually have (625 + **X** bytes) to send, we need to allow sending at least some, but maybe not all of **X**?

For #5, if we find that we have `400 bytes` we should make up the remaining 200 bytes with padding packets

packet/src/pacer.rs
k0nserv: That all makes sense, except the singular use of *packet*, isn't it the next time when *packets* should be sent i.e. we queue up packets until the next send time and then send them all. My understanding is that this "burstiness" is important to establish and identify the packet groups to later calculate the inter group delay variation

packet/src/pacer.rs
algesten: I don't think we need 2 in the first iteration.

k0nserv: We definitely need it, but it doesn't need to be within the pacer. For now we can keep using the `PacketizingBuffer`s for it

packet/src/pacer.rs
pthatcher: The entire point of a pacer is to _avoid_ burstiness.  And googcc does not rely on burstiness to function.  

You may be thinking of probing, which has specific bursts that are done at a higher rate than the target rate (thus decreasing the interval, or increasing the burstiness).  But even when probing, you're not increasing burstiness such that the packets are sent with _0_ interval between them.  It's just a probe with a decreased delay (an increased rate).  In fact, when probing, there's a specific rate/interval between them that is being probed.

For example, see here:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/pacing/bitrate_prober.cc;l=178;drc=3297fb6d70a02d8c4af8db2bc9e315053744442b;bpv=1;bpt=1

k0nserv: Hmm,

my understanding is based on section 4 of the IETF draft:

>  Pacing is used to actuate the target bitrate computed by the
   controllers.
>   When media encoder produces data, this is fed into a Pacer queue.
   The Pacer sends a group of packets to the network every burst_time
   interval.  RECOMMENDED value for burst_time is 5 ms.  The size of a
   group of packets is computed as the product between the target
   bitrate and the burst_time.

as well as the notion that forming these packet groups and identifying their receive deltas is important to calculate the inter group delay variation.

It sounds like this is one of those instances when the IETF draft is outdated then.


With the approach you outline and what libWebRTC does wouldn't knowledge of the buffers be required to make decisions around when the next send time should be meaning the proposed design here is no good?

pthatcher: Interesting.  libwebrtc can send more than one packet a time if certain field trails are on (see here: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/pacing/task_queue_paced_sender.cc;l=81;drc=3297fb6d70a02d8c4af8db2bc9e315053744442b;bpv=1;bpt=1 and https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/pacing/pacing_controller.cc;l=75;drc=3297fb6d70a02d8c4af8db2bc9e315053744442b;bpv=1;bpt=1)

I don't believe that's the default behavior.  The default appears to be a 0 for the burst interval, which means no bursting. But I could be wrong.  Maybe I'm missing something.  

I do see one place where a 5ms constant is used that has the comment "Time limit in milliseconds between packet bursts.", but that value is only used once: to give the initial budget size, which I think means the pacers acts as though 5ms has already passed when it starts.  

So, it's definitely possible I'm missing something fundamental.  The only way to get it right may be to read the libwebrtc code very, very carefully (perhaps more carefully than I thought I did already).

k0nserv: The true value seems to be 40. It's [set](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/peerconnection/rtc_peer_connection_handler.cc;l=1089?q=pacer_burst_interval&ss=chromium) in Chromium an passed all the way down to `PacingController` where it's stored in `send_burst_interval_`. You can see how it's used in `PacingController::NextSendTime`.

It's strange that other parts of the code, to do with identifying inter arrival groups, still use the value 5ms

str0m/src/session.rs
k0nserv: Looking at this more and further reading libWebRTC's code I do think we want to move the queues from the many `PacketizingBuffer`s into the pacer. The pacer needs to have knowledge of queue sizes as well as being able to pick out different media. For example, consider the 40ms burst interval I discovered above, this is not workable for audio, hence libWebRTC handles this by not pacing audio. I see they also have a field trial to not pace retransmissions, which makes sense for similar reasons.

packet/src/pacer.rs
k0nserv: libWebRTC uses the concept of debt instead of a budget which seems much nicer as a way to model this. You incur debt when sending and having excessive outstanding debt influences when you are allowed to send again.

str0m/src/media/mline.rs
```diff
@@ -991,6 +1001,55 @@ impl MLine {
         }
     }
 
+    /// Return the queue state for outbound RTP traffic for this mline.
+    ///
+    /// For the purposes of pacing each mline is treated as its own packet queue. Internally this
+    /// is spread across many [`PacketizingBuffer`] which is where the actual packets are queued.
+    pub fn send_queue_state(&self, now: Instant) -> QueueState {
```
algesten, line 1008: I think to be consistent with the naming, this would be better to tie into `buffer_tx`. Something like `buffer_tx_queue_state`.

str0m/src/media/mline.rs
```diff
@@ -991,6 +1001,55 @@ impl MLine {
...
+                },
+            );
+
+        QueueState {
+            queue_id: QueueId::new(self.index() as u64),
+            size,
+            packet_count: packet_count as u32,
+            queue_time,
+            media_states: MediaStates {
+                audio: audio_kind_sate,
```
algesten, line 1047: typo

str0m/src/media/mline.rs
```diff
@@ -991,6 +1001,55 @@ impl MLine {
...
+                video: video_kind_state,
```
algesten, line 1048: Why are these values split into audio/video? I understand we need that separation higher up, but an m-line can only be one of them. Wouldn't it be better to just have `media_state` and a flag `is_audio: bool`.

packet/src/pacer.rs
```diff
@@ -0,0 +1,835 @@
...
+/// The state of a single upstream queue.
+/// The pacer manages packets across several upstream queues.
+#[derive(Debug)]
+pub struct QueueState {
+    /// The unique identifier for this queue within the session.
+    pub queue_id: QueueId,
+    /// The total size of queued payloads within the queue.
+    pub size: DataSize,
+    /// The total number of packets in the queue.
+    pub packet_count: u32,
```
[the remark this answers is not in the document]
k0nserv: We need to divide it with `Duration` and `Duration / usize` isn't implemented. Rather than using `usize` here and truncating I figured it'd be better to communicate this limitation here.

packet/src/pacer.rs
```diff
@@ -0,0 +1,835 @@
...
+/// Various queue state broken down by media type.
+#[derive(Debug)]
+pub struct MediaStates {
+    pub audio: MediaState,
+    pub video: MediaState,
+}
+
+/// The outcome of a call to [`Pacer::poll_action`].
+#[derive(Debug, Clone, Copy)]
+pub enum PollOutcome<'a> {
```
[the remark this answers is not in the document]
k0nserv: Damn I though I had

packet/src/pacer.rs
```diff
@@ -0,0 +1,835 @@
...
+            );
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PacketKind {
+    Audio,
+    Video,
+}
```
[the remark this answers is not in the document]
k0nserv: I was thinking the same thing, I'll try and find it

rtp/src/bandwidth.rs
```diff
@@ -8,10 +11,16 @@ use std::fmt;
...
+    pub const ZERO: Self = Self::new(0);
+
+    pub const fn new(bps: u64) -> Self {
         Bitrate(bps as f64)
     }
 
+    pub const fn gbps(gbps: u64) -> Self {
+        Self::new(gbps * 10_u64.pow(9))
+    }
+
```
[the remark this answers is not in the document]
k0nserv: I had a use case for this specific case and I know you dislike dead code so I decided to not add `mbps` or `kbps` until we have a need for them

str0m/src/media/mline.rs
k0nserv: Sounds good

str0m/src/media/mline.rs
k0nserv: Ah, a forgotten rename.

str0m/src/media/mline.rs
k0nserv: It's a bit more future proof i.e. if we were to move to a single queue instead of one per mline the pacer code already handles it. This fact is utilized in the pacer tests where there's a single packet queue that may contain both audio and video

packet/src/buffer_tx.rs
```diff
@@ -22,22 +23,27 @@ pub struct PacketizedMeta {
...
     queue: VecDeque<Packetized>,
     emit_next: usize,
+    last_emit: Option<Instant>,
+    is_audio: bool,
     max_retain: usize,
 }
 
 impl PacketizingBuffer {
-    pub fn new(pack: CodecPacketizer, max_retain: usize) -> Self {
+    pub fn new(pack: CodecPacketizer, max_retain: usize, is_audio: bool) -> Self {
```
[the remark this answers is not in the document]
k0nserv: I considered that but then decide to do this instead, I don't feel strongly.

k0nserv:
> I'm thinking it should be possible for the PacketizingBuffers to just produce QueueState straight away (without all those independent loops that you code commented on).

Yeah that code could probably do with some finessing

packet/src/pacer.rs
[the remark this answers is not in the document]
k0nserv: There's `MediaKind` but it's in `str0m`. I could also foresee us wanting to have a `Retransmission` variant here

packet/src/pacer.rs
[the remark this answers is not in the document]
algesten: ok. let's do a code note.

packet/src/pacer.rs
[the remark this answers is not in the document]
algesten: alright. leave it in for now.

rtp/src/bandwidth.rs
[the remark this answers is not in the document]
algesten: 👍

str0m/src/media/mline.rs
algesten: Hmmm… "future proof" and "test reasons" are not very high on my agenda. It kinda confuses me because I expect these two be used at the same time, and they won't be.

packet/src/pacer.rs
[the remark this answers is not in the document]
k0nserv: Sure thing

str0m/src/media/mline.rs
k0nserv: Okay I guess I can complicate the tests instead

str0m/src/media/mline.rs
```diff
@@ -250,7 +255,8 @@ impl MLine {
         tx.update_clocks(rtp_time, wallclock);
 
         let buf = self.buffers_tx.entry(pt).or_insert_with(|| {
-            let max_retain = if codec.is_audio() { 4096 } else { 2048 };
+            let is_audio = codec.is_audio();
+            let max_retain = if is_audio { 4096 } else { 2048 };
```
algesten, line 259: This change is not needed anymore.

str0m/src/session.rs
```diff
@@ -207,6 +217,11 @@ impl Session {
                 }
             }
         }
+
+        let queue_states: Vec<_> = only_m_line(&self.m_lines)
+            .map(|line| line.buffers_tx_queue_state(now))
+            .collect();
```
algesten, line 223: This is a heap allocation on every handle_timeout – something I tried to avoid when possible.

I haven't looked into how Rust optimizes this, but since the m-line count is pretty static, I wonder if we should at least move the vec to be a reused instance variable on Session?

str0m/src/session.rs
```diff
@@ -626,24 +641,60 @@ impl Session {
...
+        let real_now = Instant::now();
+
+        let queue_states: Vec<_> = only_m_line(&self.m_lines)
+            .map(|line| line.buffers_tx_queue_state(now))
+            .collect();
 
+        // Figure out which, if any, queue to poll
+        let PollOutcome::PollQueue(queue) = self.pacer.poll_action(now, &queue_states) else {
+            return None;
+        };
```
algesten, line 654: API wise it doesn't look right to double up on the queue states here. 

The pattern everywhere else is:

1. handle_timeout -> decide what to do. Make state change such as deciding which packet goes next.
2. poll_output -> Enact the state change in 1

Philosophically time doesn't move between 1 and 2, so there's no need to redo any calculations based on time in 2.

algesten (changes requested):
Some bits i discovered on a second pass.

str0m/src/media/mline.rs
k0nserv: Right you are

str0m/src/session.rs
k0nserv: I've been thinking the same thing, but have been ignoring it as a premature optimisation for the time being. Can fix though via instance variable on Session before merging, if you wish

str0m/src/session.rs
k0nserv: Things within the queue state can and do change between calls to `Pacer::poll_action` that impact the calculations. For example, packets are removed from the queues, which might mean they become empty, this also impacts the total queue size as well as the average queue time for the packets within the queue.

When we end up in the drain queue state the following calling pattern might happen:

1. `Pacer::poll_action` returns `PollOutcome::Nothing` because we need to wait until the next send time
2. We call `Pacer::handle_time` as a side effect of time moving forward
3. We repeatedly call `Pacer::poll_action` until the queues have been drained

str0m/src/session.rs
k0nserv: Addressed in 9dd30d2

str0m/src/session.rs
```diff
@@ -625,25 +641,58 @@ impl Session {
...
+                        .checked_duration_since(queued_at)
+                        .map(|d| d.as_micros() as i64)
+                        .unwrap()
+                };
+                let kind = if mline.kind() == MediaKind::Audio {
+                    "audio"
+                } else {
+                    "video"
+                };
+                println!("QUEUE_STAT {},{},{}", header.ssrc, delta_us, kind);
```
davibe, line 685: use debug instead of println

packet/src/pacer.rs
```diff
@@ -0,0 +1,967 @@
...
+const MAX_BITRATE: Bitrate = Bitrate::gbps(10);
+const MAX_MEDIA_DEBT_IN_TIME: Duration = Duration::from_millis(500);
+
+/// A packet Pacer.
+///
+/// The pacer is responsible for ensuring correct pacing of packets onto the network at a given
+/// bitrate.
+pub trait Pacer {
+    /// Set the target bitrate to send at. This can be exceeded if required.
+    fn set_target_bitrate(&mut self, target_bitrate: Bitrate);
```
davibe, line 17: Initially I did not expect this. I thought it was not needed because sr0m knows (measures) the bitrate of the streams flowing, and it's more realistic compared to a supposed target bitrate - especially in the case of streams like screen shares which can diverge a lot from the target.
Then I also thought this could be misleading as someone could think this is the target bitrate you want to test for (the one you want to reach) and not the current one.
All in all, I don't think these are blockers at our stage.

davibe (dismissed):

str0m/src/session.rs
k0nserv: This is temporary to get this output to STDOUT, I wrote a small python script to summarise these stats.

<details>
<summary>queue-latency.py</summary>

```python
import sys
import pandas as pd
import numpy as np
import io


def read_stats(path):
    output = io.StringIO()
    with open(path) as f:
        for line in f:
            if not line.startswith("QUEUE_STAT"):
                continue

            clean = line.removeprefix("QUEUE_STAT").strip()
            print(clean, file=output)

    output.seek(0)
    return output


def percentile(n):
    def percentile_(x):
        return np.percentile(x, n)

    percentile_.__name__ = "percentile_%s" % n
    return percentile_


if __name__ == "__main__":
    path = sys.argv[1]
    stats = read_stats(path)

    df = pd.read_csv(
        stats, header=None, names=["SSRC", "queue_time", "kind"], on_bad_lines="error"
    )
    queue_counts = df["queue_time"].value_counts()
    any_bad_values = False
    # df.drop(columns=["kind"], inplace=True)

    if any_bad_values:
        count = queue_counts[-1]
        print(f"Got {count} bad values for `queue_time`")

    # df = df.head(3000)
    # assert df.shape[0] == 3000

    stats = (
        df.groupby(["SSRC", "kind"])
        .agg(
            [
                percentile(50),
                percentile(75),
                percentile(95),
                percentile(99),
                "max",
                "count",
            ]
        )
        .reset_index()
    )

    print(stats)
```

</details>

I'll remove this before I merge but maybe it's something we want long term, it was useful for this work.

davibe: nice

str0m/src/session.rs
algesten: @k0nserv I think the way to model that would be:

1. Pacer::handle_timeout -> change state
2. Pacer::poll_action -> get poll packet
3. Pacer::poll_timeout -> straight away
4. repeat from 1

I maintain we should not need the queue state update also on `poll_action`.

k0nserv: Well, in the instance when we are draining the queues we want to ensure all the packets are sent as a burst i.e. without any time between them and without anything being queued in between.

In libWebRTC they handle this with a loop but of course we cannot loop since we are producing a single packet at the time. What I've done here is used the fact that time only moves forward when we queue packets and handle time to ensure we get the same behaviour. This works like this:


1. When we enter queue draining mode we record the current value of `now`
2. Then any subsequent calls to `poll_action` will calculate the send time for those packets as the instant they were queued. 

This ensures that when we enter the queue draining mode we call `poll_action` until we have drained the queues fully. This assumption breaks down if we allow the time to move forward between these calls

algesten: Ok. Then I suggest you take ownership of the queue states.

So on `handle_timeout` you do:

`Pacer::get_queue_states(&mut self) -> &mut Vec<QueueState>`

```
let states = pacer.get_queue_states();
states.clear();
loop buffer in buffer_txs {
   states.push(buffer.queue_state());
}
```

And then in `poll_output` we simply work from the queue state that we got in `handle_timeout`.

k0nserv: As it currently stands this doesn't work for the following reason.

When we remove a packet from the queue we would need to reduce the `total_queue_time` value based on the amount of time that packet was queued, this information is currently only available within `PacketizingBuffer`. For this to be possible `QueueState` would need to contain information about all the packets in the queue. A similar argument applies to `QueueState::size` which needs to be reduced based on the size of the packet we remove from the queue

algesten: Ok. Rather than me trying to figure out the solve, can we instead turn it around?

I need `handle_timeout` to do the state change, and `poll_packet` to deliver packets without doing state changes.

How do we achieve that?

algesten: https://github.com/algesten/str0m/pull/45#discussion_r1121672266

I studied the code a bit more, and I circle back to that this one should be possible. _straight away_ here means now, so time will not move forward just because we do a handle_timeout.

k0nserv: It sounds like you are suggesting that every call to `Pacer::poll_action` is followed by a call to `Pacer::handle_time`. This doesn't work at this level, but it would if it applied to `Session::poll_datagram` and `Session::handle_timeout`.

To be clear the problem we end up with here is this: `Session::queue_states` is a snapshot of the packet queues, when we call `Session::poll_datagram` the state of those packet queues can changes causing `Session::queue_states` to no longer be an accurate snapshot.

> I need handle_timeout to do the state change, and poll_packet to deliver packets without doing state changes.

I'm a bit confused about the motivation for this, `Session::poll_packet` already causes state changes(moving the TWCC counter and removing packets from the selected queue). What I'm currently doing is ensuring that the snapshot of the packet queues stays up to date with these state changes that are already happening.

algesten (approved):
LGTM!

algesten (approved):

str0m/src/session.rs
k0nserv: Removed via rebase now.

davibe (approved):

packet/src/buffer_tx.rs
```diff
@@ -103,6 +109,42 @@ impl PacketizingBuffer {
...
+            });
+
+        state.update_last_send_time(self.last_emit);
+        state.update_leading_queue_time(self.leading_queue_time());
+
+        state
+    }
+
+    fn leading_queue_time(&self) -> Option<Instant> {
+        self.queued_packets().next().map(|p| p.meta.queued_at)
```
pthatcher, line 137: Nit:  slightly nicer as ```Some(self.queued_packets().next()?.meta.queued_at)```

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
+use std::collections::HashMap;
+use std::fmt;
+use std::time::{Duration, Instant};
+
+use rtp::{Bitrate, DataSize};
+
+const MAX_BITRATE: Bitrate = Bitrate::gbps(10);
+const MAX_MEDIA_DEBT_IN_TIME: Duration = Duration::from_millis(500);
+
+pub enum PacerImpl {
```
pthatcher, line 10: Nit: I kind of like the naming scheme of KnownPacer for this type of approach.

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+    }
+}
+
+/// A packet Pacer.
+///
+/// The pacer is responsible for ensuring correct pacing of packets onto the network at a given
+/// bitrate.
+pub trait Pacer {
+    /// Set the target bitrate to send at. This can be exceeded if required to drain excessively
+    /// long packet queues.
```
pthatcher, line 58: I think whether or not the given rate should be exceeded should be configurable.  For example, libwebrtc has the PacingController::drain_large_queues_ to control that.  

And we should probably comment that if you enable sending above the given rate, that's really only intended to handle bursts (like key frames).  If you send over a long period above the given rate and enable going above it, you can of defeat the purpose of having a given rate in the first place.

packet/src/pacer.rs
pthatcher: libwebrtc calls it the "pacing rate" or "data rate"and definitely has a setter for it.  It also separately has a "padding rate" or "pad rate".   These rates are based on the output of the congestion controller, but they are *not* just the target send rate from the congestion controller.  For example, by default WebRTC applies 1.1 "pacing factor". 

So, I would call this set_pacing_bitrate or set_pacing_bitrates with a TODO to have a separate padding bitrate.

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+
+/// The outcome of a call to [`Pacer::poll_action`].
+#[derive(Debug, Clone, Copy)]
+pub enum PollOutcome {
+    /// The caller **MUST** poll the next packet from the queue that produced the contained [`QueueState`]
+    /// and send it.
+    PollQueue(QueueId),
+    /// The caller MUST do nothing for now.
+    Nothing,
+    // TODO: Produce padding when we are sending very infrequently or if we need it to probe.
```
pthatcher, line 119: Padding it's just for when sending infrequently.  It (or probing) is important for an SFU that can't just send more when ever it wants be adjusting an encoder.

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+    fn register_send(&mut self, now: Instant, _packet_size: DataSize, from: QueueId) {
+        let e = self.last_sends.entry(from).or_insert(now);
+        *e = now;
+    }
+}
+
+/// A leaky bucket pacer that can overshoot the target bitrate when required.
+pub struct LeakyBucketPacer {
+    /// Target bitrate.
+    target_bitrate: Bitrate,
```
pthatcher, line 168: Again, I would suggest not using the name "target" here, since it would easily be confused with the output of the congestion controller.  Better to call this "pacing rate" or "data rate" or maybe "unadjusted_rate".

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+
+    need_immediate_timeout: bool,
+
+    next_poll_outcome: Option<PollOutcome>,
+}
+
+impl Pacer for LeakyBucketPacer {
+    fn set_target_bitrate(&mut self, new_rate: Bitrate) {
+        self.target_bitrate = new_rate;
+    }
```
pthatcher, line 195: Don't you need to call self.maybe_update_adjusted_bitrate() here?

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+
+    fn poll_timeout(&self) -> Option<Instant> {
+        if self.need_immediate_timeout {
+            return self.last_send_time;
+        }
+
+        let next_handle_time = self.last_handle_time.map(|lh| lh + self.pacing);
+
+        match (next_handle_time, self.next_send_time) {
```
pthatcher, line 204: Nit: this could use a option_min helper method.

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+            (None, Some(ls)) => Some(ls),
+            (Some(nh), None) => Some(nh),
+            (None, None) => None,
+        }
+    }
+
+    fn handle_timeout(&mut self, now: Instant, iter: impl Iterator<Item = QueueState>) {
+        self.need_immediate_timeout = false;
+        self.queue_states.clear();
+        self.queue_states.extend(iter);
```
pthatcher, line 215: Is this not the same as self.queue_states = iter.collect()?  If so, why not just always pass in a Vec<QueueState>?

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+    fn poll_action(&mut self) -> PollOutcome {
+        let Some(next) = self.next_poll_outcome.take() else {
+            return PollOutcome::Nothing;
+        };
+
+        self.need_immediate_timeout = true;
+        next
+    }
+
+    fn register_send(&mut self, now: Instant, packet_size: DataSize, _from: QueueId) {
```
pthatcher, line 259: Nit: I'd call this handle_packet_sent or handle_sent.

packet/src/pacer.rs
```diff
@@ -0,0 +1,1032 @@
...
+
+    fn next_send_time(&self, now: Instant) -> (Instant, Option<&QueueState>) {
+        // If we have never sent before, do so immediately on an arbitrary non-empty queue.
+        if self.last_send_time.is_none() {
+            let mut queues = self.queue_states.iter().filter(|q| q.packet_count > 0);
+
+            return (now, queues.next());
+        };
+
+        let unpaced_audio = self
```
pthatcher, line 314: If the purpose of the PacketKind is to pace or not pace, why call it Audio and Video?  Why not instead call it Unpaced and Paced?

PR #46, opened 2023-01-26
Add periodic events with peer and media statistics

davibe:


str0m/src/stats.rs
```diff
@@ -0,0 +1,147 @@
...
+    pub tx: Bytes,
+    pub rx: Bytes,
+    pub ts: Instant,
+}
+
+impl StatsSnapshot {
+    fn new(ts: Instant) -> StatsSnapshot {
+        // TODO: is there a way to only provide the `ts` property
+        // which does not conform to Default,
+        // and have the rest generated by ..Default::default() ?
```
k0nserv, line 29: Unfortunately not, I think I recall seeing an RFC to address this but I can't find it

davibe: amazing, thx

str0m/Cargo.toml
```diff
@@ -9,6 +9,8 @@ tracing = "0.1"
 rand = "0.8"
 bitvec = "1"
 once_cell = "1"
+serde = "1"
+itertools = "0.10"
```
algesten, line 13: Do we really need this? I don't want more dependencies than we need. Does it solve a big problem?

str0m/src/lib.rs
```diff
@@ -237,6 +245,12 @@ pub enum Event {
 
     /// A data channel has been closed.
     ChannelClose(ChannelId),
+
+    PeerStats(PeerStats),
+
+    MediaIngressStats(MediaIngressStats),
+
+    MediaEgressStats(MediaEgressStats),
```
algesten, line 253: These 3 events needs doc.

str0m/src/lib.rs
```diff
@@ -989,15 +1018,32 @@ impl Rtc {
 
     fn do_handle_timeout(&mut self, now: Instant) {
         self.last_now = now;
+        info!("timeout");
```
algesten, line 1021: Remove

str0m/src/lib.rs
```diff
@@ -989,15 +1018,32 @@ impl Rtc {
...
         self.ice.handle_timeout(now);
         self.sctp.handle_timeout(now);
         self.session.handle_timeout(now);
+        if self.stats.handles_timeout(now) {
```
algesten, line 1025: Let's rename this `wants_timeout`

str0m/src/media/mod.rs
```diff
@@ -248,12 +248,13 @@ pub struct Media {
     /// These are created first time we observe the SSRC in an incoming RTP packet.
     /// Each source keeps track of packet loss, nack, reports etc. Receiving sources are
     /// cleaned up when we haven't received any data for the SSRC for a while.
-    sources_rx: Vec<ReceiverSource>,
+    // TODO: probably not ok to make them public
+    pub sources_rx: Vec<ReceiverSource>,
```
algesten, line 252: Very not ok. `Media` is public API and `ReceiverSource` isn't.

str0m/src/media/sender.rs
```diff
@@ -10,9 +10,10 @@ use super::Source;
 pub struct SenderSource {
     ssrc: Ssrc,
     repairs: Option<Ssrc>,
-    rid: Option<Rid>,
+    pub rid: Option<Rid>,
```
algesten, line 13: There's accessor methods for rid.

str0m/src/media/receiver.rs
```diff
@@ -16,6 +16,7 @@ pub struct ReceiverSource {
     sender_info: Option<SenderInfo>,
     sender_info_at: Option<Instant>,
     fir_seq_no: u8,
+    pub bytes_rx: u64,
```
algesten, line 19: :(

str0m/src/stats.rs
```diff
@@ -0,0 +1,231 @@
...
+
+use crate::{media::Source, Mid, Rtc};
+use rtp::Rid;
+
+pub struct Stats {
+    last_snapshot: StatsSnapshot,
+    events: VecDeque<StatEvent>,
+}
+
+type Bytes = u64;
```
algesten, line 15: This will be confusing for readers since a lot of people are familiar with the `bytes` crate. Let's just use u64.

algesten (changes requested):
Looks quite good.

I think overall structure needs to change. Now we have this `StatsSnapshot` that requires access to very low level fields of each level – which in turn requires you to make a whole bunch of fields `pub`, that really shouldn't be pub.

It's especially bad when there's deep field dependencies two levels. Like Session level reaches all the way into a public field on a ReceiverSource.

Rather than manipulating fields directly I think you should pass the data via the in between structures and make proper functions for them.

I.e. `Media::update_rececived_bytes(&mut self, ssrc: Ssrc, len: u64)`. Which in turn calls `ReceiverSource::update_received_bytes(&mut self, len: u64)`.

Similar in the other direction when creating a report. Let each layer create the reports for that layer. `Media::get_stats_report(…)` and they in turn delegates to `ReceiverSource::get_stats_report(…)`.

str0m/Cargo.toml
davibe: I used it for group_by() alone. Up to you.

algesten: Lets try without

davibe: done

davibe: done

rtp/Cargo.toml
```diff
@@ -13,6 +13,7 @@ once_cell = "1"
 # instead of relying on multiple TLS libraries.
 openssl = { version = "0.10", features = ["vendored"] }
 openssl-sys = "^0.9"
+serde = "1"
```
algesten, line 16: Maybe a comment that we can move this dep if we move the id generation to a util crate?

str0m/src/media/mod.rs
```diff
@@ -1193,6 +1198,31 @@ impl Media {
...
+            snapshot.tx += bytes_tx;
+            let key = (self.mid, s.rid());
+            if let Some(v) = snapshot.ingress.get_mut(&key) {
+                *v += bytes_tx;
+            } else {
+                snapshot.ingress.insert(key, bytes_tx);
+            }
+            // TODO: handle remote stats for this egress source
+        }
+    }
```
algesten, line 1225: Wouldn't it be cleaner to implement `visit_stats` on ReceiverSource and SenderSource here?

str0m/src/stats.rs
```diff
@@ -0,0 +1,190 @@
...
+/// note: when simulcast is disabled, `rid` is `None`
+#[derive(Debug, Clone)]
+pub struct MediaEgressStats {
+    pub mid: Mid,
+    pub rid: Option<Rid>,
+
+    pub bitrate_tx: f32,
+    pub ts: Instant,
+    // TODO
+    pub remote: RemoteIngressStats,
```
algesten, line 75: Maybe comment this out?

str0m/src/stats.rs
```diff
@@ -0,0 +1,190 @@
...
+/// note: when simulcast is disabled, `rid` is `None`
+#[derive(Debug, Clone)]
+pub struct MediaIngressStats {
+    pub mid: Mid,
+    pub rid: Option<Rid>,
+
+    pub bitrate_rx: f32,
+    pub ts: Instant,
+    // TODO
+    pub remote: RemoteEgressStats,
```
algesten, line 94: ditto

algesten (changes requested):
Looks good! A couple of small questions.

str0m/src/stats.rs
davibe: makes sense

str0m/src/media/mod.rs
algesten: Yeah, pass args in `visit_stats`. I like that.

algesten (approved):
Very nice!

PR #47, opened 2023-01-26
Add Rate Controller for BWE

k0nserv:


algesten (approved):
Yeah, right. Ok then.

k0nserv:
Why did CI fail? I don't think I touched any of the code it's mad at

packet/src/bwe/rate_controller.rs
```diff
@@ -0,0 +1,376 @@
...
+    Increase,
+    Decrease,
+}
+
+impl State {
+    fn transition(&self, signal: Signal) -> Self {
+        match signal {
+            Signal::Overuse => self.overuse(),
+            Signal::Underuse => self.underuse(),
+            Signal::Normal => self.normal(),
```
pthatcher, line 165: I think this would be more clear as:

```
match (self, signal) {
  (_, Signal::Overuse) => Self::Decrease,
  (_, Signal::Underuse) => Self::Hold,
  (Self::Hold | Self::Increase, Signal::Normal) => Self::Increase,
  (Self::Decrease, Signal::Normal) => Self::Hold,
}
```

packet/src/bwe/rate_controller.rs
```diff
@@ -0,0 +1,376 @@
...
+            let response_time_ms = self.last_rtt_us.map(|rtt| rtt / 1000.0).unwrap_or(0.0) + 100.0;
+            let alpha = 0.5 * (since_last_update / response_time_ms).min(1.0);
+            let expected_packet_size = self.estimated_packet_size();
+            self.estimated_bitrate + (alpha * expected_packet_size).max(1000.0)
+        } else {
+            // Multiplicative increase
+            let eta = 1.08_f64.powf(since_last_update.min(1.0));
+            eta * self.estimated_bitrate
+        };
+        let max = self.last_observed_bitrate.map(|r| r * 1.5_f64);
```
pthatcher, line 102: I think a lot of these constants should be given constant names (like BETA and OBSERVED_BIT_RATE_SMOOTHING_FACTOR).

packet/src/bwe/rate_controller.rs
```diff
@@ -0,0 +1,376 @@
...
+    averaged_observed_bitrate: MovingAverage,
+    /// The last time we updated the estimated bitrate.
+    last_estimate_update: Instant,
+    // Last RTT estimate in micro-seconds
+    last_rtt_us: Option<f64>,
+}
+
+impl RateController {
+    pub(super) fn new(
+        estimated_bitrate: u64,
```
pthatcher, line 35: I would call this the start_bitrate.

packet/src/bwe/rate_controller.rs
```diff
@@ -0,0 +1,376 @@
...
+        min_bitrate: u64,
+        max_bitrate: u64,
+        now: Instant,
```
pthatcher, line 38: As mentioned in the other PR, I think it would be valuable to create DataRate and DataSize structs to wrap bitrates and sizes to avoid mistakes with bitrate units.

packet/src/bwe/rate_controller.rs
k0nserv: Good shout 👍🏼

packet/src/bwe/rate_controller.rs
k0nserv: You're right, I wrote those methods first and then just hammered out this transition method

PR #48, opened 2023-01-26
Run `cargo clippy --fix` on Rust 1.67.0

k0nserv:


algesten:
This needs a cargo fmt I think

algesten:
Test failure?

k0nserv:
Yes, but I didn't touch that code 🤔

k0nserv:
Huh, I thought this was just a precision problem but it turns out to be a huge change

```
thread 'media::register::test::jitter_at_20' panicked at 'Expected jitter to converge at 20.0, jitter was: 9844.839', str0m/src/media/register.rs:531:9
```

Something is definitely wrong

k0nserv:
Turns out clippy's fix for `unchecked_duration_subtraction` broke some order of operations code. I fixed it in 07b6d33

algesten:
Is that clippy sane? I mean in this instance there's no way it's going to overflow.

k0nserv:
This line looks like one that might be problematic it can underflow panic if the system was started less than 60 seconds ago, on certain platforms.

```
now - Duration::from_secs(60)
```

algesten (approved):

PR #49, opened 2023-01-29
Add receive bitrate reporting to TwccSendRegister

k0nserv:


str0m/src/session.rs
```diff
@@ -192,6 +194,13 @@ impl Session {
             for m in only_media_mut(&mut self.media) {
                 m.maybe_create_regular_feedback(now, sender_ssrc, &mut self.feedback);
             }
+
+            if let Some(received_bitrate) = self
```
k0nserv, line 198: This is temporary

str0m/src/session.rs
```diff
@@ -120,7 +120,9 @@ impl Session {
             feedback: VecDeque::new(),
             twcc: 0,
             twcc_rx_register: TwccRecvRegister::new(100),
-            twcc_tx_register: TwccSendRegister::new(1000),
+            // Enough to accurately measure received bandwidths up to 20Mbit/s, assuming 1200 bytes
+            // MTU.
+            twcc_tx_register: TwccSendRegister::new(2083),
```
k0nserv, line 125: This logic and the above comment is wrong as it assumes 1200bytes average packet size. For video this might be close to the truth(a 1 Mbit/s stream usually results in around ~110 packets per second), however for audio a quick test yields more like 65.25bytes per packet. 

Taking the midpoint and weighting video up a bit because of the higher send rate a real value is more like 750bytes per packet.

Hence we'd need to keep north of 3k packets to measure with a window of 1 second, 1.5k packets for a window of half a second. I think we'll be likely to use a 0.5 second window.

Each `SendRecord` is currently 56bytes large resulting in 168kb of memory usage. I think we could be more aggressive and aim for ~1-2MB of memory usage here.

k0nserv: I updated this to use 1000 bytes as an average size for packets, this matches what I observed with Wireshark and some testing.

algesten (approved):

PR #51, opened 2023-02-01
Fix TWCC later report bug

k0nserv:
When a TWCC feedback report arrives so late that the packets it pertains
to are no longer in the TwccSendRecvRegister's queue we panic with an
assertion error.

For example, if the oldest packet in the queue has TWCC sequence number
1200 and we receive a feedback packet with a base sequence number
of 1199 an assert will trigger a panic.

### DIFF
--- a/rtp/src/rtcp/twcc.rs
+++ b/rtp/src/rtcp/twcc.rs
@@ -1115,6 +1115,12 @@ impl TwccSendRegister {
 
         Some(result.into())
     }
+
+    pub fn send_record(&self, seq: SeqNo) -> Option<&SendRecord> {
+        let index = self.queue.binary_search_by_key(&seq, |r| r.seq).ok()?;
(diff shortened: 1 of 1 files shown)


rtp/src/rtcp/twcc.rs
```diff
@@ -1049,7 +1057,7 @@ impl TwccSendRegister {
             panic!(
                 "Unexpected TWCC sequence when applying TWCC report. \
                     Send Record Seq({record_seq}) does not match Report Seq({report_seq}).\
-                    \nLast 100 entires in queue: {queue_tail:?}"
+                    \nLast 100 entires in queue: {queue_tail:?}."
```
algesten, line 1060: ah yes. that full stop 👌 

:D

algesten (approved):
Love it!

PR #53, opened 2023-02-01
feature/twcc record packet status

k0nserv:
- Record packet status in SendRecord
- Account for packet status for bitrate estimates

### DIFF
--- a/rtp/src/rtcp/twcc.rs
+++ b/rtp/src/rtcp/twcc.rs
@@ -1120,9 +1120,11 @@ impl TwccSendRegister {
             .skip_while(|record| record.seq > last.seq)
             .take_while(|record| record.remote_recv_time.map(|r| r > cutoff).unwrap_or(true))
             .filter_map(|record| {
-                // If we have this field we have handled a TWCC report for the packet.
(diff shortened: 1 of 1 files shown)


algesten (approved):
It's getting a bit state heavy. But I can clean that up later.

k0nserv:
> It's getting a bit state heavy. But I can clean that up later.

Do you mean `SendRecord` or the return type on the iterator(I was thinking that might be due to become a struct)?

algesten:
More the interaction between all the fields. The status vs instant. They both encode an implicit state that we received feedback. Would be nice to make that explicit.

k0nserv:
> More the interaction between all the fields. The status vs instant. They both encode an implicit state that we received feedback. Would be nice to make that explicit.

Ah yes I was thinking that too, something like


```rust
struct SendRecord {
  seq_no: SeqNo,
  local_send_time: Instant,
  size: u16,
  recv_report: Option<RecvReport>
}
```

algesten:
Yeah. That's nice.

k0nserv:
> can any of this go out in the stats ?
No I don't think anything here is needed for stats

> Yeah. That's nice.
I can move it to something like that next time I touch the code

PR #54, opened 2023-02-02
add stats for firs, plis, nacks, packet counts

davibe:


str0m/src/media/sender.rs
```diff
@@ -15,8 +16,15 @@ pub struct SenderSource {
     rid: Option<Rid>,
     next_seq_no: SeqNo,
     last_used: Instant,
-    bytes_tx: u64,
-    bytes_tx_resent: u64,
+    // bytes includes resent
+    bytes: u64,
+    bytes_resent: u64,
+    // packets includes resent
```
algesten, line 25: That `bytes` and `packets` also holds the resent is according to spec I assume?

Can we clarify the comment that this is according to spec? Maybe even link to the spec?

str0m/src/stats.rs
```diff
@@ -97,6 +99,8 @@ pub struct MediaIngressStats {
     pub rid: Option<Rid>,
     // total bytes received
     pub bytes: u64,
+    // total number of rtp packets received
+    pub packets: u64,
```
algesten, line 103: This commit is a bit strange in isolation since it just adds struct fields without reading or writing to them.

algesten (approved):
LGTM! Two small comments

str0m/src/media/sender.rs
davibe: correct, ok

davibe: I add the links to the stats events, not the sender/receiver because that is where the "public api" of the stats is

PR #55, opened 2023-02-03
fix how we avoid sending empty stats

davibe:
It's wrong to not count m-lines that are not sending or receiving because what we want to count is also the total peer traffic.
If we exclude a track on a change of direction the byte counts would decrease as if there was a negative traffic, which does not make sense.

With this patch we simply avoid outputting stats related to sender / receiver sources with no traffic


k0nserv (approved):
LGTM

PR #57, opened 2023-02-09
add egress RTT stats

davibe:


str0m/src/util.rs
```diff
@@ -35,6 +35,55 @@ impl Soonest for (Option<Instant>, &'static str) {
...
+        .as_nanos() as u64;
+    let mut s = u / 1_000_000_000;
+    s += 0x83AA7E80; //offset in seconds between unix epoch and ntp epoch
+    let mut f = u % 1_000_000_000;
+    f <<= 32;
+    f /= 1_000_000_000;
+    s <<= 32;
+
+    s | f
+}
```
algesten, line 85: Similar calculation here: https://github.com/algesten/str0m/blob/main/rtp/src/mtime.rs#L191-L218

Wonder if we should try and unite them?

algesten (approved):
LGTM. One question of code reuse.

str0m/src/util.rs
davibe: Good input. I have an in progress branch where I am already re-using this one. I will think about this after I finalise that one.

PR #58, opened 2023-02-09
Split Media into a public API Media<'a> and private API MLine

algesten:
The motivation for this split is that 

1. Media is getting unwieldy because it is part of the public API as well as a central role in internal.
2. #24 requires a refactoring so that Media::write can use SDP negotiated session level parameters.


str0m/src/media/transceiver.rs
```diff
@@ -0,0 +1,1073 @@
...
+    /// found in the `ssrc_info_rx`.
+    cname: String,
+
+    /// "Stream and track" identifiers.
+    ///
+    /// This is for _outgoing_ SDP. Incoming Msid details
+    /// can be found in the `ssrc_info_rx`.
+    msid: Msid,
+
+    /// Audio eller video.
```
k0nserv, line 73:
```suggestion
    /// Audio or video.
```

:D

str0m/src/media/transceiver.rs
```diff
@@ -0,0 +1,1073 @@
...
+    kind: MediaKind,
+
+    /// The extenions for this m-line.
```
k0nserv, line 76:
```suggestion
    /// The extensions for this m-line.
```

k0nserv (approved):
Direction seems good. From what I can tell there were few changes outside of the renaming.

One thing to consider is whether we want to support the sender/receiver split at some point and if that has implications here.

algesten:
> One thing to consider is whether we want to support the sender/receiver split at some point and if that has implications here.

My plan is that this `Transceiver` is never going to be public API. The way to interact with str0m media is via `Media`. By that logic it would make sense to avoid the "transceiver" word altogether. I also considered `MLine`. 

What we want is effectively the internal functions of organizing `SenderSource`, `ReceiverSource`, stats gathering, packetizer, depacketizer etc. All the constituent parts of the m-line.

I'm open to naming suggestions… something that groups these functions, something that corresponds 1-1 to an m-line. Something internal, since the corresponding external API has taken the word "media".

algesten:
Alright, I've convinced myself to go for `MLine` instead of `Transceiver`. It's not great, because it is a bit more to it than just the `m=` SDP line, but "transceiver" sends a more wrong message with regards to the spec and normal WebRTC API.

PR #60, opened 2023-02-10
use RTCP XR to calculate ingress RTT

davibe:


rtp/src/rtcp/mod.rs
```diff
@@ -202,6 +206,7 @@ impl Rtcp {
         match self {
             Rtcp::SenderReport(v) => v.reports.is_full(),
             Rtcp::ReceiverReport(v) => v.reports.is_full(),
+            Rtcp::ExtendedReport(_) => false,
```
algesten, line 209: That's a strange value.

This function is used to determine if the RTCP packets to be sent can be combined into a larger UDP packet with all RTCP reports stapled together. By saying `false` here, you effectively say that it can be combined forever together.

The only good choices are _something that checks if there is more space_, or `true` if it's impossible to know, or that the report definitely must be sent by itself.

rtp/src/rtcp/rtcpfb.rs
```diff
@@ -1,24 +1,29 @@
-use crate::{FirEntry, NackEntry, ReceptionReport, ReportList, Rtcp, Sdes, SenderInfo, Ssrc, Twcc};
+use crate::{
+    DlrrItem, FirEntry, NackEntry, ReceiverReferenceTime, ReceptionReport, ReportList, Rtcp, Sdes,
+    SenderInfo, Ssrc, Twcc, ReportBlock,
+};
```
algesten, line 4: :'(

rtp/src/rtcp/xr.rs
```diff
@@ -0,0 +1,258 @@
...
+        Ok(ExtendedReport {
+            sender_ssrc,
+            blocks,
+        })
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum ReportBlock {
+    ReceiverReferenceTime(ReceiverReferenceTime),
```
algesten, line 86: Since we have Dlrr, let's call this Rrtr.

Google does it: https://webrtc.googlesource.com/src//+/d090952628606228afa47c842812927f15227290/modules/rtp_rtcp/source/rtcp_transceiver_config.h#172

rtp/src/rtcp/rtcpfb.rs
```diff
@@ -1,24 +1,29 @@
...
+    SenderInfo(SenderInfo),                       // tx -> rx
+    ReceptionReport(ReceptionReport),             // rx -> tx
+    DlrrItem(DlrrItem),                           // rx <- tx
+    ReceiverReferenceTime((ReceiverReferenceTime, Ssrc)), // rx -> tx
+    SourceDescription(Sdes),                      // tx -> rx
+    Goodbye(Ssrc),                                // tx -> rx
+    Nack(Ssrc, ReportList<NackEntry>),            // rx -> tx
+    Pli(Ssrc),                                    // rx -> tx
+    Fir(FirEntry),                                // rx -> tx
+    Twcc(Twcc),                                   // rx -> tx
```
algesten, line 19: Hopefully this pointless whitespace change back if we call it Rrtr.

str0m/src/media/receiver.rs
```diff
@@ -152,6 +172,21 @@ impl ReceiverSource {
...
+                self.set_sender_info(now, *v);
+            }
+            DlrrItem(v) => {
+                let now = (unix2ntp(SystemTime::now()) >> 16) as u32;
+                let rtt = calculate_rtt_ms(now, v.last_rr_delay, v.last_rr_time);
+                self.rtt = rtt;
+            },
+            _ => {}
+        }
+    }
```
algesten, line 188: I prefer if we stick to splitting the RtcpFb enum on the `MLine` level and have `set_sender_info` and a new `set_dlrr_item`.

algesten (commented):
Looking promising

rtp/src/rtcp/rtcpfb.rs
davibe: `cargo fmt`

str0m/src/media/receiver.rs
davibe: I don't like 'set_dlrr_item' because it's not setting a dlrr item. I would prefer "update_with_dlrr"

rtp/src/rtcp/rtcpfb.rs
davibe: anything I should do ?

algesten (approved):
Looks good!

Does it trigger Chrome to send XR back?

algesten (changes requested):
Nope. No now()

str0m/src/media/receiver.rs
```diff
@@ -152,6 +169,12 @@ impl ReceiverSource {
         self.nacks += 1;
     }
 
+    pub fn set_dlrr_item(&mut self, dlrr: DlrrItem) {
+        let now = (unix2ntp(SystemTime::now()) >> 16) as u32;
```
algesten, line 173: Sorry. Missed it in my first review. This is never OK in str0m.

algesten (changes requested):

rtp/src/rtcp/mod.rs
algesten: Right I see what you mean. The buffer capacity is checked external from this mechanic, so `false` could work _IF_ we are allowed to merge multiple XR. Confusingly TWCC could in theory be merged, but Chrome does't accept it, so the question is whether XR would be accepted in a merged form. `true` means we don't merge.

str0m/src/media/receiver.rs
algesten: `update_with_dlrr` is fine. I only noticed the asymmetry between SenderSource and ReceiverSource now. I pushed a fix for it.

algesten (approved):
I think we're ready to merge

davibe (commented):
LGTM :)

PR #64, opened 2023-02-15
Improve Ingress NACK

k0nserv:
The previous version of ingress NACK would only generate a NACK for a
missing packet a single time and then move forward, never again
considering if this packet should be NACKed.

With this PR we change the behaviour of `ReceiveRegister` to
continue repeatedly NACKing packets until received or we decided that
we've tried too many times.

This greatly improves the stability of the resulting RTP stream at 10%
packet loss.

### DIFF
--- a/str0m/src/lib.rs
+++ b/str0m/src/lib.rs
@@ -1306,7 +1306,6 @@ impl fmt::Debug for MediaData {
             .field("rid", &self.rid)
             .field("time", &self.time)
             .field("len", &self.data.len())
-            .field("packet_range", &self.packet_range())
             .finish()
     }
 }
--- a/str0m/src/lib.rs
+++ b/str0m/src/lib.rs
@@ -1306,6 +1306,7 @@ impl fmt::Debug for MediaData {
             .field("rid", &self.rid)
             .field("time", &self.time)
             .field("len", &self.data.len())
+            .field("packet_range", &self.packet_range())
             .finish()
     }
 }
--- a/str0m/src/media/event.rs
+++ b/str0m/src/media/event.rs
@@ -1,6 +1,7 @@
 use std::time::Instant;
 
 pub use packet::RtpMeta;
+use rtp::SeqNo;
 pub use rtp::{Direction, ExtensionValues, MediaTime, Mid, Pt, Rid, Ssrc};
 pub use sdp::{Codec, FormatParams};
 
@@ -142,6 +143,16 @@ pub struct MediaData {
 
     /// The individual packet metadata that were part of making the Sample in `data`.
     pub meta: Vec<RtpMeta>,
+}
+
+impl MediaData {
+    /// The packet range this was built from.
+    pub(crate) fn packet_range(&self) -> Option<(SeqNo, SeqNo)> {
+        let first = self.meta.first()?;
+        let last = self.meta.last()?;
+
+        Some((first.seq_no, last.seq_no))
+    }
 }
 
 /// Details for an incoming a keyframe request (PLI or FIR).
--- a/str0m/src/media/event.rs
+++ b/str0m/src/media/event.rs
@@ -1,7 +1,6 @@
 use std::time::Instant;
 
 pub use packet::RtpMeta;
-use rtp::SeqNo;
 pub use rtp::{Direction, ExtensionValues, MediaTime, Mid, Pt, Rid, Ssrc};
 pub use sdp::{Codec, FormatParams};
 
@@ -143,16 +142,6 @@ pub struct MediaData {
 
     /// The individual packet metadata that were part of making the Sample in `data`.
     pub meta: Vec<RtpMeta>,
-}
-
-impl MediaData {
-    /// The packet range this was built from.
-    pub(crate) fn packet_range(&self) -> Option<(SeqNo, SeqNo)> {
-        let first = self.meta.first()?;
-        let last = self.meta.last()?;
-
-        Some((first.seq_no, last.seq_no))
-    }
 }
 
 /// Details for an incoming a keyframe request (PLI or FIR).
--- a/str0m/src/media/receiver.rs
+++ b/str0m/src/media/receiver.rs
@@ -173,8 +173,8 @@ impl ReceiverSource {
         }
     }
 
-    pub fn update_with_nack(&mut self, count: u64) {
-        self.nacks += count;
+    pub fn update_with_nack(&mut self) {
+        self.nacks += 1;
     }
 
     pub fn set_dlrr_item(&mut self, now: Instant, dlrr: DlrrItem) {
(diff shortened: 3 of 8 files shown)


str0m/src/lib.rs
```diff
@@ -1306,6 +1306,7 @@ impl fmt::Debug for MediaData {
             .field("rid", &self.rid)
             .field("time", &self.time)
             .field("len", &self.data.len())
+            .field("packet_range", &self.packet_range())
```
davibe, line 1309: this is only for debug purposes, correct ?

str0m/src/media/register.rs
```diff
@@ -261,47 +328,57 @@ impl ReceiverRegister {
...
+            nacks.push(NackEntry {
                 pid: (first % u16::MAX as u64) as u16,
                 blp: bitmask,
-            }
-            .into(),
-        })
+            });
+        }
+
+        let reports = ReportList::lists_from_iter(nacks).into_iter();
```
k0nserv, line 371: I would like `lists_from_iter` to return another iterator, this is possible with `itertools`'s `chunks` but I didn't feel worth it to add a dependency just for that

str0m/src/media/register.rs
```diff
@@ -399,6 +480,21 @@ impl TimePoint {
     }
 }
 
+/// The max number of NACKs we perform for a single packet:
+const MAX_NACKS: u8 = 5;
```
k0nserv, line 484: Open to debate on this, I just picked something

str0m/src/media/receiver.rs
```diff
@@ -171,8 +173,8 @@ impl ReceiverSource {
         }
     }
 
-    pub fn update_with_nack(&mut self) {
-        self.nacks += 1;
+    pub fn update_with_nack(&mut self, count: u64) {
```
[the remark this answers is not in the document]
davibe: yes it's what goes out in the stats. the count of nacks. I think it's correct.

str0m/src/lib.rs
algesten, line 1309: `MediaData` is public API, and it's a bit strange to include the `pub(crate)` result here.

str0m/src/media/event.rs
```diff
@@ -144,6 +145,16 @@ pub struct MediaData {
     pub meta: Vec<RtpMeta>,
 }
 
+impl MediaData {
+    /// The packet range this was built from.
+    pub(crate) fn packet_range(&self) -> Option<(SeqNo, SeqNo)> {
```
algesten, line 150: It's impossible to create a `MediaData` without any underlying RTP packets. We are making that assumption elsewhere, so we should be consistent and unwrap to avoid `Option` here too.

str0m/src/media/register.rs
davibe: slice has chunks(), would that help ?

k0nserv: Hmmm I wanted to take `IntoIterator` as the argument but since I happen to have a vec here maybe I should just use that

str0m/src/media/register.rs
algesten: I'll check chrome tonight.

str0m/src/media/receiver.rs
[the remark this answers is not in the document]
k0nserv: > Count the total number of Negative ACKnowledgement (NACK) packets, as defined in [[RFC4585](https://www.w3.org/TR/webrtc-stats/#bib-rfc4585)] section [6.2.1](https://www.rfc-editor.org/rfc/rfc4585#section-6.2.1), sent by this receiver.


I think we should actually keep what we had before, a NACK packet with several reports is still a single packet I believe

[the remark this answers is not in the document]
davibe: Right, that would also be consistent with what we do at egress
https://github.com/algesten/str0m/blob/2363d3dd4b953e66e7eb8c6f1a8b776c6391b522/str0m/src/media/mline.rs#L746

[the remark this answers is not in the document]
k0nserv: Cool, reverting

str0m/src/session.rs
```diff
@@ -416,8 +416,8 @@ impl Session {
             if rid.is_none() && repaired_source.rid().is_some() {
                 rid = repaired_source.rid();
             }
+            let orig_seq_no = repaired_source.update(now, &header, clock_rate);
             let source = m_line.get_or_create_source_rx(ssrc);
-            let orig_seq_no = source.update(now, &header, clock_rate);
```
algesten, line 420: This is not right. It's the main source that has the register for what is received and not. So `update()` must be on the main source.

It also returns the lengthened `orig_seq_no`. However you can't lengthen with the repaired sequence. It must be the real source (given that we set `header.sequence_number = orig_seq_16` higher up).

algesten (changes requested):
One problem, the rest small things.

str0m/src/lib.rs
k0nserv: Good point, do you think we should make `packet_range` public or remove it from debug output?

str0m/src/media/event.rs
k0nserv: Will action pending outcome on https://github.com/algesten/str0m/pull/64#discussion_r1107334098

str0m/src/lib.rs
algesten: In some respects I want consumers to stop thinking about the RTP layer :) 

So I favor not having it.

str0m/src/session.rs
k0nserv: That's what's now happening, before it wasn't correctly doing it on the main source.

`ssrc` here is the value from the header, which for a repaired packet is going to be the repair stream. The non-RTX case is handled on line 359. `repaired_ssrc` is the SSRC stream for the stream we are repairing

str0m/src/lib.rs
k0nserv: Aight, I found it very useful for debugging str0m itself but can always add it back on a need by need basis

str0m/src/session.rs
algesten: Ah. Ok. As you were. :)

algesten (approved):

str0m/src/media/register.rs
k0nserv: https://webrtc.googlesource.com/src/+/2e540a28c05104a4af4d44c978d59344e3cb53f7/modules/video_coding/nack_requester.cc#29

algesten: Ok. But do they have a timer as well, so defacto they don't get all the way to 10?

PR #69, opened 2023-02-17
stats: only non-rtx sources contribute to rtt and loss

davibe:


algesten (approved):
LGTM. I wouldn't know whether this is correct or not :P

PR #70, opened 2023-02-20
Fix TwccIter delta accumulation

k0nserv:
The delta values in a TWCC report should accumulate from one packet
status to the next when iterating through them, this wasn't the case.

### DIFF
--- a/rtp/src/rtcp/twcc.rs
+++ b/rtp/src/rtcp/twcc.rs
@@ -99,6 +99,9 @@ impl Iterator for TwccIter {
             _ => unreachable!(),
         };
 
+        if let Some(new_timebase) = instant {
+            self.time_base = new_timebase;
+        }
         let seq: SeqNo = (self.base_seq + self.index as u64).into();
 
         self.index += 1;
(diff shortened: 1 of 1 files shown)


davibe (approved):

rtp/src/rtcp/twcc.rs
```diff
@@ -1676,4 +1679,94 @@ mod test {
...
+        let twcc = Twcc {
+            sender_ssrc: 0.into(),
+            ssrc: 0.into(),
+            base_seq: 1,
+            status_count: 12,
+            reference_time: 1337,
+            feedback_count: 0,
+            chunks: [
+                PacketChunk::Run(PacketStatus::ReceivedSmallDelta, 2),
+                PacketChunk::VectorDouble(0b1101_0010_0101_0000, 7),
```
k0nserv, line 1694: Would like to include `PacketChunk::VectorSingle` here too, but I'm not sure how that should work. Does a `1` bit mean `ReceivedSmallDelta` in that case?

algesten (approved):

PR #71, opened 2023-02-23
egress loss: average stats from intermediate receiver reports

davibe:
Before this commit we were calculating stats from latest RR received. Turns out browsers send RRs much more frequently especially in case of network problems (10+ per second). Downsampling to the latest every second looses a lot of information considering that each RR reports the fraction lost from the previous one.

With this patch we are averaging the values weighted against the amount of packet seen by the receiver, which makes much more sense in the case of a periodic update.

### DIFF
--- a/str0m/src/media/sender.rs
+++ b/str0m/src/media/sender.rs
@@ -1,8 +1,8 @@
 use std::time::Instant;
 
 use rtp::{
-    Descriptions, InstantExt, MediaTime, Mid, ReceptionReport, ReportList, Rid, Sdes, SdesType,
-    SenderInfo,
+    extend_seq, Descriptions, InstantExt, MediaTime, Mid, ReceptionReport, ReportList, Rid, Sdes,
+    SdesType, SenderInfo,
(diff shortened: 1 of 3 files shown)


str0m/src/media/sender.rs
```diff
@@ -148,15 +149,46 @@ impl SenderSource {
         let rtt = calculate_rtt_ms(ntp_time, r.last_sr_delay, r.last_sr_time);
         self.rtt = rtt;
 
-        self.loss = Some(r.fraction_lost as f32 / u8::MAX as f32);
+        let ext_seq = {
+            let ext = (r.max_seq >> 16) as u64;
+            let num = (r.max_seq & 0xffff) as u16;
+            extend_seq(Some(ext), num)
```
algesten, line 155: The `Some(ext)` part here must be saved on ReceiverSource and reused for every incoming RR.

algesten (changes requested):

algesten (approved):

PR #73, opened 2023-02-27
add Vp8 codec extra

davibe:
Example use case: A user (of the lib) can decide to
- avoid packaging discardable frames to save bandwidth
- switch off VP8 layer 1 to save bandwidth

Internally str0m could use such information for an internal congestion control to compensate the overhead cause by answering Nacks


algesten (approved):
LGTM!

PR #81, opened 2023-03-08
avoid too many retransmissions

davibe:
I have observed in case of congestion we can end up retransmitting almost 100% of the packets which leads to more congestion and kills the client (not only the media stream) and does not leave space for a smooth recover


str0m/src/util/value_history.rs
```diff
@@ -0,0 +1,80 @@
...
+    /// Adds a timed value
+    /// Note: time should always monotonically increase in subsequent calls to add()
+    pub fn add(&mut self, t: Instant, v: T) {
+        self.value += v;
+        self.history.push_back((t, v));
+        self.drain(t);
+    }
+
+    /// Returns the sum of the values more recent than given Instant
+    pub fn sum_from(&self, t: Instant) -> T {
```
algesten, line 38: Maybe `sum_since`?

str0m/src/util/value_history.rs
```diff
@@ -0,0 +1,80 @@
...
+        ValueHistory {
+            value: initial,
+            history: VecDeque::new(),
+            max_time,
+        }
+    }
+
+    /// Adds a timed value
+    /// Note: time should always monotonically increase in subsequent calls to add()
+    pub fn add(&mut self, t: Instant, v: T) {
```
algesten, line 31: Maybe `push` to be consistent with `Vec` and avoiding confusion with "adding numbers together"?

str0m/src/media/mline.rs
```diff
@@ -399,7 +406,21 @@ impl MLine {
...
             };
 
+            let from = now.checked_sub(Duration::from_secs(1)).unwrap_or(now);
+            let bytes_transmitted = self.bytes_transmitted.sum_from(from);
+            let bytes_retransmitted = self.bytes_retransmitted.sum_from(from);
+            let ratio =
+                bytes_retransmitted as f32 / (bytes_retransmitted + bytes_transmitted) as f32;
+            let ratio = if ratio.is_finite() { ratio } else { 0_f32 };
+
+            if ratio > 0.15_f32 {
```
algesten, line 416: Do we have a reason for 0.15 here?

algesten (changes requested):

str0m/src/media/mline.rs
davibe: Not a strong one. I tested with high values of loss and congestion and 0.1 was fine, only rarely there was some small image corruption. Then I increased to 0.15 to be safer.

str0m/src/util/value_history.rs
davibe: sure

str0m/src/util/value_history.rs
davibe: ok

str0m/src/media/mline.rs
algesten: Alright. Let's try and see!

PR #82, opened 2023-03-13
Implement padding support in LeakyBucketPacer

k0nserv:


str0m/src/media/mline.rs
```diff
@@ -1146,6 +1185,111 @@ impl From<MediaType> for MediaKind {
...
+                &mut buf[..],
+                header_len,
+                body_len + original_seq_len,
+                SRTP_BLOCK_SIZE,
+            );
+
+            buf.truncate(header_len + body_len + original_seq_len + pad_len);
+        }
+        NextPacket::Padding { len, .. } => {
+            let rounded_len = if len as usize % SRTP_BLOCK_SIZE == 0 {
```
k0nserv, line 1272: Maybe generating this padding is a bit to low-level for this crate. Perhaps we should have

```rust
RtpHeader::generate_padding_packet(&mut buf[..], len, SRTP_BLOCK_SIZE)
```

or something like it

str0m/src/media/mline.rs
```diff
@@ -146,12 +149,22 @@ pub(crate) struct MLine {
...
+        seq_no: SeqNo,
+        orig_seq_no: Option<SeqNo>,
+    },
+    /// An empty padding packet to be generated.
+    Padding {
+        pt: Pt,
+        ssrc: Ssrc,
+        seq_no: SeqNo,
+        len: u8,
+    },
```
algesten, line 167: I suggest making this enum like so:

```rs
enum NextPacket<'a> {
    Regular {
        ident: NextPacketIdent,
        pkt: &'a Packetized,
    },
    Resend {
        ident: NextPacketIdent,
        pkt: &'a Packetized,
        orig_seq_no: SeqNo,
    },
    Padding {
        ident: NextPacketIdent,
        len: u8,
    }
}
```

str0m/src/media/mline.rs
```diff
@@ -446,6 +410,81 @@ impl MLine {
...
     }
 
+    pub fn create_padding_packet(
+        &mut self,
+        target_size: DataSize,
+        now: Instant,
+        exts: &Extensions,
+        twcc: &mut u64,
+        buf: &mut Vec<u8>,
+    ) -> Option<(RtpHeader, SeqNo)> {
```
algesten, line 420: Rather than this being a `pub fn` by itself, I like the code path to go via `poll_packet`. In Session, you already have the distinction that "if there is a padding length, do this, otherwise do that". We just move that down one level, so `poll_packet` gets another argument `pad_len: Option<u8>`.

Then we don't need to break out `finalize_packet`, but keep that code intact (less git diff).

Finally, yes, as you say, i think the padding packet creation should be pushed down into the `rtp` subcrate.

algesten (changes requested):
So exciting!!!!

str0m/src/media/mline.rs
k0nserv: Ah, that's nice

k0nserv: I did this but slightly differently

str0m/src/media/mline.rs
```diff
@@ -319,11 +330,14 @@ impl MLine {
         now: Instant,
         exts: &Extensions,
         twcc: &mut u64,
+        pad_size: Option<DataSize>,
         buf: &mut Vec<u8>,
     ) -> Option<(RtpHeader, SeqNo)> {
         let mid = self.mid;
 
-        let next = if let Some(next) = self.poll_packet_resend(now) {
+        let next = if let Some(next) = pad_size.and_then(|p| self.create_padding_packet(p, now)) {
```
algesten, line 344: I prefer this to be isolated out to `if let Some(next) = poll_packet_padding(now, pad_size)` to be uniform with the below.

str0m/src/media/mline.rs
```diff
@@ -367,23 +380,40 @@ impl MLine {
...
+                    SRTP_BLOCK_SIZE,
+                );
+                if len == 0 {
+                    return None;
+                }
 
-        buf.truncate(header_len + body_len + original_seq_len + pad_len);
+                buf.truncate(header_len + len);
+            }
+        }
```
algesten, line 416: Maybe move `buf.truncate`  to below and let the whole `match` return the body len

```rs
let body_len = match next.body {
 ...
}

buf.truncate(header_len + body_len);
```

str0m/src/media/mline.rs
```diff
@@ -460,10 +489,76 @@ impl MLine {
...
+                .iter()
+                .map(|(pt, buffer)| (pt, buffer.history_size()))
+                .filter(|(_, size)| *size > 0)
+                .collect();
+            vec.sort_by_key(|(_, size)| *size);
+
+            // Use the last PT i.e. the one that has the most RTX history, this is a poor
+            // approximation for most recent sends.
+            vec.last().map(|(pt, _)| **pt)
+        };
```
algesten, line 511: This looks like we can use `max_by_key` to  simplify.

str0m/src/media/mline.rs
```diff
@@ -460,10 +489,76 @@ impl MLine {
...
+                        orig_seq_no: Some(packet.seq_no.expect(
+                            "this packet to have been sent and therefore have a sequence number",
+                        )),
+                    },
+                });
+            }
+        }
+
+        // NB: If we cannot generate padding here for some reason we'll get stuck forever.
+        let pt_to_use = self.buffers_tx.keys().next()?;
```
algesten, line 545: Feels like we should be able to merge this into the above case and not have `if let Some(pt_to_use) = pt_to_use` above.

algesten (changes requested):
Some nits

str0m/src/media/mline.rs
k0nserv: Sounds good 👍🏼

str0m/src/media/mline.rs
k0nserv: I'll check

str0m/src/media/mline.rs
k0nserv: Sure 👍🏼

str0m/src/media/mline.rs
k0nserv: The branches differ not just in the Pt that gets used but also in the kind of padding packet. The first case does a resend of a previous packet, this part generates an empty padding packet

algesten: Ah yes. Spurious resends :D

algesten (approved):

PR #84, opened 2023-03-14
Tidy up padding packets

algesten:


packet/src/buffer_tx.rs
```diff
@@ -156,7 +156,24 @@ impl PacketizingBuffer {
 
     /// An iterator over all packets that are queued, but have not yet been sent.
     fn queued_packets(&self) -> impl Iterator<Item = &Packetized> {
-        (self.emit_next..).map_while(|idx| self.queue.get(idx))
+        // seq_no is some as long as packet has not been sent
+        self.queue.iter().skip_while(|p| p.seq_no.is_some())
+    }
+
+    /// Find a historic packet that is smaller than the given max_size.
+    pub fn historic_packet_smaller_than(&self, max_size: usize) -> Option<&Packetized> {
```
algesten, line 164: Should this have a max cap of how far back we allow it to go?

k0nserv, line 164: This differs from libWebRTC. They simply use the last RTX packet, they do have code paths that cause it to behave in a more round robin fashion. Unsure if this is significant

algesten: > They simply use the last RTX packet, they do have code paths that cause it to behave in a more round robin fashion

So what's the round robin code path like?

k0nserv: Every time a packet is used for padding a counter is incremented on it, then they use this counter to send the packet with the lowest value(using a priority queue I think)

PR #86, opened 2023-03-16
Delay based BWE

k0nserv:
Implements a delay based bandwidth estimator(googcc) inspired by libWebRTC's implementation.

## Status of the PR

There are some failing/commented out tests and a bit of experimental code still. I aim to clean it up further, but it's ready for a first review pass.

⚠️ The details of exactly how the padding generation works in `LeakyBucketPacer` need to revisited and the code needs to be cleaned up. I have several iterations of thinking in there at the moment, feel free to  ignore it

## How it works:

We send packets onto the network, paced if necessary. When we receive back TWCC reports for these packets we form Packet Groups based on packets being sent as a burst and a few other criteria. Based on these packet groups we can calculate the inter-group delay variation i.e. the difference in the differences between send times and arrival times. This delay lets us understand if packets are spending an increasing amount of time queued on their way to the receiver, indicating overuse. 

To dampen the impact of network jitter linear regression is used to estimate the delay trend. Further, to reduce the impact of other loss based protocols running in parallel(e.g. TCP) an adaptive delay threshold is employed. 

The existing rate controller is used to modulate our bitrate estimates based on the output of the delay regressions. 

Finally we use padding packets to probe the available bitrate e.g. if we are sending media at 250Kbit/s and the estimated bitrate is 680Kbit/s we'll fill in the difference between the media rate and the estimate with padding to ensure we discover overuse if we are hitting the link limit.

## Source files:

* `src/packet/bwe/arrival_group.rs` Everything related to arrival groups, calculating delays, constructing groups etc.
* `src/packet/bwe/macros.rs` Macros that report various statistics which, when visualized, are useful to understand what is happening within the BWE system.
* `src/packet/bwe/trendline_estimator.rs` Implements the linear regression of inter-group delay variations as well as the overuse detection and adaptive threshold.
* `src/packet/bwe/mod.rs` The BWE system itself which utilises the previously described components to implement the estimate.

## Public API

There are a few changes to the public API.

### `RtcConfig::use_bwe`

Enables/disables the BWE system. BWE is turned off by default and can be enabled as follows:

```rust
Rtc::builder().use_bwe(true)
```

### `Rtc::configure_bwe`

A method to configure BWE. See the documentation for details of its use.

### `Event::EgressBitrateEstimate`

A periodic event emitted when a new estimate for the egress bitrate is available.

### DIFF
--- a/src/rtp/bandwidth.rs
+++ b/src/rtp/bandwidth.rs
@@ -19,12 +19,12 @@ impl Bitrate {
         Bitrate(bps as f64)
     }
 
-    /// Create a bitrate of some **Kilobit** per second(kbps).
+    /// Create a bitrate of some **Kilobits** per second(kbps).
     pub const fn kbps(kbps: u64) -> Self {
         Self::bps(kbps * 10_u64.pow(3))
     }
 
-    /// Create a bitrate of some **Megabit** per second(mbps).
+    /// Create a bitrate of some **Megabits** per second(mbps).
     pub const fn mbps(mbps: u64) -> Self {
         Self::bps(mbps * 10_u64.pow(6))
     }


src/session.rs
```diff
@@ -653,16 +687,50 @@ impl Session {
...
+                }
+            }
+            let kind = if pad_size.is_some() && queued_at.is_none() {
+                "padding"
+            } else {
+                "media"
+            };
+
+            {
+                use std::time::SystemTime;
```
k0nserv, line 712: This should also be removed before merging, again it's useful though

src/session.rs
```diff
@@ -769,6 +837,32 @@ impl Session {
...
+        desired_birrate_bps: u64,
+        now: Instant,
+    ) {
+        const PACING_FACTOR: f64 = 2.5;
+        const PADDING_FACTOR: f64 = 0.7;
+
+        if let Some(bwe) = &mut self.bwe {
+            let pacing_rate = (current_bitrate_bps as f64 * PACING_FACTOR).round() as u64;
+            let padding_rate = match bwe.last_estimate() {
+                // If the estimate exceeds the desired bitrate we don't need to use probing to
```
k0nserv, line 853: Grammar

src/lib.rs
```diff
@@ -1153,6 +1159,45 @@ impl Rtc {
...
         self.session.m_line_by_index_mut(index)
     }
+
+    /// Configure the BWE system.
+    ///
+    /// Configure the bandwidth estimation system with the current bitrate and the desired maximum bitrate.
+    /// **Note:** This only has an effect if BWE has been enabled via `RtcConfig::use_bwe`.
+    ///
+    /// * `current_bitrate` A best estimate of the bitrate that's currently being sent. The more
+    /// accurate this value is the better bandwidth estimation performs. Only video bitrate is
```
k0nserv, line 1169: Grammar(comma after "is")

src/media/sender.rs
```diff
@@ -34,6 +34,7 @@ pub(crate) struct SenderSource {
     rtt: Option<f32>,
     // losses collecter from RR (known packets, lost ratio)
     losses: Vec<(u64, f32)>,
+
```
k0nserv, line 37: Remove this

src/packet/bwe/macros.rs
k0nserv, line 1: I'd like to keep this macros, I'm sure they'll be useful for debugging. 


They should be enabled by a runtime or buildtime flag

src/packet/bwe/rate_control.rs
```diff
@@ -34,18 +37,18 @@ struct RateControl {
...
     averaged_observed_bitrate: MovingAverage,
     /// The last time we updated the estimated bitrate.
-    last_estimate_update: Instant,
+    last_estimate_update: Option<Instant>,
     // Last RTT estimate in micro-seconds
-    last_rtt_us: Option<f64>,
+    last_rtt: Option<Duration>,
+
+    is_probing: bool,
+    stoppped_probing_at: Option<Instant>,
```
k0nserv, line 45: The need for this needs to be revisited, it was a bit of a hack when I wasn't correctly generating padding near the estimated rate. Shouldn't be required anymore

src/packet/bwe/rate_control.rs
```diff
@@ -95,34 +100,115 @@ impl RateControl {
...
+            let increase = ((eta - 1.0) * self.estimated_bitrate.as_f64()).max(1_000.0);
+
+            self.estimated_bitrate.as_f64() + increase
         };
-        let max = self
-            .last_observed_bitrate
-            .map(|r| r.as_f64() * MAX_ESTIMATE_RATIO);
+        let max = observed_bitrate.as_f64() * MAX_ESTIMATE_RATIO;
+
+        // Give ourselves a bit of leeway to establish the new bitrate when we stop probing.
```
k0nserv, line 158: This shouldn't happen anymore now that we properly generate padding at a rate near the estimate, remove

src/packet/bwe/rate_control.rs
```diff
@@ -274,96 +421,99 @@ mod test {
...
-            assert_eq!(rate_controller.estimated_bitrate().as_u64(), 66326);
-
-            rate_controller.update_delay(Signal::Normal, now + duration_ms(3500));
-            assert_eq!(rate_controller.estimated_bitrate().as_u64(), 66326);
-
-            // NB: Additive increase because we are nearing convergence
-            rate_controller.update_delay(Signal::Normal, now + duration_ms(3550));
-            assert_eq!(rate_controller.estimated_bitrate().as_u64(), 67326);
-        }
+        // #[test]
```
k0nserv, line 436: Fix these tests again

src/packet/bwe/trendline_estimator.rs
```diff
@@ -0,0 +1,355 @@
...
+    overuse: Option<Overuse>,
+
+    /// The last time we updated the adaptive threshold.
+    last_threshold_update: Option<Instant>,
+
+    /// Our current hypothesis about the bandwidth usage.
+    hypothesis: BandwithUsage,
+}
+
+impl TrendlineEstimator {
```
k0nserv, line 53: Most of this is a direct port from libWebRTC. See https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/trendline_estimator.cc

src/packet/bwe/trendline_estimator.rs
```diff
@@ -0,0 +1,355 @@
...
+    count: usize,
+    time_overusing: f64,
+}
+
+#[cfg(test)]
+mod test {
+    use std::time::{Duration, Instant};
+
+    use super::{InterGroupDelayVariation, TrendlineEstimator};
+    // TODO: Fix tests
```
k0nserv, line 292: Fix the tests

src/packet/pacer.rs
k0nserv, line 1: I need to review the logic to generate padding here. I suspect it can be simpler

k0nserv (commented):
I'm a bit concerned about how resending packets for padding might enable replay attacks. The sequence number is part of the encrypted payload so perhaps this is fine tho

src/session.rs
```diff
@@ -57,6 +59,8 @@ pub(crate) struct Session {
     twcc_rx_register: TwccRecvRegister,
     twcc_tx_register: TwccSendRegister,
 
+    bwe: Option<SendSideBandwithEstimator>,
```
k0nserv, line 62: Could also have

```rust
trait Bwe {}
enum BweImpl {
  SendSide(SendSideBandwithEstimator),
  NullBwe,
}
```

Instead of this optional business.

src/lib.rs
```diff
@@ -1153,6 +1159,45 @@ impl Rtc {
...
+    /// relevant here, audio can be ignored.
+    /// * `desired_bitrate` The bitrate that is the desired target where BWE should stop probing
```
pthatcher, line 1171: Would it make more sense to call this max_bitrate?

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///
+    /// * [`Belongs::Belongs`] if the packet belongs to the group and was added.
+    /// * [`Belongs::NewGroup`] if the packet is the start of a new group and thus wasn't added.
+    ///                         The caller should create a new group and add this packet to the
+    ///                         group.
+    /// * [`Belongs::Skipped`] if the packet was skipped i.e. because it was out of order.
+    fn add_packet(&mut self, packet: AckedPacket) -> Belongs {
+        let belongs = self.belongs_to_group(&packet);
+
+        if belongs == Belongs::NewGroup || belongs == Belongs::Skipped {
```
pthatcher, line 33: Could a let-else guard be used here to let the type system know it's Belongs::Belongs and thus you will catch an unchecked state if a new state is added?

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+    /// Maybe add a packet to the group.
+    ///
+    /// ## Return
+    ///
+    /// * [`Belongs::Belongs`] if the packet belongs to the group and was added.
+    /// * [`Belongs::NewGroup`] if the packet is the start of a new group and thus wasn't added.
+    ///                         The caller should create a new group and add this packet to the
+    ///                         group.
+    /// * [`Belongs::Skipped`] if the packet was skipped i.e. because it was out of order.
+    fn add_packet(&mut self, packet: AckedPacket) -> Belongs {
```
pthatcher, line 30: The only place I see where this return value is used, it's just checking if a new group was created.  So... why not simplify this down to just that?

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+                // Out of order
+                return Belongs::Skipped;
+        };
+
+        if send_diff < BURST_TIME_INTERVAL {
+            // Sent within the same burst interval
+            return Belongs::Belongs;
+        }
+
+        let inter_arrival_time = packet
```
pthatcher, line 70: "remove_recv_diff" might match the naming with "send_diff" better.

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+            .remote_recv_time
+            .checked_duration_since(self.remote_recv_time());
+
+        let Some(inter_arrival_time) = inter_arrival_time else {
+            warn!("Out of order arrival");
+            return Belongs::Skipped;
+        };
+
+        let inter_group_delay_variation = inter_arrival_time.as_secs_f64()
```
pthatcher, line 79: Same here: I'm not a fan of using .expect() on .local_send_time.

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+            - (packet.local_send_time - self.local_send_time()).as_secs_f64();
```
pthatcher, line 80: Might be more readable to have a make a "send_diff" variable for "packet.local_send_time - self.local_send_time()" and to use checked diff for it.

I'm not sure how to name it vs the other send_diff, though.

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+    ///                         group.
+    /// * [`Belongs::Skipped`] if the packet was skipped i.e. because it was out of order.
+    fn add_packet(&mut self, packet: AckedPacket) -> Belongs {
+        let belongs = self.belongs_to_group(&packet);
+
+        if belongs == Belongs::NewGroup || belongs == Belongs::Skipped {
+            return belongs;
+        }
+
+        self.remote_recv_time = Some(packet.remote_recv_time);
```
pthatcher, line 37: I think it would make sense to rename remote_recv_time to indicate it's the latest remove recv time.    last_remote_recv_time?

Along with that, you are making sure that these is always the largest/latest time value even if the packets are added out of order?

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        if self.first_local_send_time.is_none() {
+            self.first_local_send_time = Some(packet.local_send_time);
+            self.first_seq_no = Some(packet.seq_no);
```
pthatcher, line 41: It seems like putting the (send_time, seq_no) together under "first" with one optional would make sense.

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        self.local_send_time = self.local_send_time.max(Some(packet.local_send_time));
+        self.size += 1;
+        self.last_seq_no = Some(packet.seq_no);
```
pthatcher, line 46: Is true if packets are added out of order?

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    /// Calculate the inter group delay variation between self and a subsequent group.
+    pub(super) fn inter_group_delay_variation(&self, other: &Self) -> f64 {
+        let arrival_variation = self.arrival_variation(&other);
+        let departure_variation = self.departure_variation(&other);
+
+        assert!(arrival_variation >= 0.0);
+
+        let result = arrival_variation - departure_variation;
+        let first_seq_no = self.first_seq_no.unwrap();
```
pthatcher, line 94: I think it would make sense to put this behind an "if let".  Better to fail to trace than panic I think, if you mess up.

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        result
+    }
+
+    pub(super) fn departure_variation(&self, other: &Self) -> f64 {
+        other
+            .local_send_time()
+            .checked_duration_since(self.local_send_time())
+            .unwrap_or_else(|| {
+                panic!(
```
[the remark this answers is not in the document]
pthatcher: +1

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let result = arrival_variation - departure_variation;
+        let first_seq_no = self.first_seq_no.unwrap();
+        let last_seq_no = self.last_seq_no.unwrap();
+        trace!("Delay variation for group({first_seq_no}..={last_seq_no}. {result} = {arrival_variation} - {departure_variation}");
+
+        result
+    }
+
+    pub(super) fn departure_variation(&self, other: &Self) -> f64 {
```
pthatcher, line 101: In one place you have "send_time" and in the other "departure".  I'd suggest picking one.  And if I had to pick, I'd say make this "local_send_time_variation".

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+                panic!(
+                    "other.departure_time() = {:?} to be later than self.departure_time() = {:?}",
+                    other.local_send_time(),
+                    self.local_send_time()
+                )
+            })
+            .as_millis() as f64
+    }
+
+    fn arrival_variation(&self, other: &Self) -> f64 {
```
pthatcher, line 115: Same here: pick a naming scheme, perhaps "remote_recv_time_variation".

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+                    other,
+                    self,
+                )
+            }) .as_millis() as f64
+    }
+
+    /// The local send time i.e. departure time, for the group.
+    ///
+    /// Panics if the group doesn't have at least one packet.
+    fn local_send_time(&self) -> Instant {
```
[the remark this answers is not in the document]
pthatcher: +1

src/packet/bwe/trendline_estimator.rs
```diff
@@ -0,0 +1,355 @@
...
+    /// Our current hypothesis about the bandwidth usage.
+    hypothesis: BandwithUsage,
+}
+
+impl TrendlineEstimator {
+    pub(super) fn new(window_size: usize) -> Self {
+        Self {
+            window_size,
+            zero_time: None,
+            history: VecDeque::default(),
```
xnorpx, line 58: with_capacity(window_size) ?

src/lib.rs
```diff
@@ -1153,6 +1159,45 @@ impl Rtc {
     fn m_line_mut(&mut self, index: usize) -> &mut MLine {
         self.session.m_line_by_index_mut(index)
     }
+
+    /// Configure the BWE system.
+    ///
+    /// Configure the bandwidth estimation system with the current bitrate and the desired maximum bitrate.
+    /// **Note:** This only has an effect if BWE has been enabled via `RtcConfig::use_bwe`.
+    ///
+    /// * `current_bitrate` A best estimate of the bitrate that's currently being sent. The more
```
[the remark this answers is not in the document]
k0nserv: That's now how we are thinking of it, you'd be expected to call `configure_bwe` repeatedly for example when adding new egress tracks or switching simulcast layers.

We do something like this and call configure_bwe fairly frequently


```rust
let active_bitrates = egress_tracks.iter().map(|t| t.active_layer_bitrate).sum();
# All tracks using the highest simulcast layer
let desired_bitrate = egress_tracks.iter().map(|t| t.max_layer_bitrate).sum();


rtc.configure_bwe(active_bitrates, desired_bitrate)
```

src/packet/bwe/arrival_group.rs
k0nserv: That's what I had initially but then I created the enum to also support communicating something being skipped. That said you are right that this method, as opposed to `ArrivalGroup::belongs_to_group`, can return a `bool`

src/packet/bwe/arrival_group.rs
k0nserv: Hmm, I had this before but I think I might have regressed it in some iteration of the code.

src/packet/bwe/arrival_group.rs
k0nserv: Yeah true they do go together

src/packet/bwe/arrival_group.rs
k0nserv: It wouldn't be, can it happen though? This derives from TWCC reports which are always ordered, I guess individual reports could arrive out of order though

In any case this and first_seq_no are only used for logging, still worth fixing.

src/packet/bwe/arrival_group.rs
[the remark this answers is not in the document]
k0nserv: Regarding the expects, I have a todo to get rid of them, just had them this way to detect any issues.

k0nserv: Depends on that other decision with regards to which naming convention we use here, if we do decided to stick with recv and send then I agree

src/packet/bwe/arrival_group.rs
k0nserv: I'll have a think

src/packet/bwe/arrival_group.rs
```diff
@@ -0,0 +1,309 @@
...
+
+        let Some(inter_arrival_time) = inter_arrival_time else {
+            warn!("Out of order arrival");
+            return Belongs::Skipped;
+        };
+
+        let inter_group_delay_variation = inter_arrival_time.as_secs_f64()
+            - (packet.local_send_time - self.local_send_time()).as_secs_f64();
+
+        return (inter_group_delay_variation < 0.0 && inter_arrival_time < BURST_TIME_INTERVAL)
```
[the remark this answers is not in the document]
k0nserv: hah, fair

src/packet/bwe/arrival_group.rs
k0nserv: Yes, agree. I'm going to review all the unwraps, panics, and expects here

src/packet/bwe/arrival_group.rs
k0nserv: Yup, I raised this inconsistency elsewhere.

The spec uses "departure variation" but libWebRTC uses "send diff" or "send delta". Then the question is still "local send" vs just "send", i.e. is it worth using the "local" and "remote" prefixes for clarity

src/packet/bwe/trendline_estimator.rs
k0nserv: I suppose, I was thinking we should have a type `LimitedHistory` or something that is a FIFO queue with a builtin limit, the pattern is quite common in the codebase

src/packet/bwe/macros.rs
algesten: Can you do them as a hidden feature flag?

src/session.rs
```diff
@@ -4,10 +4,12 @@ use std::time::{Duration, Instant};
 use crate::dtls::KeyingMaterial;
 use crate::io::{DatagramSend, DATAGRAM_MTU, DATAGRAM_MTU_WARN};
 use crate::media::{App, CodecConfig, MediaAdded, MediaChanged, Source};
-use crate::packet::{NullPacer, Pacer, PacerImpl, PollOutcome, RtpMeta};
+use crate::packet::{
+    LeakyBucketPacer, NullPacer, Pacer, PacerImpl, PollOutcome, RtpMeta, SendSideBandwithEstimator,
+};
```
algesten, line 9: :'(

src/session.rs
algesten: Is "SendSide" a good name here? I understand this is as opposed to REMB.

@pthatcher thoughts?

algesten (commented):
Looks great!

Some things I picked up on.

Overall I think we need to think about accuracy of state. As I said in a comment, I prefer panics over sweeping unexpected state under the rug.

src/lib.rs
```diff
@@ -299,6 +299,9 @@ pub enum Event {
 
     /// Aggregated statistics for each media (mid, rid) in the egress direction
     MediaEgressStats(MediaEgressStats),
+
+    /// A new estimate from the bandwidth estimation subsystem.
+    EgressBitrateEstimate(u64),
```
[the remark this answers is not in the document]
k0nserv: Aight, I can make this `Bitrate` then

src/lib.rs
```diff
@@ -1153,6 +1159,45 @@ impl Rtc {
...
+    /// rtc.configure_bwe(750_000, 1_500_000)
+    /// ````
+    ///
+    /// As mentioned above this works better if the current bitrate matches closely what you are
+    /// actually intending to send. This can be achieved by measuring the observed ingress bitrate
+    pub fn configure_bwe(
+        &mut self,
+        current_bitrate_bps: u64,
+        desired_bitrate_bps: u64,
+        now: Instant,
```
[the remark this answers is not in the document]
k0nserv: This might not be needed at all but I'll do that if so

src/lib.rs
```diff
@@ -1285,6 +1332,13 @@ impl RtcConfig {
         self
     }
 
+    /// Whether to use bandwidth estimation to discover the available send bandwidth.
+    pub fn use_bwe(mut self, use_bwe: bool) -> Self {
+        self.use_bwe = use_bwe;
+
+        self
+    }
+
```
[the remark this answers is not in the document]
k0nserv: It's possible, the pacer would need to be replaced and the BWE system initialized on the first call to `configure_bwe`. Personally I think explicit is better than implicit here

src/packet/buffer_tx.rs
```diff
@@ -78,7 +78,7 @@ impl PacketizingBuffer {
         // Scale back retained count to max_retain
         while self.queue.len() > self.max_retain {
             self.queue.pop_front();
-            self.emit_next -= 1;
+            self.emit_next = self.emit_next.saturating_sub(1);
```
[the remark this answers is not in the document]
k0nserv: Overflow and underflow doesn't panic in release builds

src/packet/bwe/arrival_group.rs
k0nserv: I will review all panics in here and have a think. This specific case is an invariant i.e. groups should be non-empty when calling `inter_group_delay_variation`. Could of course make the return type `Option<f64>` and bubble the problem

src/packet/bwe/macros.rs
k0nserv: Not sure what you mean by hidden?

src/packet/bwe/trendline_estimator.rs
k0nserv: Yes, in general I think it's worth adding a comment to this effect in a few places

src/session.rs
k0nserv: Why the sad face?

src/session.rs
k0nserv: I saw this phrasing in libWebRTC, but I was thinking we should rename this struct anyway. I think it ought to be called `GoogccBwe` or something to that effect. In theory we could have other BWE strategies in the future which is an argument for the trait/enum I suggested above

src/session.rs
k0nserv: The problem with logging these statistics via the regular logging system is that it introduces a lot of noise which makes parsing more difficult. Ultimately I'm parsing all the stats produced via this and the macros as CSV roughly doing

```python
if line.removeprefix('PACKET_SENT) != line:
  # Had PACKET_SENT prefix
  data = parse_line_as_csv(line.strip(), ['ssrc', 'bits', 'kind', 'timestamp'])
```

Also, the fact that I'm logging to stdout means I can can do

```
cargo run thing > stats.log
```

and still read all the regular logs, which are logged to stderr, without having these, extremely frequent, values clog that up

algesten: Right. But you can feature flag it, and push it down into `poll_packet` right? And avoid getting that 3rd tuple back from `poll_packet`.

src/packet/bwe/macros.rs
algesten: Something our users don't need to see :)

src/lib.rs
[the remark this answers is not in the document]
algesten: Alright. I'll leave it for now.

src/media/mline.rs
```diff
@@ -414,6 +414,15 @@ impl MLine {
         };
         buf.truncate(header_len + body_len);
 
+        #[cfg(feature = "log_stats")]
+        if let Some(delay) = next
+            .body
+            .queued_at()
+            .map(|i| Instant::now().duration_since(i))
```
algesten, line 421: The `now` passed in as argument seems more appropriate since that would let us control time externally.

src/session.rs
algesten: I've tried to keep the `use crate` breaking up on several lines away. When I merged the packets, I removed a lot of these breaking `use crate`.

Just a tic I have.

src/media/mline.rs
k0nserv: Ah, I didn't realise we had `now` in this context. Will change

src/lib.rs
```diff
@@ -1178,13 +1178,19 @@ impl Rtc {
     /// Staring at the lower layer, call:
     ///
     /// ```no_run
+    /// # use str0m::{Rtc};
```
algesten, line 1181: We could actually run these.

And we don't need `{}` around that import :)

src/lib.rs
```diff
@@ -1178,13 +1178,19 @@ impl Rtc {
...
     /// ````
     ///
     /// When a new estimate is made available that indicates a switch to the medium layer is
     /// possible, make the switch and then update the configuration:
     ///
     /// ```no_run
+    /// # use str0m::{Rtc};
+    /// let mut rtc = Rtc::new();
+    ///
     /// rtc.set_bwe_current_bitrate(750_000.into())
```
algesten, line 1194: Could the Bitrate type have `Bitrate::from_kbit()` and `Bitrate::from_mbit()`?

src/lib.rs
```diff
@@ -1411,12 +1417,13 @@ impl fmt::Debug for Rtc {
...
 ///
-/// ```no_run
+/// ```ignore
 /// log_stat!("MY_STAT", 1, "hello", 3);
 /// ```
 ///
 /// will result in the following being printed
-/// ```
+///
+/// ```ignore
```
algesten, line 1426: Don't you mean `text`, not ignore?

algesten (approved):
Some small things to fix, then merge.

src/lib.rs
k0nserv: I suppose

src/packet/bwe/arrival_group.rs
k0nserv: The code that checks group membership already excludes packets that arrive out of order so there's no need to take the max here. I reviewed this code against libWebRTC and it now matches what they do

PR #87, opened 2023-03-18
remove comment out duplicate

xnorpx:
was trying to compile on windows and saw this one

note that one need a windows version of perl like strawberry perl in the path to get openssl to compile


algesten (approved):
Ah yes. I was too fast there when merging. This dep will be needed if we ever move to rustls for DTLS though, since I can't get the sha-1 crate to work on windows.

PR #92, opened 2023-03-21
Delay allocation of actual SCTP channel until after init

### LINKED ISSUE #91: SCTP channel allocation before SCTP init is wrong
If n the inital OFFER we add a data channel, there's a risk we allocate the incorrect SCTP channel id.

```
        // RFC 8831
        // Unless otherwise defined or negotiated, the
        // streams are picked based on the DTLS role (the client picks even
        // stream identifiers, and the server picks odd stream identifiers).
```

To allocate odd/even means we must know if we're client or server in the SCTP setup. The setup is controlled by the DTLS `a=setup:active` attribute that we only know once we've made the very first OFFER/ANSWER dance. Hence we got a chicken and egg situation – can't allocate the sctp channel before OFFER/ANSWER, but also want to create the initial channel already in the first OFFER.

algesten:
This change makes it so ChannelId is not longer the actual SCTP channel id, but rather mapped via an internal association lookup. The association can then be delayed until after the SCTP init.

Closes #91

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -28,18 +28,18 @@ pub struct ChannelData {
 ///
 /// Get this handle from [`Rtc::channel()`][crate::Rtc::channel()].
 pub struct Channel<'a> {
-    id: ChannelId,
+    sctp_channel: u16,
     rtc: &'a mut Rtc,
 }
 
 impl<'a> Channel<'a> {
-    pub(crate) fn new(id: ChannelId, rtc: &'a mut Rtc) -> Self {
-        Channel { rtc, id }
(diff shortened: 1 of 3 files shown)


src/channel.rs
```diff
@@ -28,18 +28,18 @@ pub struct ChannelData {
 ///
 /// Get this handle from [`Rtc::channel()`][crate::Rtc::channel()].
 pub struct Channel<'a> {
-    id: ChannelId,
+    sctp_channel: u16,
```
xnorpx, line 31: not worth having a type for this?

k0nserv (approved):

PR #94, opened 2023-03-21
BWE Loss controller impl simply based on goog spec

davibe:
This is the "naive loss controller" that is solely based on the spec and not on goog libwebrtc implementation.

- spec: https://datatracker.ietf.org/doc/html/draft-ietf-rmcat-gcc-02#section-6
- zulip stream: https://str0m.zulipchat.com/#narrow/stream/377857-new-features/topic/BWE.20LossController


algesten:
🧟‍♂️

PR #96, opened 2023-03-21
Dev/misc

xnorpx:
misc small things

1. add a comment how to override the vendored openssl flag for people that want to use system installed openssl (this maybe should be a feature flag instead?)
2. change a variable name to be channel instead of media in data channel example in docs
3. fix misc clippy warnings that showed up in our environment.

### DIFF
--- a/src/rtp/rtcp/twcc.rs
+++ b/src/rtp/rtcp/twcc.rs
@@ -563,7 +563,7 @@ impl PacketChunk {
                 break;
             }
 
-            // The stauts is not possible to add in this chunk. This could be
+            // The status is not possible to add in this chunk. This could be
             // a large delta in a single, or a mismatching run.
             if !to_fill.can_append_status(i.status()) {
                 reached_end = false;
@@ -1316,6 +1316,7 @@ impl fmt::Debug for Twcc {
     }
 }
 
+#[allow(clippy::assign_op_pattern)]
 #[cfg(test)]
 mod test {
     use std::time::Duration;
@@ -1784,7 +1785,7 @@ mod test {
 
         let now = Instant::now();
         let base = now + Duration::from_millis(1337 * 64);
-        let expeceted = vec![
+        let expected = vec![
             (
                 1.into(),
                 PacketStatus::ReceivedSmallDelta,
@@ -1837,7 +1838,7 @@ mod test {
 
         let result: Vec<_> = twcc.into_iter(now, 1.into()).collect();
 
-        assert_eq!(result, expeceted);
+        assert_eq!(result, expected);
     }
 
     #[test]
--- a/src/util/value_history.rs
+++ b/src/util/value_history.rs
@@ -52,6 +52,7 @@ where
     }
 }
 
+#[allow(clippy::unchecked_duration_subtraction)]
 #[cfg(test)]
 mod test {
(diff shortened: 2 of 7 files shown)


src/rtp/rtcp/twcc.rs
```diff
@@ -1693,7 +1693,7 @@ mod test {
 
         for i in 0..50 {
             reg.register_seq(i.into(), now, 0);
-            now = now + Duration::from_micros(15);
+            now += Duration::from_micros(15);
```
algesten, line 1696: why are these changing?

src/util/value_history.rs
```diff
@@ -64,17 +64,17 @@ mod test {
 
         let mut h = ValueHistory::new(11, Duration::from_secs(1));
 
-        h.push(now - Duration::from_millis(1500), 22);
-        h.push(now - Duration::from_millis(500), 22);
+        h.push(now.checked_sub(Duration::from_millis(1500)).unwrap(), 22);
+        h.push(now.checked_sub(Duration::from_millis(500)).unwrap(), 22);
```
algesten, line 68: These `-` are deliberate. If we are using `checked_sub` we need to motivate why. I need to write a line about it, cause I only discussed it internally.

This is from our internal Slack:

> As an overall observation of the code going into str0m. I'm seeing an increased use of "saturating_sub" and "saturating_duration". Not saying there is a problem (I should review it). Generally `-` is vastly simpler to read for an observer. Whenenver we use "saturating_xxx" I'd like to see a code comment why the use is correct.
>
> Like: is underflow expected at this call site, if yes, use saturating_sub (and code comment why) 
> Is it not expected? Normal sub and panic if it's an incorrect state.

algesten (commented):
Most things OK. Two things I think need changing.

src/util/value_history.rs
xnorpx: fixed

src/rtp/rtcp/twcc.rs
xnorpx: fixed

algesten (approved):
LGTM!

src/util/value_history.rs
k0nserv: There's a good reason here, see https://rust-lang.github.io/rust-clippy/master/index.html#unchecked_duration_subtraction and https://github.com/rust-lang/rust-clippy/pull/9570.

On some systems this can panic if you run it early enough in the process lifecycle because of how Instant is represented internally. If this where to actually panic we'd fix it by using `checked_sub.unwrap_or(now)` or something like that, so I think we can do it immediately. Generally, all panics of `Instant - Duration` will not be useful to us and should use the above pattern I think.

algesten: These are tests. If we use `checked_sub()` and replace it with `now` (or saturating), the tests will fail instead. The complexity introduced of not using `-` here is not worth it.

k0nserv: Sure, for the tests I don't disagree. Elsewhere though

algesten: I don't want to generalize and say "we should always do this",  but look at case by case. 

My gut feeling is that most instances the logic would break. I prefer panics to incorrect state.

PR #97, opened 2023-03-22
DirectStrategy for changing the session

algesten:
This strategy allows the user to directly change the `Rtc` session without going through the SDP offer/answer dance.


src/change/direct.rs
```diff
@@ -11,7 +11,11 @@ pub struct DirectApi<'a> {
 }
 
 impl<'a> DirectApi<'a> {
-    pub(crate) fn new(rtc: &'a mut Rtc) -> Self {
+    /// Creates a new instance of the `DirectApi` struct with the specified `Rtc` instance.
+    ///
+    /// The `DirectApi` struct provides a high-level API for interacting with a WebRTC peer connection,
+    /// and the `Rtc` instance provides low-level access to the underlying WebRTC functionality.
+    pub fn new(rtc: &'a mut Rtc) -> Self {
```
algesten, line 20: Did you see the `Rtc::direct_api()` call? That's how I intended to instantiate it.

src/dtls/mod.rs
```diff
@@ -48,6 +48,43 @@ pub struct Fingerprint {
...
+            "{} {}",
+            self.hash_func,
+            self.bytes
+                .iter()
+                .map(|b| format!("{:02X}", b))
+                .collect::<Vec<_>>()
+                .join(":")
+        )
+    }
+}
```
algesten, line 64: This is actually doubling up on functionality we have. For some reason, and it should probably be cleaned up sometime. The types that are serialized to SDP does so via the `Display` trait. Very wrong. 

I don't mind having it as `ToString` too, but let's just delegate to the one impl in that case. The `Display` variant is a bit more efficient, not requiring an intermediate Vec.

src/dtls/mod.rs
```diff
@@ -48,6 +48,43 @@ pub struct Fingerprint {
...
+            bytes.push(byte);
+        }
+
+        Ok(Self {
+            hash_func: hash_func.to_owned(),
+            bytes,
+        })
+    }
+}
+
```
algesten, line 87: This one is fine though. Let's keep it but change to do `use std::str:FromStr`.

src/lib.rs
```diff
@@ -498,6 +498,7 @@ use thiserror::Error;
 
 pub use ice::IceConnectionState;
 
+pub use dtls::Fingerprint;
```
algesten, line 501: I will move this eventually since it's low level API use, and I don't want that top level in the crate (again for teachability). Later problem though since I don't have a clear idea of where and what to group together.

src/lib.rs
```diff
@@ -496,7 +496,7 @@ use sctp::{RtcSctp, SctpEvent};
 use stats::{MediaEgressStats, MediaIngressStats, PeerStats, Stats, StatsEvent};
 use thiserror::Error;
 
-pub use ice::IceConnectionState;
+pub use ice::{IceConnectionState, IceCreds};
```
algesten, line 499: Ditto as for `Fingerprint`

src/change/direct.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Sets the ICE controlling flag for this peer connection.
+    ///
+    /// If `controlling` is `true`, this peer connection is set as the ICE controlling agent,
+    /// meaning it will take the initiative to send connectivity checks and control the pace of
+    /// connectivity checks sent between two peers during the ICE session.
+    ///
+    /// If `controlling` is `false`, this peer connection is set as the ICE controlled agent,
+    /// meaning it will respond to connectivity checks sent by the controlling agent.
+    pub fn ice_controlling(&mut self, controlling: bool) {
```
algesten, line 32: Because this API mixes getter/setters, I want to make it clear what is what. Currently `local_ice_credentials()` and `remote_ice_credentials()` sounds like they are the same, but one is a getter and one is a setter. I think I want a `set_xxx` prefix on the setters. 

What do you think?

algesten (commented):
Looking good! Left some comments.

src/change/direct.rs
xnorpx: I am blind thanks! Reverted that commit.

src/lib.rs
xnorpx: I leave these open for you to move.

src/lib.rs
xnorpx: I leave these open for you to move.

src/dtls/mod.rs
xnorpx: ok, so maybe break out some of the code from impl<'a> std::fmt::Display for FingerprintFmt<'a> into a separate function and then use it from both places?

src/dtls/mod.rs
```diff
@@ -55,15 +55,24 @@ impl ToString for Fingerprint {
...
+    let mut formatted = String::with_capacity(string_size);
+    for (i, byte) in bytes.iter().enumerate() {
+        if i > 0 {
+            formatted.push(':');
+        }
+        formatted.push_str(&format!("{:02X}", byte));
+    }
+    formatted
+}
+
```
algesten, line 75: I suspect the fmt API the Display trait is even more optimized, not using an intermediate String. But I don't really care since the serialization of this happens so rarely.

xnorpx: switched back to Display trait

PR #104, opened 2023-03-25
Cheaper queue state

algesten:
This PR aims to reduce the number of QueueState calculations to see if we can improve the flamegraph of where code time is spent.

This splits QueueState into two halves, QueueState/QueueSnapshot and
the snapshot is used on the PacketizingBuffer level and QueueState on
the MediaInner level.  This is to avoid needing all knowledge about
what kind of data we're handling inside the PacketizingBuffer.

The more important change is adding an internal TotalQueue struct in
PacketizingBuffer that continuously keeps track of the total queue
duration in the buffer. This way we can avoid an expensive
recalculation on every MediaInner::queue_state() call.

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -7,8 +7,8 @@ pub use crate::rtp::{Direction, ExtensionValues, MediaTime, Mid, Pt, Rid, Ssrc};
 pub use crate::sdp::{Codec, FormatParams};
 
 use crate::io::{Id, DATAGRAM_MTU};
-use crate::packet::{DepacketizingBuffer, MediaKind, Packetized};
-use crate::packet::{PacketizedMeta, PacketizingBuffer, QueueId, QueueState};
+use crate::packet::{DepacketizingBuffer, MediaKind, Packetized, QueueSnapshot};
+use crate::packet::{PacketizedMeta, PacketizingBuffer, QueueState};
 use crate::rtp::{Extensions, Fir, FirEntry, NackEntry, Pli, Rtcp};
 use crate::rtp::{RtcpFb, RtpHeader, SdesType};
 use crate::rtp::{SeqNo, MAX_PADDING_PACKET_SIZE, SRTP_BLOCK_SIZE, SRTP_OVERHEAD};
@@ -444,12 +444,19 @@ impl MediaInner {
     }
 
     fn poll_packet_resend(&mut self, now: Instant, is_padding: bool) -> Option<NextPacket<'_>> {
-        loop {
+        let (resend, source) = loop {
             let resend = self.resends.pop_front()?;
+
+            // If there is no buffer for this resend, we return None. This is
+            // a weird situation though, since it means the other side sent a nack for
+            // an SSRC that matched this Media, but didnt match a buffer_tx.
+            let buffer = self.buffers_tx.values().find(|p| p.has_ssrc(resend.ssrc))?;
+
+            let pkt = buffer.get(resend.seq_no);
 
             // The seq_no could simply be too old to exist in the buffer, in which
             // case we will not do a resend.
-            let Some(pkt) = packet_for_resend(&self.buffers_tx, &resend) else {
+            let Some(pkt) = pkt else {
                 continue;
             };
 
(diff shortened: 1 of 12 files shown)


src/packet/buffer_tx.rs
```diff
@@ -165,6 +203,80 @@ impl PacketizingBuffer {
...
+//      |         |@@@@@@@@|                  what's
+//      +---------+--------+                  added
+//          +-----+--------+
+//          |     |@@@@@@@@|
+//          +-----+--------+
+//                       +-+
+//                       |@|
+//                       +-+
+#[derive(Debug, Default)]
+struct TotalQueue {
```
k0nserv, line 238: Maybe some doc comments on the fields here

src/packet/buffer_tx.rs
```diff
@@ -165,6 +203,80 @@ impl PacketizingBuffer {
...
+        self.queue_time += queue_time;
+        self.last = Some(now);
+    }
+
+    fn decrease(&mut self, size: usize, queue_time: Duration) {
+        self.unsent_count -= 1;
+        self.unsent_size -= size;
+        self.queue_time -= queue_time;
+        if self.unsent_count == 0 {
+            self.unsent_size = 0;
```
k0nserv, line 273: This ought to be asserts, no? If the code is correct `self.unsent_size` and `self.queue_time` ought to have zero values here already, only `self.last` needs to actually be reset

src/packet/pacer.rs
```diff
@@ -93,53 +94,68 @@ pub trait Pacer {
...
+    /// Last time something was emitted from this queue.
+    pub last_emitted: Option<Instant>,
+    /// Time the first unsent packet has spent in the queue.
+    pub first_unsent: Option<Instant>,
+}
+
+impl Default for QueueSnapshot {
+    fn default() -> Self {
+        Self {
+            created_at: not_happening(),
```
k0nserv, line 123: Wouldn't it be nicer to `QueueSnapshot::new(created_at: Instant)`? No need for this special zero value that way

src/media/inner.rs
```diff
@@ -1065,53 +1106,43 @@ impl MediaInner {
...
 
-            state.packet_count += 1;
-            state.size += original_packet.data.len().into();
-            state.total_queue_time += now.saturating_duration_since(resend.queued_at);
-            state.update_leading_queue_time(Some(resend.queued_at));
-        }
+        let state = QueueState {
+            mid: self.mid,
+            is_audio: self.kind == MediaKind::Audio,
+            use_for_padding: self.has_tx_rtx(),
```
k0nserv, line 1125: This is a potential regression, we don't verify that `self.kind == MediaKind::Video`, might be moot point since we don't generally have RTX for audio.

k0nserv (commented):
Gonna try locally

src/packet/buffer_tx.rs
algesten: sure

src/packet/buffer_tx.rs
algesten: true that

k0nserv:
Also, I added MediaKind in the pacer because there is some code in libWebRTC that prioritises retransmits over video, but I hadn't decide if we wanted to port that. The vision was to have `MediaKind::Retransmission` in addition to `MediaKind::Audio` and MediaKind::Video`

src/media/inner.rs
algesten: Yeah. Let's fix that.

algesten:
> Also, I added MediaKind in the pacer because there is some code in libWebRTC that prioritises retransmits over video, but I hadn't decide if we wanted to port that. The vision was to have `MediaKind::Retransmission` in addition to `MediaKind::Audio` and `MediaKind::Video`

That's still fine though. Since we have `QueueState(QueueSnapshot)` and what type of queue lies on the outer `QueueState`, we can introduce whatever flags we want there without requiring that for the snapshots/packetizing buffers.

I also think that rather than making Pacer hard bound to media, we could abstract it to just a prio field (high, normal, low) or something.

src/packet/pacer.rs
k0nserv: Your logic makes sense with regards to why we couldn't use that constructor, I don't fully agree with regards to the encapsulation bit but w/e

k0nserv:
> I also think that rather than making Pacer hard bound to media, we could abstract it to just a prio field (high, normal, low) or something.

Sounds good

src/packet/buffer_tx.rs
```diff
@@ -165,6 +203,80 @@ impl PacketizingBuffer {
...
+        self.unsent_count += 1;
+        self.unsent_size += size;
+        self.queue_time += queue_time;
+        self.last = Some(now);
+    }
+
+    fn decrease(&mut self, size: usize, queue_time: Duration) {
+        self.unsent_count -= 1;
+        self.unsent_size -= size;
+        self.queue_time -= queue_time;
```
k0nserv, line 275: Quick test paniced here pretty immediately, with some extra debug info

```
thread 'tokio-runtime-worker' panicked at 'Expected queue_time(22.843375ms) larger than self.queue_time(1.006668ms)', /Users/hugotunius/Code/Personal/str0m/src/packet/buffer_tx.rs:271:9
```

PR #105, opened 2023-03-25
[WIP] Drop media packets before ice connection 

morajabi:
Zulip stream: https://str0m.zulipchat.com/#narrow/stream/377845-general/topic/Writing.20media.20before.20ice.20connection.20causes.20delay.3F

Update: my initial test show this fix did not fix the problem. I fixed this in my app code to not write media until state is not disconnected or new and fixed my issue. But we should handle this in str0m.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1198,6 +1198,10 @@ impl Rtc {
                 };
                 return Ok(Output::Transmit(t));
             }
+        } else {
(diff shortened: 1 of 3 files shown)


algesten:
Sorry I sent you on a goose chase. I realised we had some handy functions around already for emptying the send buffers. It's in main now.

PR #109, opened 2023-03-27
bwe: initial bitrate improvements

davibe:


src/session.rs
```diff
@@ -91,23 +91,25 @@ pub enum MediaEvent {
...
         let (pacer, bwe) = if use_bwe {
-            let initial_bitrate = 300_000.into();
-            let mut pacer = PacerImpl::LeakyBucket(LeakyBucketPacer::new(
-                initial_bitrate * PACING_FACTOR * 2.0,
+            let pacer = PacerImpl::LeakyBucket(LeakyBucketPacer::new(
+                bwe_initial_estimate * PACING_FACTOR * 2.0,
                 Duration::from_millis(40),
             ));
 
-            pacer.set_padding_rate(initial_bitrate * PADDING_FACTOR);
```
k0nserv, line 108: We always want padding to be set in the same way, I don't see why we should remove this

src/lib.rs
```diff
@@ -1576,6 +1587,7 @@ impl Default for RtcConfig {
             codec_config: CodecConfig::new_with_defaults(),
             stats_interval: Duration::from_secs(1),
             bwe: false,
+            bwe_initial_bitrate: Bitrate::kbps(1500),
```
algesten, line 1590: What is a good value here? 300 before, now 1500. Do we have a reason? Can we document the reason in code?

algesten (approved):

src/lib.rs
k0nserv, line 1590: The 300Kbit/s value I used was lifted from libWebRTC somewhere I think, that said since the initial bitrate is quite tied to what you are initially intending to send perhaps we should force the user to always provide it instead?

src/session.rs
davibe: if you have no desired bitrate to reach, why would you have a padding target ?

k0nserv: Initial bitrate == desired bitrate. If we don't pad the estimates will be lies, for example if the user says `initial_bitrate: Bitrate::kbps(1500)` but then sends only 750Kbit/s of media the resulting estimates will not be reliable since we aren't actually sending near the estimate

src/lib.rs
davibe: No problem with that. In that case we can use the optional for enabling instead of a separate flag.

src/session.rs
k0nserv: Either case might be a bit moot since the user will probably reconfigure the padding and pacing rate immediately anyway.

One thing to consider is what happens if we make BWE ignore audio packets, as we intend to. Then I think it might be suitable to set the desired bitrate to something like `expected_ingress_video_rate * 1.15`, thus instructing the BWE system to discover if we have enough bitrate for even a single video track. 

With all of this maybe we should just add an explicit desired_bitrate argument to this method as well?

src/lib.rs
k0nserv: Based on the other comment thread I think the signature ought to be

```rust
fn enable_bwe(initial_estimate: Bitrate, desired_bitrate: Bitrate)
```

src/session.rs
davibe: Not sure I understand to which method we would add desired_bitrate and why.

I do understand having a margin for audio, which can be taken into account when calling set_bwe_desired_bitrate()

davibe: ah, i see the other comment

src/lib.rs
davibe: I don't think this is needed.
It makes sense to stick to configuration for initial_estimate because it's useful to start in high quality depending on the use case and because it cannot change afterwards (so the api is not misleading). Instead, for something that can be already set otherwise, and is not vital, we already have the setter which is enough.

src/session.rs
k0nserv: I'm suggesting this method should accept both desired and initial rates. The idea is to ignore audio entirely for BWE purposes.

However if we start with only audio I think we'd want the initial values to be something like 

```
config.enable_bwe(Bitrate::kbps(1500), Bitrate::kbps(1800)
```

As you said before we start at 1.5Mbit/s because that's faster than ramping up from 300Kbit/s and we'll quickly decrease the estimate if we are exceeding. Typically, at least for our use case, we'll initially only be sending an audio track but we still want to discover enough bitrate to later receive some video tracks so we'd set the desired to something that could sustain either a high quality video track or a few lower quality tracks. This way, when we do receive some video tracks we'll have already discovered some bitrate

src/lib.rs
k0nserv: The only problem is that if you don't eventually set the desired bitrate we'll just keep increasing padding with no limit.

k0nserv: No you are right, I thought it would pad to the estimate anyway but it only does that if you actually set the desired rate

davibe: @k0nserv I update making the initial estimate mandatory. See if you're relatively happy with this or still a blocker

src/session.rs
davibe: lets continue this in the other "thread" below

PR #110, opened 2023-03-28
Robuster Acked Bitrate

k0nserv:
Ports part of the, more robust, acked bitrate estimator from libWebRTC.

### DIFF
--- a/src/packet/bwe/acked_bitrate_estimator.rs
+++ b/src/packet/bwe/acked_bitrate_estimator.rs
@@ -1,0 +1,161 @@
+use std::time::{Duration, Instant};
+
+use crate::rtp::DataSize;
+use crate::Bitrate;
+
+pub struct AckedBitrateEstimator {
+    /// The initial window to use for the first estimate.
+    initial_window: Duration,
+    /// The window to use for subsequent estimates after achieving the initial estimate.
+    window: Duration,
+    /// The estimate of the acked bitrate.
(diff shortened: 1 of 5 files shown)


src/packet/bwe/acked_bitrate_estimator.rs
```diff
@@ -0,0 +1,161 @@
...
+    fn update_window(
+        &mut self,
+        receive_time: Instant,
+        packet_size: DataSize,
+        window: Duration,
+    ) -> Option<Bitrate> {
+        if self
+            .last_update
+            .map(|ls| receive_time < ls)
+            .unwrap_or(false)
```
algesten, line 61: I think this could be clearer with

```rust
let time_moved_back = Some(receive_time) < self.last_update;
if time_moved_back {
…
}
```

algesten: This is also a case where `will_not_happen()` as initial value will make easier logic than juggling that `Option`.

algesten (approved):

src/packet/bwe/acked_bitrate_estimator.rs
k0nserv: Are you sure, it would need to be `already_happened` but then I do math on this so that would need special handling to not calculate a window that's several hours.

I did what you suggested above

PR #112, opened 2023-03-28
sctp/data channel refactoring

algesten:
This tidies up the init sequence for data channels in anticipation of direct api needs.

The correct sequence is:

```
self.rtc.chan.new_channel(&config);
self.rtc.chan.confirm(id, config);
```

There is a new `ChannelConfig` object to configure all the options of the data channel.

### DIFF
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -1,7 +1,9 @@
+use crate::channel::ChannelId;
 use crate::dtls::Fingerprint;
 use crate::ice::IceCreds;
 use crate::rtp::Direction;
 use crate::rtp::Mid;
+use crate::sctp::ChannelConfig;
 use crate::Rtc;
 use crate::RtcError;
 
@@ -29,7 +31,7 @@ impl<'a> DirectApi<'a> {
     ///
     /// If `controlling` is `false`, this peer connection is set as the ICE controlled agent,
     /// meaning it will respond to connectivity checks sent by the controlling agent.
-    pub fn ice_controlling(&mut self, controlling: bool) {
+    pub fn set_ice_controlling(&mut self, controlling: bool) {
         self.rtc.ice.set_controlling(controlling);
     }
 
@@ -37,12 +39,12 @@ impl<'a> DirectApi<'a> {
     ///
     /// The ICE credentials consist of the username and password used by the ICE agent during
     /// the ICE session to authenticate and exchange connectivity checks with the remote peer.
-    pub fn local_ice_credentials(&self) -> &IceCreds {
-        self.rtc.ice.local_credentials()
+    pub fn local_ice_credentials(&self) -> IceCreds {
+        self.rtc.ice.local_credentials().clone()
     }
 
     /// Sets the remote ICE credentials.
-    pub fn remote_ice_credentials(&mut self, remote_ice_credentials: IceCreds) {
+    pub fn set_remote_ice_credentials(&mut self, remote_ice_credentials: IceCreds) {
         self.rtc.ice.set_remote_credentials(remote_ice_credentials);
     }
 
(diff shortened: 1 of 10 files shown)


src/sctp/mod.rs
```diff
@@ -72,22 +72,79 @@ impl RtcSctpState {
...
+    /// If we have sent DcepOpen and are waiting for the ack.
     AwaitDcepAck,
+    /// Stream is open, ready to send data.
     Open,
+    /// If some error occurs.
     Closed,
 }
 
+#[derive(Debug, Default)]
+pub struct ChannelConfig {
```
xnorpx, line 127: should this be exported in lib?

src/channel.rs
```diff
@@ -52,3 +64,169 @@ impl fmt::Debug for ChannelData {
...
+        let id = ChannelId(self.allocations.len());
+
+        // For out-of-band negotiated, the id is already set.
+        let sctp_stream_id = config.negotiated;
+        if let Some(sctp_stream_id) = sctp_stream_id {
+            let exists = self
+                .allocations
+                .iter()
+                .any(|a| a.sctp_stream_id == Some(sctp_stream_id));
+            assert!(exists, "sctp_stream_id ({}) exists already", sctp_stream_id);
```
xnorpx, line 96: should this be !exists?

src/sctp/mod.rs
algesten: Yes. Once we got an API that uses it.

src/channel.rs
algesten: No? I mean to check that one doesn't set a channel that already has been set before automatically or manually.

src/sctp/mod.rs
algesten: Done

src/channel.rs
algesten: You're right! Fixed.

k0nserv (approved):
Some minor things

src/change/direct.rs
```diff
@@ -37,8 +39,8 @@ impl<'a> DirectApi<'a> {
     ///
     /// The ICE credentials consist of the username and password used by the ICE agent during
     /// the ICE session to authenticate and exchange connectivity checks with the remote peer.
-    pub fn local_ice_credentials(&self) -> &IceCreds {
```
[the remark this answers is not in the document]
algesten: True that. I need to revise this API overall.

src/sctp/mod.rs
```diff
@@ -72,22 +72,93 @@ impl RtcSctpState {
...
+/// Reliability setting of a data channel.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Reliability {
+    /// Packets are delivered in order, with retransmits.
+    #[default]
+    Reliable,
+    /// Packets delivered out of order with a max lifetime.
+    MaxPacketLifetime {
+        /// The lifetime of a packet in milliseconds.
+        lifetime: u16,
```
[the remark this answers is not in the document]
algesten: It's the same as this: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/createDataChannel#maxpacketlifetime

Maybe should link to that doc?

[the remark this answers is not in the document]
algesten: But isn't a Duration is too permissive? The value, on SCTP level, is a u16 milliseconds. With Duration we can set "years". By using this datatype, we only permit the values that are actually accepted.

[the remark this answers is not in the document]
k0nserv: The spec already talks about the application potentially limiting it further, I guess if we want to use `u16` to enforce the 65536ms maximum and not restrict it further then we can keep it as `u16`. If we want to impose further limits we'd need custom code regardless of the type we pick

PR #116, opened 2023-03-29
Make accounting for audio in pacer configurable

k0nserv:
- Make accounting for audio conditional in pacers
- Fix comment

### DIFF
--- a/src/packet/bwe/mod.rs
+++ b/src/packet/bwe/mod.rs
@@ -37,7 +37,7 @@ pub struct SendSideBandwithEstimator {
     trendline_estimator: TrendlineEstimator,
     rate_control: RateControl,
     acked_bitrate_estimator: AckedBitrateEstimator,
-    /// Last unpolled bitrate estimate. [`None`] before the first poll and after each poll that,
+    /// Last unpolled bitrate estimate. [`None`] before the first poll and after each poll,
     /// updated when we get a new estimate.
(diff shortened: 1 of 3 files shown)


src/packet/bwe/mod.rs
```diff
@@ -37,7 +37,7 @@ pub struct SendSideBandwithEstimator {
     trendline_estimator: TrendlineEstimator,
     rate_control: RateControl,
     acked_bitrate_estimator: AckedBitrateEstimator,
-    /// Last unpolled bitrate estimate. [`None`] before the first poll and after each poll that,
+    /// Last unpolled bitrate estimate. [`None`] before the first poll and after each poll,
```
algesten, line 40: Full stop instead of comma ;)

algesten (commented):
Instead of pushing the knowledge of audio down into Pacer, why isn't it terminated in session, so you check if it is audio and add it to the pacer in that case?

algesten:
Or put the other way: Why does pacer need to know whether it's audio?

k0nserv:
> Or put the other way: Why does pacer need to know whether it's audio?

Two reasons:

1. Audio is never paced, the pacer needs to know which packets, or as currently, which queues are audio to do this.
2. Depending on this new `accounts_for_audio` it needs to either include audio packets when accruing debt or not.


I guess both these goals could be achieved in different ways though. This is based on what libWebRTC does

PR #119, opened 2023-04-01
Make PayloadParams, CodecSpec and Codec public API

algesten:
This splits out the codec configuration related structs from the sdp module, making it clearer what is about SDP parsing/serializing and what is generic config of the session.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -5,7 +5,8 @@ use std::fmt;
 
 use crate::media::MediaKind;
 use crate::rtp::Pt;
-use crate::sdp;
+use crate::sdp::FormatParam;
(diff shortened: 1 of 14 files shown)


src/format.rs
```diff
@@ -19,64 +20,165 @@ use super::MediaKind;
...
+
         PayloadParams {
-            inner: p,
+            pt,
+            resend,
+
+            spec,
+
+            // Both audio and video use TWCC
+            fb_transport_cc: true,
```
xnorpx, line 149: this you might want to leave configurable for new

xnorpx (approved):
lgtm

PR #123, opened 2023-04-04
New argument to add_media for setting stream_id

algesten:
The stream id is part of the msid mechanism for identifying streams that belong together and therefore should be synced.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -115,7 +115,7 @@ impl<'a> SdpApi<'a> {
     /// let mut rtc = Rtc::new();
     ///
     /// let mut changes = rtc.sdp_api();
-    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendOnly, None);
+    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendOnly, None, None);
     /// let (offer, pending) = changes.apply().unwrap();
     ///
     /// // send offer to remote peer, receive answer back
@@ -179,7 +179,7 @@ impl<'a> SdpApi<'a> {
     /// let mut changes = rtc.sdp_api();
     /// assert!(!changes.has_changes());
     ///
-    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendRecv, None);
+    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendRecv, None, None);
     /// assert!(changes.has_changes());
     /// ```
     pub fn has_changes(&self) -> bool {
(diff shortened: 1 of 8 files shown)


src/change/sdp.rs
```diff
@@ -193,33 +193,50 @@ impl<'a> SdpApi<'a> {
...
+                || (0x23..=0x27).contains(&u)
+                || (0x2a..=0x2b).contains(&u)
+                || (0x2d..=0x2e).contains(&u)
+                || (0x30..=0x39).contains(&u)
+                || (0x41..=0x5a).contains(&u)
+                || (0x5e..0x7e).contains(&u)
+        }
+
+        let stream_id = if let Some(stream_id) = stream_id {
+            stream_id.chars().filter(is_token_char).take(64).collect()
```
k0nserv, line 233: Should probably error if the user used an illegal stream id instead of silently fixing it

src/change/sdp.rs
```diff
@@ -193,33 +193,50 @@ impl<'a> SdpApi<'a> {
...
     ///
-    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendRecv, None);
+    /// let mid = changes.add_media(MediaKind::Audio, Direction::SendRecv, None, None);
     /// ```
-    pub fn add_media(&mut self, kind: MediaKind, dir: Direction, cname: Option<String>) -> Mid {
+    pub fn add_media(
+        &mut self,
+        kind: MediaKind,
+        dir: Direction,
+        stream_id: Option<String>,
```
k0nserv, line 212: Note: Per the spec a given track can belong to many streams, not just the one and in the spec API you can add new streams after the initial negotiation. Not something we have to necessarily address, we can limit ourselves to one stream, but I thought I'd mention it.

Some relevant discussions:

* https://bugs.chromium.org/p/chromium/issues/detail?id=1342010
* https://github.com/w3c/webrtc-pc/issues/2782

src/change/sdp.rs
algesten: Maybe. Don't want to be overly restrictive.

k0nserv: We'll the way it works now it might cause confusing errors instead, fail fast after all ;)

algesten: I'm more in the camp of providing an easy-to-use API than an overly correct API. Rust libs tend to go for super correct. It's a philosophical question.
 
"What are the chances the user provides some crazy value here?"
Probably not that high. 

"Is it worth introducing yet another Result return type the user has to deal with?"
Maybe, let's see how it goes.

k0nserv: Could just panic too, in the spirit of the rest of the lib I think

algesten: Sure. Panics are nice.

PR #124, opened 2023-04-04
Continuously update padding

k0nserv:
- Continuously update padding with each BWE estimate
- Reduce padding factor to 100%
- Make sure we don't pad when there's media to send


src/packet/pacer.rs
```diff
@@ -287,7 +287,16 @@ impl Pacer for LeakyBucketPacer {
...
 
-        if self.next_poll_outcome.is_some() {
+        // We skip recalculating the next poll outcome if we already have one, with the exception
+        // when the next action is padding. The reason we still recalculate if the outcome is
+        // padding is that something more important might still be required i.e. we might have
+        // queued media packets.
+        let skip_recalcualting = self
+            .next_poll_outcome
+            .map(|o| !matches!(o, PollOutcome::PollPadding(_, _)))
+            .unwrap_or(false);
```
algesten, line 297: This would be simpler to read like:

```rs
        let skip_recalcualting =
            !matches!(self.next_poll_outcome, Some(PollOutcome::PollPadding(_, _)));
```

davibe (approved):

src/session.rs
```diff
@@ -853,4 +860,16 @@ impl Session {
...
+        pacer: &mut PacerImpl,
+        last_estimate: Option<Bitrate>,
+        desired_bitrate: Bitrate,
+    ) {
+        let padding_rate = last_estimate
+            .map(|bwe| (bwe * PADDING_FACTOR).min(desired_bitrate))
+            .unwrap_or(Bitrate::ZERO);
+
+        pacer.set_padding_rate(padding_rate);
+    }
```
algesten, line 874: This looks too complicated. Consider doing something like this:

```diff
diff --git a/src/session.rs b/src/session.rs
index 77dee7a..3865470 100644
--- a/src/session.rs
+++ b/src/session.rs
@@ -69,6 +69,7 @@ pub(crate) struct Session {
     twcc_tx_register: TwccSendRegister,
 
     bwe: Option<SendSideBandwithEstimator>,
+    bwe_desired_bitrate: Bitrate,
 
     enable_twcc_feedback: bool,
 
@@ -130,6 +131,7 @@ impl Session {
             twcc_rx_register: TwccRecvRegister::new(100),
             twcc_tx_register: TwccSendRegister::new(1000),
             bwe,
+            bwe_desired_bitrate: Bitrate::ZERO,
             enable_twcc_feedback: false,
             pacer,
             poll_packet_buf: vec![0; 2000],
@@ -512,6 +514,7 @@ impl Session {
                     let records = range.and_then(|range| self.twcc_tx_register.send_records(range));
 
                     if let Some(records) = records {
+                        self.configure_pacer_padding();
                         bwe.update(records, now);
                     }
                 }
@@ -826,14 +829,8 @@ impl Session {
     }
 
     pub fn set_bwe_desired_bitrate(&mut self, desired_bitrate: Bitrate) {
-        if let Some(bwe) = &mut self.bwe {
-            let padding_rate = bwe
-                .last_estimate()
-                .map(|bwe| (bwe * PADDING_FACTOR).min(desired_bitrate))
-                .unwrap_or(Bitrate::ZERO);
-
-            self.pacer.set_padding_rate(padding_rate);
-        }
+        self.bwe_desired_bitrate = desired_bitrate;
+        self.configure_pacer_padding();
     }
 
     pub fn line_count(&self) -> usize {
@@ -853,4 +850,18 @@ impl Session {
             m.clear_send_buffers();
         }
     }
+
+    fn configure_pacer_padding(&mut self) {
+        let Some(last_estimate) = self.bwe.as_ref().map(|b| b.last_estimate()) else {
+            return;
+        };
+
+        let padding_rate = if let Some(last) = last_estimate {
+            (last * PADDING_FACTOR).min(self.bwe_desired_bitrate)
+        } else {
+            self.bwe_desired_bitrate
+        };
+
+        self.pacer.set_padding_rate(padding_rate);
+    }
 }
```

algesten (changes requested):

src/packet/pacer.rs
k0nserv: Good point

src/session.rs
k0nserv: I did something like this initially but had problems with overlapping `&mut self` borrows. Does that actually compile?

algesten: Does indeed compile

k0nserv: Hmm, let me try again

src/packet/pacer.rs
k0nserv: Actually, no this will be wrong for `None` in which case we do want to calculate the next action

src/session.rs
k0nserv: I still get a conflicting borrow between `records` and the call to `configure_pacer_padding` but fixed it slightly differently

algesten (approved):

PR #127, opened 2023-04-04
Let Packetizer decide if it is a marker

algesten:


xnorpx:
set it to false now, but I only tested audio

xnorpx (approved):
lgtm

PR #131, opened 2023-04-06
BWE tweaks

k0nserv:
- Don't use queues for padding before first media
- Increase max consecutive padding packets
- Tweak pacing and padding rates

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -1128,9 +1128,7 @@ impl MediaInner {
         let state = QueueState {
             mid: self.mid,
             is_audio: self.kind.is_audio(),
-            use_for_padding: self.kind.is_video()
-                && self.has_tx_rtx()
-                && snapshot.last_emitted.is_some(),
+            use_for_padding: self.kind.is_video() && self.has_tx_rtx() && snapshot.has_ever_sent(),
             snapshot,
         };
 
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -1128,7 +1128,9 @@ impl MediaInner {
         let state = QueueState {
             mid: self.mid,
             is_audio: self.kind.is_audio(),
-            use_for_padding: self.kind.is_video() && self.has_tx_rtx(),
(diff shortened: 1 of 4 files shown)


src/media/inner.rs
```diff
@@ -1128,7 +1128,9 @@ impl MediaInner {
         let state = QueueState {
             mid: self.mid,
             is_audio: self.kind.is_audio(),
-            use_for_padding: self.kind.is_video() && self.has_tx_rtx(),
+            use_for_padding: self.kind.is_video()
+                && self.has_tx_rtx()
+                && snapshot.last_emitted.is_some(),
```
algesten, line 1133: What is the semantic meaning of this third argument?

src/packet/pacer.rs
```diff
@@ -11,7 +11,7 @@ const MAX_BITRATE: Bitrate = Bitrate::gbps(10);
 const MAX_DEBT_IN_TIME: Duration = Duration::from_millis(500);
 const MAX_PADDING_PACKET_SIZE: DataSize = DataSize::bytes(224);
 const PADDING_BURST_INTERVAL: Duration = Duration::from_millis(5);
-const MAX_CONSECUTIVE_PADDING_PACKETS: usize = 5_000;
+const MAX_CONSECUTIVE_PADDING_PACKETS: usize = 1 << 16;
```
algesten, line 14: u16::max ?

algesten (commented):
Some small things.

I like the cleanup in Session!

src/media/inner.rs
k0nserv: It's saying that we only use queues for padding if we have previously sent something on them, definitionally that thing will be media

algesten: Could you maybe make that clearer by having `snapshot.has_sent_something()` or similar?

k0nserv: Sure

algesten (approved):
LGTM

davibe (approved):

PR #132, opened 2023-04-06
Don't bail if there is no resend padding packet

k0nserv:


src/media/inner.rs
```diff
@@ -579,23 +579,23 @@ impl MediaInner {
...
-                assert!(self.resends.is_empty());
-                self.resends.push_front(Resend {
-                    pt,
-                    ssrc: packet.meta.ssrc,
-                    seq_no,
-                    queued_at: now,
-                });
-
-                return self.poll_packet_resend(now, true);
+                if let Some(packet) = buffer.historic_packet_smaller_than(max_size) {
```
algesten, line 582: Can you avoid reindentation by using let-else and early return?

algesten (changes requested):

src/media/inner.rs
k0nserv: No because the other block we want to run is in the same function

k0nserv: The problem before was that this code did an early return here, that's exactly what we don't

algesten (approved):
Alright

PR #134, opened 2023-04-06
Config for rx buffer reordering sizes

algesten:


morajabi (approved):
Oh neat

src/lib.rs
```diff
@@ -1584,6 +1586,32 @@ impl RtcConfig {
         self
     }
 
+    /// Sets the number of packets held back for reordering audio packets.
+    ///
+    /// Str0m tries to deliver the samples in order. This number determines how many
+    /// packets to "wait" before releasing media out of order.
+    ///
+    /// Default: 15
```
k0nserv, line 1594: Nitpic: Maybe we make these values exported constants so they can be referred to here i.e.

```
Default: see [`REORDERING_SIZE_AUDIO`]
```

or so,

saves us keeping this documentation comment in sync with the values in `build`.

k0nserv (approved):

src/lib.rs
algesten: Good point. But how far do we take that idea? Like what about "Defaults to `false`" for ice-lite – do we turn that bool into a constant too?

Another idea is to ensure all values have getters on the `RtcConfig` and then lock down the documented value with a doc test?

k0nserv: > Another idea is to ensure all values have getters on the RtcConfig and then lock down the documented value with a doc test?

Oh, that's a nice way to do it 🤩

PR #135, opened 2023-04-07
Extend MediaTime

algesten:
This fixes a bug where the `MediaData::time` is not extended. The intention is that time will just keep going up.

### DIFF
--- a/src/rtp/header.rs
+++ b/src/rtp/header.rs
@@ -259,10 +259,16 @@ impl RtpHeader {
 }
 
 macro_rules! mk_extend {
-    ($id:ident, $t:ty, $bits:expr) => {
+    ($id:ident, $t:ty) => {
         /// "extend" a less than 64 bit sequence number into a 64 bit by
         /// using the knowledge of the previous such sequence number.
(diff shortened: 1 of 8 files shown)


davibe (approved):

src/rtp/header.rs
```diff
@@ -253,50 +253,61 @@ impl RtpHeader {
...
-    if prev_ext_seq.is_none() {
-        // No wrap-around so far.
-        return seq;
-    }
+macro_rules! mk_extend {
+    ($id:ident, $t:ty, $bits:expr) => {
+        /// "extend" a less than 64 bit sequence number into a 64 bit by
+        /// using the knowledge of the previous such sequence number.
+        pub fn $id(prev_ext_seq: Option<u64>, seq: $t) -> u64 {
+            const MAX: u64 = 2_u64.pow($bits); // u16: 65_536;
```
k0nserv, line 266: Wouldn't `$ty::MAX` work here? That way the `$bits` argument can be removed

algesten: Good point. Let's see if I can fix.

PR #138, opened 2023-04-12
Add a realistic pacer test

k0nserv:
Attempting to chase down the source of our lacklustre padding. This introduces a test that simulates realistic conditions for the pacer and ensure it's asking for the correct amount of padding.

### DIFF
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -645,6 +645,7 @@ impl From<Option<&QueueState>> for PollOutcome {
 
 #[cfg(test)]
 mod test {
+    use std::ops::Range;
     use std::thread;
     use std::time::{Duration, Instant};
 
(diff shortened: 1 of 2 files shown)


algesten (approved):

davibe (approved):

src/packet/pacer.rs
```diff
@@ -1142,14 +1169,162 @@ mod test {
...
+        spike_probability: u8,
+        max_overshoot_factor: f32,
+        frame_pacing: Duration,
+    }
+
+    impl Default for RealisticTestConfig {
+        fn default() -> Self {
+            RealisticTestConfig {
+                media_rate: Bitrate::kbps(250),
+                padding_rate: Bitrate::kbps(800),
```
davibe, line 1196: I am just thinking I had problems with 250kbps of media, to validate bandwidth up to 1500kbps. This could validate that scenario ? i.e. the ability to pad to that rate

k0nserv: Yea, the actual test uses 2.5Mbit/s in fact.

PR #140, opened 2023-04-13
Pace padding as part of packet queues

k0nserv:
Previously packet burst of 5ms worth of the padding rate were sent
immediately and were not paced, in the process interrupting media
transmission. To alleviate this, this PR causes padding, like all other
packets, to be queued and paced. Padding is prioritized below all other
types of packets.

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -10,7 +10,7 @@ use crate::packet::{DepacketizingBuffer, MediaKind, Packetized, QueuePriority, Q
 use crate::packet::{PacketizedMeta, PacketizingBuffer, QueueState};
 use crate::rtp::{ExtensionMap, Fir, FirEntry, NackEntry, Pli, Rtcp};
 use crate::rtp::{RtcpFb, RtpHeader, SdesType};
-use crate::rtp::{SeqNo, MAX_PADDING_PACKET_SIZE, SRTP_BLOCK_SIZE, SRTP_OVERHEAD};
+use crate::rtp::{SeqNo, MAX_BLANK_PADDING_PAYLOAD_SIZE, SRTP_BLOCK_SIZE, SRTP_OVERHEAD};
 use crate::sdp::{MediaLine, Msid, Simulcast as SdpSimulcast};
 use crate::stats::StatsSnapshot;
 use crate::util::already_happened;
@@ -210,8 +210,8 @@ enum NextPacketBody<'a> {
         pkt: &'a Packetized,
         orig_seq_no: Option<SeqNo>,
     },
-    /// An empty padding packet to be generated.
-    Padding { len: u8 },
+    /// An blank padding packet to be generated.
+    Blank { len: u8 },
 }
 
 impl MediaInner {
@@ -445,7 +445,7 @@ impl MediaInner {
         exts: &ExtensionMap,
         twcc: &mut u64,
         buf: &mut Vec<u8>,
-    ) -> Option<(RtpHeader, SeqNo, bool, usize)> {
+    ) -> Option<PolledPacket> {
         let mid = self.mid;
 
         let (next, is_padding) = if let Some(next) = self.poll_packet_resend_to_cap(now) {
@@ -507,7 +507,7 @@ impl MediaInner {
 
                 body_len + original_seq_len + pad_len
             }
-            NextPacketBody::Padding { len } => {
+            NextPacketBody::Blank { len } => {
                 let len = RtpHeader::create_padding_packet(
                     &mut buf[..],
                     len,
@@ -528,7 +528,13 @@ impl MediaInner {
             crate::log_stat!("QUEUE_DELAY", header.ssrc, delay.as_millis());
         }
 
-        Some((header, next.seq_no, is_padding, body_len))
+        // Some((header, next.seq_no, is_padding, body_len))
+        Some(PolledPacket {
+            header,
+            twcc_seq_no: next.seq_no,
+            is_padding,
+            payload_size: body_len,
+        })
     }
 
     fn poll_packet_resend_to_cap(&mut self, now: Instant) -> Option<NextPacket> {
@@ -643,12 +649,12 @@ impl MediaInner {
 
     fn poll_packet_padding(&mut self, now: Instant) -> Option<NextPacket> {
         loop {
-            let padding_or_resend = self.padding.pop_front()?;
+            let padding = self.padding.pop_front()?;
 
             // Force recaching since padding changed.
             self.queue_state = None;
 
-            match padding_or_resend {
+            match padding {
                 Padding::Blank { ssrc, pt, size, .. } => {
                     let source_tx = get_or_create_source_tx(
                         &mut self.sources_tx,
@@ -657,12 +663,14 @@ impl MediaInner {
                     );
                     let seq_no = source_tx.next_seq_no(now, None);
 
-                    trace!("Generating padding packet of size {size} on {ssrc} with pt: {pt}");
+                    trace!(
+                        "Generating blank padding packet of size {size} on {ssrc} with pt: {pt}"
+                    );
                     return Some(NextPacket {
                         pt,
                         ssrc,
                         seq_no,
-                        body: NextPacketBody::Padding { len: size as u8 },
+                        body: NextPacketBody::Blank { len: size as u8 },
                     });
                 }
                 Padding::Spurious(resend) => {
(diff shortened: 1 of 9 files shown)


src/media/inner.rs
```diff
@@ -424,17 +438,16 @@ impl MediaInner {
         now: Instant,
         exts: &ExtensionMap,
         twcc: &mut u64,
-        pad_size: Option<usize>,
         buf: &mut Vec<u8>,
-    ) -> Option<(RtpHeader, SeqNo)> {
+    ) -> Option<(RtpHeader, SeqNo, bool, usize)> {
```
k0nserv, line 442: Maybe it's time we have a name typed returned here rather than a tuple with four elements.

```rust
struct PolledPacket {
  header: RtpHeader,
  twcc_seq_no: SeqNo,
  is_padding: bool,
  payload_size: usize,
}
```

src/media/inner.rs
```diff
@@ -622,86 +635,178 @@ impl MediaInner {
...
-                    return self.poll_packet_resend(now, true);
+                    let pkt = buffer.get(resend.seq_no);
+
+                    // The seq_no could simply be too old to exist in the buffer, in which
+                    // case we will not do a resend.
+                    let Some(pkt) = pkt else {
+                        continue;
+                    };
+
+                    // The send source, to get a contiguous seq_no for the resend.
```
k0nserv, line 679: This shares code with `poll_packet_resend`, might warrant a method

src/media/inner.rs
```diff
@@ -622,86 +635,178 @@ impl MediaInner {
...
-                .next_seq_no(now, None);
-
-            Some(NextPacket {
-                pt: pt_rtx,
-                ssrc: ssrc_rtx,
-                seq_no,
-                body: NextPacketBody::Padding { len: padding as u8 },
-            })
+    /// Queue some padding to be sent later.
+    pub fn queue_padding(&mut self, now: Instant, mut pad_size: usize) {
```
k0nserv, line 712: This should probably be renamed `generate_padding`

src/session.rs
```diff
@@ -722,28 +722,29 @@ impl Session {
...
-                } else {
-                    "media"
-                };
+                let kind = if is_padding { "padding" } else { "media" };
 
-                crate::log_stat!("PACKET_SENT", header.ssrc, buf.len(), kind);
+                crate::log_stat!("PACKET_SENT", header.ssrc, payload_size, kind);
             }
 
-            let payload_size = buf.len();
```
k0nserv, line 745: This is a small, but important, change. This meant that before we were accounting for RTP header overhead here, when the intent was to only consider the payload size.

src/media/inner.rs
```diff
@@ -622,86 +641,178 @@ impl MediaInner {
         })
     }
 
-    fn poll_packet_padding(&mut self, now: Instant, pad_size: Option<usize>) -> Option<NextPacket> {
-        // We only produce padding packet if there is an asked for padding size.
-        let pad_size = pad_size?;
+    fn poll_packet_padding(&mut self, now: Instant) -> Option<NextPacket> {
+        loop {
+            let padding_or_resend = self.padding.pop_front()?;
```
algesten, line 646: Name change here.

src/media/inner.rs
```diff
@@ -622,86 +641,178 @@ impl MediaInner {
...
                     });
+                }
+                Padding::Spurious(resend) => {
+                    // If there is no buffer for this padding, we return None. This is
+                    // a weird situation though, since it means we queued padding for a buffer we don't
+                    // have.
+                    let buffer = self
+                        .buffers_tx
+                        .values()
+                        .find(|p| p.has_ssrc(padding_or_resend.ssrc()))?;
```
algesten, line 675: .expect and panic. If this happens we have a problem somewhere.

src/media/inner.rs
```diff
@@ -622,86 +641,178 @@ impl MediaInner {
...
+                    return;
+                };
+                let ssrc_rtx = self
+                    .sources_tx
+                    .iter()
+                    .find(|s| s.is_rtx())
+                    .map(|s| s.ssrc())
+                    .expect("at least one rtx source");
+
+                let padding = pad_size.clamp(SRTP_BLOCK_SIZE, MAX_PADDING_PACKET_SIZE);
```
algesten, line 798: MAX_BLANK_PACKET_SIZE?

src/media/inner.rs
```diff
@@ -622,86 +641,178 @@ impl MediaInner {
...
+
+            // This second scope sends an empty padding packet. This is a fallback strategy if we fail
+            // to find a suitable rtx packet above.
+            // NB: If we cannot generate padding here for some reason we'll get stuck forever.
+            {
+                let Some(&pt) = self.buffers_tx.keys().next() else {
+                    return;
+                };
+                let Some(pt_rtx) = pt_rtx(&self.params, pt) else {
+                    warn!("Found no PT to send raw padding on");
```
algesten, line 788: blank padding

src/media/inner.rs
```diff
@@ -1241,6 +1354,45 @@ impl MediaInner {
...
+                        queued_at,
+                        ..
+                    }) => (
+                        acc.0 + 1,
+                        acc.1 + body_size,
+                        acc.2 + now.duration_since(*queued_at),
+                        if is_first { Some(*queued_at) } else { acc.3 },
+                    ),
+                }
+            });
```
algesten, line 1384: Wouldn't it be cleaner to do `impl From<Padding> for QueueSnapshot` and use the `merge` function we already have for snapshot?

src/media/inner.rs
```diff
@@ -1428,3 +1590,20 @@ fn get_buffer_tx(
 ) -> Option<&PacketizingBuffer> {
     buffers_tx.get(&pt)
 }
+
+/// Separate in wait for polonius.
+fn get_or_create_source_tx<'a>(
```
algesten, line 1595: The other one needs to delegate to this, so there's only one impl.

src/packet/pacer.rs
```diff
@@ -116,6 +116,28 @@ pub struct QueueSnapshot {
     pub last_emitted: Option<Instant>,
     /// Time the first unsent packet has spent in the queue.
     pub first_unsent: Option<Instant>,
+    /// The priority of the most important packet in the queue.
+    pub priority: QueuePriority,
+}
```
algesten, line 121: At the very least we need to move `QueuePriority` up on `QueueState`. The snapshot should just be a statement of fact "this is how the buffer looks", without any judgement calls on priority. That can be introduced on the next level up.

However overall I'm not quite feeling this `QueuePriority` – functionally it's hardly used in Pacer and it complicates the code. It creates a confusion because the _actual_ prio is what `poll_packet` does, i.e. resends first, regular second and padding last.

If we need an indication of which mid contains the most regular and resends (not padding), we could add a secondary counter into `QueueState`, so that we have `padding_count_padding` and `padding_count_regular`.

algesten (approved):
Looks good! Some nice cleanup!

A few comments.

src/packet/pacer.rs
k0nserv: > At the very least we need to move QueuePriority up on QueueState. The snapshot should just be a statement of fact "this is how the buffer looks", without any judgement calls on priority. That can be introduced on the next level up.

We only return a single `QueueState` per `MediaInner`, but that is summarising many internal queues. You are right that it could move to `QueueState`, but it's not a fixed field still, it needs to be calculated based on the actual state of the queues.

> However overall I'm not quite feeling this QueuePriority – functionally it's hardly used in Pacer and it complicates the code. It creates a confusion because the actual prio is what poll_packet does, i.e. resends first, regular second and padding last.

The priority is always media first(whether resend or new) then padding. The pacer must know the state of each queue to avoid requesting padding from one mid when another contains queued media.  

> If we need an indication of which mid contains the most regular and resends (not padding), we could add a secondary counter into QueueState, so that we have padding_count_padding and padding_count_regular.

This is not what we need, the pacer needs to know among all the mids which have media vs those that only contain padding.

src/media/inner.rs
k0nserv: Can't really do `From` but I did something similar to the resends

k0nserv: See https://github.com/algesten/str0m/pull/140/commits/9f7fc94ca6fa60c9bda01ff07729e6dec5ecc1e8

src/packet/pacer.rs
algesten: Ok. Merge away, but I might circle back to this.

algesten (approved):

src/packet/pacer.rs
k0nserv: Sound good to me, the invariant to maintain is:

If we have mid 1 with only padding queued and mid 2 with media and padding queued, pacer must never emit `PollOutcome::PollQueue("1".into())`.

PR #141, opened 2023-04-13
h.264 fixes

k0nserv:
Fixes for h.264 discovered by sending from Windows usin Nvidia HVEC.

### DIFF
--- a/src/packet/h264.rs
+++ b/src/packet/h264.rs
@@ -208,7 +208,7 @@ impl Depacketizer for H264Depacketizer {
         out: &mut Vec<u8>,
         _: &mut CodecExtra,
     ) -> Result<(), PacketError> {
-        if packet.len() <= 2 {
+        if packet.len() == 0 {
             return Err(PacketError::ErrShortPacket);
         }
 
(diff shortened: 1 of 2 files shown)


src/packet/h264.rs
```diff
@@ -208,7 +208,7 @@ impl Depacketizer for H264Depacketizer {
         out: &mut Vec<u8>,
         _: &mut CodecExtra,
     ) -> Result<(), PacketError> {
-        if packet.len() <= 2 {
+        if packet.len() == 0 {
```
algesten, line 211: Why can this be zero?

algesten (approved):

src/packet/h264.rs
k0nserv: It can't ;)

k0nserv: Could write this condition as `packet.len() < 1` if you want

algesten: Right. So what was wrong with <= 2? :)

k0nserv: It errors on payloads of length 1 and 2, which are allowed

algesten: I copied this code from webrtc-rs – they would have the bug too. Maybe we should upstream?

algesten: Oh. I missed that. Thanks.

PR #144, opened 2023-04-14
Simplify pacer further

k0nserv:
Get rid of PollOutcome and generate padding as part of handle_timeout
loop.

### DIFF
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -306,8 +306,6 @@ impl Pacer for LeakyBucketPacer {
         self.maybe_update_adjusted_bitrate(now);
 
         if let Some(request) = self.maybe_create_padding_request() {
-            // We need refreshed queue state to generate a poll for the padding that is generated
-            // as a consequence of this.
             self.next_poll_queue = Some(request.mid);
             return Some(request);
         }
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -306,8 +306,9 @@ impl Pacer for LeakyBucketPacer {
         self.maybe_update_adjusted_bitrate(now);
 
         if let Some(request) = self.maybe_create_padding_request() {
-            // We need refreshed queue state to generate an action to poll the generate padding.
-            self.request_immediate_timeout();
+            // We need refreshed queue state to generate a poll for the padding that is generated
+            // as a consequence of this.
+            self.next_poll_queue = Some(request.mid);
             return Some(request);
         }
 
@@ -470,7 +471,11 @@ impl LeakyBucketPacer {
                 // both the media debt and padding debt to send some padding.
                 let mut drain_debt_time = (self.media_debt / self.adjusted_bitrate)
                     .max(self.padding_debt / self.padding_bitrate);
-                assert!(drain_debt_time > Duration::ZERO);
+                if drain_debt_time.is_zero() {
+                    // Give the main loop some time to doe something else e.g. queue media.
+                    drain_debt_time = Duration::from_micros(1);
+                }
+
                 let padding_at = self
                     .last_handle_time
                     .map(|h| h + drain_debt_time)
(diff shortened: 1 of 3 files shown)


src/packet/pacer.rs
```diff
@@ -280,85 +281,71 @@ impl Pacer for LeakyBucketPacer {
...
 
         self.clear_debt(elapsed);
         self.maybe_update_adjusted_bitrate(now);
-        let should_recalcualte =
-            matches!(self.next_poll_outcome, Some(PollOutcome::Nothing) | None);
 
-        if !should_recalcualte {
-            return;
+        if let Some(request) = self.maybe_create_padding_request() {
+            // We need refreshed queue state to generate an action to poll the generate padding.
```
k0nserv, line 309: Grammar

src/packet/pacer.rs
```diff
@@ -464,62 +450,39 @@ impl LeakyBucketPacer {
...
-            }
-            _ => {
-                // No queues to send on, no padding to generate, wait until the next handle time or
-                // until a packet is queued.
-                (
-                    self.last_handle_time.map(|h| h + PACING).unwrap_or(now),
-                    None,
-                    None,
-                )
+                assert!(drain_debt_time > Duration::ZERO);
```
k0nserv, line 473: Should get rid of this assert, but if this does happen we should return a timeout that's like `1us` so the main loop can have time to queue media if required

src/packet/pacer.rs
```diff
@@ -563,6 +526,59 @@ impl LeakyBucketPacer {
...
+        let all_queues_empty = self
+            .queue_states
+            .iter()
+            .all(|q| q.snapshot.packet_count == 0);
+        if !all_queues_empty {
+            return None;
+        }
+
+        // We must have sent some media.
+        if self.last_send_time.is_none() {
```
k0nserv, line 550: I dunno about clippy here. It suggests

```rust
        // We must have sent some media.
        self.last_send_time?;
```

which just seems harder to read, I'm inclined to ignore it

src/packet/pacer.rs
```diff
@@ -280,85 +281,71 @@ impl Pacer for LeakyBucketPacer {
...
         self.clear_debt(elapsed);
         self.maybe_update_adjusted_bitrate(now);
-        let should_recalcualte =
-            matches!(self.next_poll_outcome, Some(PollOutcome::Nothing) | None);
 
-        if !should_recalcualte {
-            return;
+        if let Some(request) = self.maybe_create_padding_request() {
+            // We need refreshed queue state to generate an action to poll the generate padding.
+            self.request_immediate_timeout();
```
algesten, line 310: Why do we need another timeout? We just got one!

src/packet/pacer.rs
algesten: I think it's ok with `?` – one always needs to be on the lookout for them,.

algesten (approved):
LGTM!

src/packet/pacer.rs
k0nserv: Because at this point the queue state is entirely empty, otherwise we wouldn't have requested padding. Now, to actually send the padding we need a new handle_timeout with a refreshed queue state so we can emit a poll for the relevant queue

k0nserv: I was thinking we could just set `next_poll_queue` here, but I decided against it...

algesten: Since poll_output follows anyway, can't you just say that you want a poll on the `mid` you just generated padding for?

algesten: Why do we need a refreshed queue state?

k0nserv: My concern was around starving out media but maybe that's not gonna be a problem 🤔

k0nserv: Changed it

algesten (approved):

PR #145, opened 2023-04-14
Clear queue state cache on direction change

k0nserv:
Without this the after setting the direction back to send the pacer
would believe it could send padding, but since we clear all the media
state this is not true.

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -1137,6 +1137,8 @@ impl MediaInner {
             self.mid, self.dir, new_dir
         );
 
+        // Clear cache
+        self.queue_state = None;
         self.need_changed_event = true;
 
         let was_receiving = self.dir.is_receiving();


algesten (approved):

src/media/inner.rs
```diff
@@ -1137,6 +1137,8 @@ impl MediaInner {
             self.mid, self.dir, new_dir
         );
 
+        // Clear cache
+        self.queue_state = None;
```
davibe, line 1141: does it make sense to do it where we clear the buffers and only then ? (see a few lines below `self.clear_receive_buffers()` and `self.clear_send_buffers()`)

k0nserv: Doesn't matter much and I'd rather clear one too many times than one too few. This is rare in any case

davibe: I am quite sure the next person calling `clear_*_buffers()` will forget to clear the queue state. Would that result in something potentially bad ?

davibe: ah, okie

k0nserv: > I am quite sure the next person calling clear_*_buffers() will forget to clear the queue state. Would that result in something potentially bad ?

That's probably a good idea, but since Martin is refactoring this already I'll skip it.

> This structure is so maddening. I don't even want to have these kinds of questions. I really need to get my branch sorted and move away from this crazyness.

There are only two hard problems in programming, naming things, cache invalidation, and off by one errors ;)

Issue #153, opened 2023-04-17
Fire ChannelOpen event when sctp channel is ready for pre-negotiated data channels


xnorpx:
While rtc.channel(cid) will return None if not ready 

it's also nice to get a Event once it's ready from  str0m::Event::ChannelOpen(channel_id, label)

PR #155, opened 2023-04-17
Fire ChannelOpen once the SCTP channel is ready for pre neg channels

xnorpx:


xnorpx:
this is fix for #153

algesten (approved):
Nice!

I pushed a small tidy-up to avoid using the DCEP in the out-of-band case.

xnorpx:
lgtm

PR #157, opened 2023-04-18
BWE tidy up and tweaks

algesten:


src/session.rs
```diff
@@ -930,7 +933,18 @@ impl Bwe {
     }
 
     fn poll_estimate(&mut self) -> Option<Bitrate> {
-        self.bwe.poll_estimate()
+        let estimate = self.bwe.poll_estimate()?;
```
k0nserv, line 936: This does a `take` internally thus you could be at `self.last_emitted_estimate * 1.0499` and as a result of this take out the value and throw it away. The frequency is high enough that this probably doesn't matter

src/session.rs
```diff
@@ -240,12 +246,7 @@ impl Session {
         if let Some(padding_request) = self.pacer.handle_timeout(now, iter) {
             let media = self
                 .media_by_mid_mut(padding_request.mid)
-                .unwrap_or_else(|| {
-                    panic!(
-                        "No media found for mid {}, to service padding request",
```
k0nserv, line 245: I liked this better, more context is better

k0nserv (approved):

src/session.rs
algesten: This should be sorted now

k0nserv (approved):

PR #163, opened 2023-04-29
unsynced clocks

xnorpx:


xnorpx:
adding:

pub fn set_str0m_start_time() {
    let mut rtc = RtcConfig::new().build();
    rtc.handle_input(Input::Timeout(Instant::now())).unwrap();
}

in a testsetup solves the issue. No str0m change needed.

PR #167, opened 2023-05-16
Reduce allocation in MediaInner::poll_packet

k0nserv:
Using `resize` meant we spent quite some time in allocation related
code. By using slices instead we avoid this overhead.


davibe (approved):

k0nserv:
This actually matters very little, red herring

davibe:
Screenshot 2023-05-25 at 18 15 29

PR #171, opened 2023-05-25
cache spurious padding candidates

davibe:
I tested locally 1 ingress camera, 3 egress viewers, this reduces `generate_padding()` impact from 6 - 7 % down to 0.3%


src/media/inner.rs
```diff
@@ -746,58 +759,30 @@ impl MediaInner {
...
-                }
-            }
+            if let Some((pt, packet)) = self.spurious_resend_candidate(pad_size) {
+                // Find a historic packet that is smaller than this max size. The max size
+                // is a headroom since we can accept slightly larger padding than asked for.
+                let seq_no = packet
+                    .seq_no
+                    .expect("this packet to have been sent and therefore have a sequence number");
+                // Saturating sub because we can overflow and want to stop when that
+                // happens.
```
algesten, line 769: I think this needs to say _why_ it can overflow.

src/packet/buffer_tx.rs
```diff
@@ -7,6 +7,7 @@ use crate::rtp::{ExtensionValues, MediaTime, Rid, RtpHeader, SeqNo, Ssrc};
 use super::{CodecPacketizer, PacketError, Packetizer, QueueSnapshot};
 use super::{MediaKind, QueuePriority};
 
+#[derive(Clone)]
```
algesten, line 10: I don't want these cloned.

algesten (changes requested):
This is a great find and a good start to fix it, but it requires some changes.

We shouldn't have duplicated Packetized in this extra structure. I think we should try and merge the ordered collection into the queue instead. So the historic_packet_smaller_than is still the entry point and the sorting logic of finding a good candidate is inside there.

k0nserv:
I like the direction, but I don't think repeatedly reusing the same packet for padding is advisable it makes replay attacks easier. This was a problem before too, but if we are adding extra code here I think we should try and avoid that.

libWebRTC uses a priority set of 63 packets and then keeps track of how many times each has been retransmitted. Then they pick the least often retransmitted packet when selecting one. However, their code has a another flaw in that if the selected packet is too large they fall back to generating blank padding packets.

I think a suitable solution would allow us to quickly consider **N** packets for spurious resends, ordered by least retransmissions firstly, and size secondarily. A `BTreeSet` seems suitable

src/packet/buffer_tx.rs
davibe: I don't have another solution

src/media/inner.rs
k0nserv: This is an old comment of mine, it's because we allow providing a packet larger than the request padding i.e. you can have 491 bytes of padding outstanding that is satisfied by a packet that's 600 bytes large.

src/packet/buffer_tx.rs
k0nserv: Storing an index into `PacketizingBuffer::queue` should work, no?

davibe: I intentionally stored up to 60 representative sizes to have a size-diversity you may not find in the current buffer while also removing the need to keep this in sync with it.

src/media/inner.rs
davibe: but because i moved the lines above related to picking the size, it was no longer in context. Now i adjusted

src/packet/buffer_tx.rs
algesten: @davibe can we use the SeqNo?

If each entry have `HashMap<Pt, BTreeMap<usize, SeqNo>>`, the selection can be quick and retrieving the SeqNo not quite as quick, but maybe enough?

Would still be nice to keep it inside buffer_tx though.

davibe: With that solution I would do it in the buffer_tx so it's easier to keep in sync, otherwise you may store a seq_no that's no longer there

algesten: @davibe @k0nserv maybe we can use `Rc` and `Weak`?

I don't want `Arc` inside str0m since it has locking, but `Rc` should be lock free right?

k0nserv: `Arc` doesn't contain a lock, it does use atomic though. Using `Rc` probably makes the whole `Rtc` instance `!Send` which will prevent it from being used with Tokio

algesten: Ah yes. We do want Send Sync.

davibe:
Closing this in favour of #179

PR #176, opened 2023-06-07
MediaData time tweaks

k0nserv:
Ensure `MediaData::network_time` is correct and add a new field `MediaData::buffer_delay` that measures the delay taken to construct the media data from RTP packets

### DIFF
--- a/src/media/event.rs
+++ b/src/media/event.rs
@@ -108,7 +108,7 @@ pub struct MediaData {
     /// For audio the timebase is 48kHz for video it is 90kHz.
     pub time: MediaTime,
 
-    /// The time of the [`Input::Receive`][crate::Input::Receive] that caused this MediaData.
+    /// The time of the [`Input::Receive`][crate::Input::Receive] of the first packet that caused this MediaData.
     ///
     /// In simple SFU setups this can be used as wallclock for [`Writer::write`][crate::media::Writer].
     pub network_time: Instant,
--- a/src/media/event.rs
+++ b/src/media/event.rs
@@ -1,5 +1,5 @@
 use std::ops::RangeInclusive;
-use std::time::Instant;
+use std::time::{Duration, Instant};
 
 use crate::rtp::{Direction, ExtensionValues, MediaTime, Mid, Pt, Rid, SeqNo};
 use crate::sdp::Simulcast as SdpSimulcast;
@@ -112,6 +112,10 @@ pub struct MediaData {
     ///
     /// In simple SFU setups this can be used as wallclock for [`Writer::write`][crate::media::Writer].
     pub network_time: Instant,
+
+    /// The delay caused by building this [`MediaData`] from RTP packets i.e. the difference
+    /// between when the first and last RTP packet that makes up this data was received.
+    pub buffer_delay: Duration,
 
     /// The (RTP) sequence numbers that made up this data.
     pub seq_range: RangeInclusive<SeqNo>,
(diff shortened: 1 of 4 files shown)


src/media/event.rs
```diff
@@ -108,11 +108,15 @@ pub struct MediaData {
...
 
-    /// The time of the [`Input::Receive`][crate::Input::Receive] that caused this MediaData.
+    /// The time of the [`Input::Receive`][crate::Input::Receive] of the first packet that caused this MediaData.
     ///
     /// In simple SFU setups this can be used as wallclock for [`Writer::write`][crate::media::Writer].
     pub network_time: Instant,
 
+    /// The delay caused by building this [`MediaData`] from RTP packets i.e. the difference
+    /// between when the first and last RTP packet that makes up this data was received.
+    pub buffer_delay: Duration,
```
davibe, line 118: Not blocking, just an observation.
This is a "reorder buffer time". It may be strange to see it as "buffer delay" since the buffer delay would extend to the moment the sample is polled out of str0m. For example you can have a very short "reorder buffer time" of one frame, but this frame is waiting for the previous frame to be emitted because the prev is incomplete, so the delay is way higher

algesten: `poll_output` doesn't pass time, so we can't really know when it's polled out, but maybe we can know when all packets in front of this one are dealt with.

k0nserv: Hmm, that's unfortunate. I think @davibe's point and the spec make a compelling case

algesten: The reason is: _when_ something i poll_output, is not a str0m concern. The contract for str0m is that the caller should poll output as soon as possible when indicated. handle_input is when we make these kinds of state changes, so assigning the buffer delay on input is correct from that pov.

algesten: Or turned around, if the caller of str0m has problems guaranteeing poll_output isn't called straight away when indicated, it's not str0m's problem.

k0nserv: Right, the sample can be ready to be pulled out but if the user fails to do so it's not a str0m problem. Could we record the time when a sample is ready to be popped but hasn't been? I guess that would involve shifting some of the code that happens in `pop` to `push` instead

algesten: This is what I mean. Each time we enqueue another incoming packet, we have a time.

```
+----+ +----+     +----+ +----+
| 3  | | 4  |     | 1  | | 2  |
+----+ +----+     +----+ +----+
                               
-------------     -------------
                               
                               
   frame 1           frame 2   
                               ```

The order of RTP packets mean "frame 2" is held back by packet 3 4.

But on packet 4, we can update the time for frame 2.

k0nserv: Yeah that's similar to what I suggested to me. We build the frames pre-emptively on `push` and any time a frame could be popped, either because it's the first frame or because it's preceded by frames that are complete, we use that time as the value for when the frame was completed

k0nserv: If we change the delay calculation to match the spec it would not be misleading though? I see your point with the current logic

k0nserv:
While we figure the bit about the delay out I extracted the fix for network_time to a separate PR so we can land that: https://github.com/algesten/str0m/pull/177

src/media/event.rs
davibe: I guess if we name it for what it is, it does not mislead. If we change the calculation as martin suggested (updating next queued samples when the prev samples completes), we still do not take into account when the samples are pulled out. So it would still seem best to let the user do it.

algesten:
@k0nserv is this still something we should do?

k0nserv:
> @k0nserv is this still something we should do?

We did what @davibe suggested and work around this in application code instead. Doesn't that feel important to address

PR #180, opened 2023-06-13
rtc: config for resend_size_audio/video

algesten:


src/lib.rs
```diff
@@ -1708,6 +1709,44 @@ impl RtcConfig {
...
+    ///
+    /// ```
+    /// # use str0m::Rtc;
+    /// let config = Rtc::builder();
+    ///
+    /// // Defaults to 50.
+    /// assert_eq!(config.resend_size_audio(), 50);
+    /// ```
+    pub fn resend_size_audio(&self) -> usize {
+        self.resend_size_audio
```
xnorpx, line 1728: what happens if I pick 0?

algesten: Setting it too low will probably just break. This configures a size of a ring buffer that is used for outgoing packets.
If you can guarantee a perfect 1 in 1 out (write/poll), you can maybe get away with 1.

For sanity I think something like 10 would feel better.

algesten: That would leave you room in case you end up doing write write write, poll poll poll

algesten: Maybe I should document this.

PR #181, opened 2023-06-14
Restrict periodic BWE adjustments

k0nserv:
We never stopped periodically adjusting BWE based on the current
trendline hypothesis, regardless of how old the hypothesis was. This
meant that if we had a hypothesis of underuse and stopped receiving TWCC
we would indefinitely increase the bandwidth estimate.

With this change we'll only consider a trendline hypothesis valid for a
while after the TWCC report that updated it.

### DIFF
--- a/src/packet/bwe/mod.rs
+++ b/src/packet/bwe/mod.rs
@@ -27,6 +27,8 @@ const MAX_RTT_HISTORY_WINDOW: usize = 32;
 const INITIAL_BITRATE_WINDOW: Duration = Duration::from_millis(500);
 const BITRATE_WINDOW: Duration = Duration::from_millis(150);
 const UPDATE_INTERVAL: Duration = Duration::from_millis(25);
+/// The maximum time we keep updating our estimate without receiving a TWCC report.
+const MAX_TWCC_GAP: Duration = Duration::from_millis(500);
 
 /// Main entry point for the Googcc inspired BWE implementation.
 ///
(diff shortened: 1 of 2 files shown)


src/packet/bwe/mod.rs
```diff
@@ -109,13 +117,25 @@ impl SendSideBandwithEstimator {
             self.mean_max_rtt(),
             now,
         );
+        self.last_twcc_report = now;
+        // We have a valid trendline hypothesis, continue increasing/decreasing the hypothesis for
```
k0nserv, line 121: Reword

src/packet/bwe/mod.rs
```diff
@@ -109,13 +117,25 @@ impl SendSideBandwithEstimator {
...
-        self.last_update + UPDATE_INTERVAL
+        self.last_update + self.poll_interval
     }
 
     pub(crate) fn handle_timeout(&mut self, now: Instant) {
+        if !self.trendline_hypothesis_valid(now) {
+            // We haven't received a TWCC report in a while. The trendline_estimate hyptohesis can
+            // no longer be considered valid. We need another TWCC report before we can update
+            // estimates.
+            self.poll_interval = self.mean_max_rtt().unwrap_or(MAX_TWCC_GAP);
```
k0nserv, line 135: Should not allow this to be less than `UPDATE_INTERVAL`

src/packet/bwe/mod.rs
```diff
@@ -169,6 +189,14 @@ impl SendSideBandwithEstimator {
         // that starves the run loop.
         self.last_update = now;
     }
+
+    fn trendline_hypothesis_valid(&self, now: Instant) -> bool {
+        now.duration_since(self.last_twcc_report)
+            <= self
+                .mean_max_rtt()
+                .map(|rtt| rtt * 2)
+                .unwrap_or(MAX_TWCC_GAP)
```
k0nserv, line 198: Should probably have some minimum value here, ~50ms or so

algesten (approved):

PR #185, opened 2023-06-16
fix endless looping and no data output

davibe:
wip

### DIFF
--- a/src/media/inner.rs
+++ b/src/media/inner.rs
@@ -961,6 +961,7 @@ impl MediaInner {
         if self.to_packetize.is_empty() {
             None
         } else {
+            assert!(self.to_packetize.len() < 2);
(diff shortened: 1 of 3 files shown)


src/packet/pacer.rs
```diff
@@ -441,7 +441,7 @@ impl LeakyBucketPacer {
                 // If we have a non-empty queue send on it as soon as possible, possibly waiting
                 // for the next pacing interval.
                 let drain_debt_time = self.media_debt / self.adjusted_bitrate;
-                let next_send_offset = if drain_debt_time > PACING {
+                let next_send_offset = if drain_debt_time < PACING {
```
k0nserv, line 444: This ain't right.

davibe: Right, I intended media debt the other way around

PR #189, opened 2023-06-19
Fuzzy matching for audio codecs

k0nserv:
Allow matching audio codecs with fuzzy logic i.e. allow disagreements about format parameters such as `useinbandfec` or `minptime`.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -187,19 +187,63 @@ impl PayloadParams {
         let c1 = o.spec;
 
         if c0 == c1 {
+            // Exact match
             return Some(100);
-        } else {
-            // TODO: fuzzy matching.
         }
 
-        // No match
+        // Attempt fuzzy matching, since we don't have an exact match
+        if c0.codec != c1.codec || (c0.codec == Codec::Unknown && c1.codec == Codec::Unknown) {
+            // Codecs must match
+            return None;
+        }
+
+        if c0.clock_rate != c1.clock_rate {
+            // Clock rates must match
+            return None;
+        }
+
+        if c0.channels != c1.channels {
+            // Channels must match
+            return None;
+        }
+
+        if c0.codec.is_audio() && c0.codec == Codec::Opus {
+            return Some(Self::match_opus_score(c0, c1));
+        }
+
+        // TODO: Fuzzy matching for any other audio codecs
+        // TODO: Fuzzy matching for video
+
         None
     }
 
+    fn match_opus_score(c0: CodecSpec, c1: CodecSpec) -> usize {
+        let mut score: usize = 100;
+
+        // If neither value is specified both sides should assume the default value, 3.
+        let either_p_time_specified =
+            c0.format.min_p_time.is_some() || c1.format.min_p_time.is_some();
+        if either_p_time_specified && c0.format.min_p_time != c1.format.min_p_time {
+            score = score.saturating_sub(1);
+        }
+
+        // If neither value is specified both sides should assume FEC is not used as this is
+        // the default.
+        let either_fec_specified =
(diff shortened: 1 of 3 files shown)


src/format.rs
```diff
@@ -189,17 +189,37 @@ impl PayloadParams {
...
+                return None;
+            }
+
+            if c0.channels != c1.channels {
+                // Channels must match
+                return None;
+            }
+
+            if c0.codec.is_audio() {
+                // Accept whatever FEC and p-time setting the remote proposes
```
k0nserv, line 208: This is very liberal i.e. we cannot force FEC or a specific p-time. Do we want some way to determine hard requirements, .e.g `FEC` must be enabled

src/format.rs
```diff
@@ -187,19 +187,57 @@ impl PayloadParams {
...
         if c0 == c1 {
+            // Exact match
             return Some(100);
-        } else {
-            // TODO: fuzzy matching.
         }
 
-        // No match
+        // Attempt fuzzy matching, since we don't have an exact match
+        if c0.codec != c1.codec || (c0.codec == Codec::Unknown && c1.codec == Codec::Unknown) {
```
algesten, line 195: Should this `&&` be `||`?

src/format.rs
```diff
@@ -187,19 +187,57 @@ impl PayloadParams {
...
+                score = score.saturating_sub(1);
+            }
+
+            // If neither value is specified both sides should assume FEC is not used as this is
+            // the default.
+            let either_fec_specified =
+                c0.format.use_inband_fec.is_some() || c1.format.use_inband_fec.is_some();
+
+            if either_fec_specified && c0.format.use_inband_fec != c1.format.use_inband_fec {
+                score = score.saturating_sub(10);
```
algesten, line 225: Could use 2 here. So the sequence goes 1, 2, 4, 8 etc.

src/format.rs
```diff
@@ -187,19 +187,57 @@ impl PayloadParams {
...
+
+        if c0.channels != c1.channels {
+            // Channels must match
+            return None;
+        }
+
+        if c0.codec.is_audio() {
+            let mut score: usize = 100;
+            // If neither value is specified both sides should assume the default value, 3.
+            let either_p_time_specified =
```
algesten, line 213: Wonder if we should break this out into a `match_score_opus`, since the logic here I think it opus specific.

algesten (commented):
Some small things.

src/format.rs
k0nserv: No I don't think so. We bail either if the codecs differ, but in the case that both codecs are unknown we also shouldn't treat that as a match, because we don't know that they are actually equal, only that str0m knows nothing about them

src/format.rs
k0nserv: Sure

src/format.rs
k0nserv: I was thinking about splitting this like that, but I kept it here for clarity since the amount of code is limited. If we start supporting format parameters for other audio codecs we should probably make a function.

Related point: should we have an if-guard to check the codec before this code?

k0nserv:
Addressed all feedback and squashed it into the first commit

algesten (approved):

davibe (approved):

src/format.rs
```diff
@@ -187,19 +187,63 @@ impl PayloadParams {
...
+            return Self::match_opus_score(c0, c1);
+        }
+
+        // TODO: Fuzzy matching for any other audio codecs
+        // TODO: Fuzzy matching for video
+
         None
     }
 
+    fn match_opus_score(c0: CodecSpec, c1: CodecSpec) -> Option<usize> {
```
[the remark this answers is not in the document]
k0nserv: Good points

PR #191, opened 2023-06-20
fix twcc out of order

davibe:


src/packet/bwe/trendline_estimator.rs
```diff
@@ -119,7 +119,15 @@ impl TrendlineEstimator {
             smoothed_delay_ms: self.smoothed_delay,
         };
 
-        self.history.push_back(timing);
+        let pos = self
+            .history
+            .iter()
+            .rev()
+            .position(|p| p.remote_recv_time_ms <= timing.remote_recv_time_ms)
+            .unwrap_or(0);
```
algesten, line 127: Does this risk becoming very inefficient? We have a VecDeque here with iteration for _every single_ TWCC, which is very similar situation to the send buffer performance issues we saw before.

algesten (commented):
Not sure :D

src/packet/bwe/trendline_estimator.rs
davibe: Most of the times the arrival report are in order. This means it will stop at the first element back (its the same as `self.history.back()`), returning a position that is 0. `insert` will act just like `push_back()`. So I would say in the most common case this has negligible impact.

For the few times that arrival_groups are out of order, I still look from the end that is more likely to find a match soon, and stop at the first match. In this case it is less efficient but still better than crashing (that's what happens)

algesten (approved):
Ah yes. We don't like crashing.

src/packet/bwe/arrival_group.rs
```diff
@@ -35,7 +35,9 @@ impl ArrivalGroup {
             ));
         }
 
-        self.last_remote_recv_time = Some(packet.remote_recv_time);
+        self.last_remote_recv_time = self
```
k0nserv, line 38: I'd double check this against libWebRTC. I'm pretty sure this was intentional

PR #192, opened 2023-06-20
safer rtp extension parsing

davibe:


src/rtp/ext.rs
```diff
@@ -457,45 +457,49 @@ impl Extension {
...
                 v.video_timing = Some(self::VideoTiming {
                     flags: buf[0],
-                    encode_start: u16::from_be_bytes([buf[1], buf[2]]),
-                    encode_finish: u16::from_be_bytes([buf[3], buf[4]]),
-                    packetize_complete: u16::from_be_bytes([buf[5], buf[6]]),
-                    last_left_pacer: u16::from_be_bytes([buf[7], buf[8]]),
+                    encode_start: u16::from_be_bytes(buf.get(0..2)?.try_into().unwrap()),
+                    encode_finish: u16::from_be_bytes(buf.get(2..4)?.try_into().unwrap()),
+                    packetize_complete: u16::from_be_bytes(buf.get(4..6)?.try_into().unwrap()),
+                    last_left_pacer: u16::from_be_bytes(buf.get(6..8)?.try_into().unwrap()),
```
algesten, line 502: These ranges look incorrectly translated.

algesten (changes requested):
```rs
- v.transport_cc = Some(u16::from_be_bytes([buf[0], buf[1]]));
+ v.transport_cc = Some(u16::from_be_bytes(buf.get(0..2)?.try_into().unwrap()));
```

I don't like this pattern change. The `[[buf[0], buf[1]]` is used many places else in the code, so we get a consistency.

I prefer to check the length instead.

PR #194, opened 2023-06-27
Prioritize ICE transmits

k0nserv:
ICE transmits are infrequent, but more important than RTP traffic,
therefore they should be prioritized highly when polling for output.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1066,6 +1066,12 @@ impl Rtc {
             }
         }
 
+        // Prioritise ICE transmits because the ICE traffic is responsible for keeping the ICE
+        // connection alive by sending STUN binding requests and responses.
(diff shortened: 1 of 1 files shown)


davibe (approved):

algesten (dismissed):

algesten:
I don't believe this will make any difference. The contract is to keep polling until timeout, which means whether you get ICE events before other events doesn't really matter.

k0nserv:
> I don't believe this will make any difference. The contract is to keep polling until timeout, which means whether you get ICE events before other events doesn't really matter.

hmmmmmmmm. This is true

algesten (changes requested):
After thinking about it, I don't see the point.

k0nserv:
> After thinking about it, I don't see the point.

Yeah you're right, at best we send a bit earlier

PR #195, opened 2023-06-27
Add egress loss stat

k0nserv:
Add an egress loss stat to `PeerStats` derived from TWCC reports.


src/rtp/rtcp/twcc.rs
```diff
@@ -1088,6 +1088,7 @@ impl TwccSendRegister {
             );
         }
 
+        println!("TWCC loss: {:?}", self.loss(Duration::from_secs(1), now));
```
algesten, line 1091: ?

src/rtp/rtcp/twcc.rs
```diff
@@ -1097,6 +1098,53 @@ impl TwccSendRegister {
...
+            (total + 1, lost + u64::from(was_lost))
+        });
+
+        if total == 0 {
+            return None;
+        }
+
+        Some((lost as f64) / (total as f64))
+    }
+
```
algesten, line 1147: This looks a bit expensive. Is it?

src/rtp/rtcp/twcc.rs
k0nserv: Soz, left from debugging

src/rtp/rtcp/twcc.rs
k0nserv: Shouldn't be, it's `O(2n)` at worst with `n == 1000` currently and runs every second. I don't really think the `2n` case will happen either

src/stats.rs
```diff
@@ -68,6 +70,8 @@ pub struct PeerStats {
     pub timestamp: Instant,
     /// The last egress bandwidth estimate from the BWE subsystem, if enabled.
     pub bwe_tx: Option<Bitrate>,
+    /// The egress loss over the last second expressed as a percentage value between 0.0 and 1.0.
```
xnorpx, line 73: Fraction or ratio?

xnorpx: or just multiply by 100 and make it an u8?

k0nserv: Hmm, I think keeping it as f64 would be nice but I can reword it

algesten: f32?

k0nserv: No preference, I just defaulted to f64. Do we have a stated preference for the project?

algesten: i think it depends on what we're talking about. a percentage number between 0.0-1.0 might not need the precision of f64.

k0nserv: Fair, f32 should be sufficient. Will change

xnorpx: also if you say egress_loss_ratio it is clear that you return a ratio and I will expect numbers between 0 and 1

k0nserv: > also if you say egress_loss_ratio it is clear that you return a ratio and I will expect numbers between 0 and 1

Good point, although I know @algesten isn't a fan of long names :D

src/rtp/rtcp/twcc.rs
```diff
@@ -1811,4 +1858,56 @@ mod test {
...
+                .into(),
+            },
+            now,
+        )
+        .expect("apply_report to return Some(_)");
+
+        now = now + Duration::from_millis(20);
+        let loss = reg
+            .loss(Duration::from_millis(150), now)
+            .expect("Should be able to calcualte loss");
```
xnorpx, line 1907: calculate

src/stats.rs
xnorpx: comment still percentage

xnorpx:
lgtm

algesten (approved):

PR #196, opened 2023-06-28
Simplify TWCC egress loss

k0nserv:
Considering the RTT and likelihood that the TWCC report has been sent
and received is a tad more robust but it can cause spurious incorrect
values instead. This simplified version simply ignores packets that we
haven't received a TWCC report for. If a whole TWCC report is lost we'll
have less data, but as long as the loss is somewhat uniform this
shouldn't matter much.


algesten (approved):

davibe:
This is considering packets sent in the last second, excluding the ones that have not been answered. So with a RTT of 200 ms we only consider 0.8s worth of packets, loosing 0.2s every second ? or am I reading it wrong ?

k0nserv:
Yeah, packets that haven't had their TWCC reports answered are ignored which reduces the span we consider. 

The reason I started looking at this is because the previous code produced false positives, reporting loss when there was none. Ultimately we don't have a good value for RTT available here, the `TwccSendRecord::rtt` is a bit flawed.

This could be better, but I think this is good enough

PR #197, opened 2023-06-28
Add ingress loss to PeerStats

k0nserv:
- Add TwccRecvRegister::loss
- Add ingress loss to PeerStats

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -22,6 +22,7 @@ pub(crate) struct StatsSnapshot {
     pub tx: u64,
     pub rx: u64,
     pub egress_loss_ratio: Option<f32>,
+    pub ingress_loss_fraction: Option<f32>,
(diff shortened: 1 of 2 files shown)


davibe (approved):

src/stats.rs
```diff
@@ -72,6 +74,8 @@ pub struct PeerStats {
     pub bwe_tx: Option<Bitrate>,
     /// The egress loss over the last second.
     pub egress_loss_fraction: Option<f32>,
+    /// The ingress losss since the last stats event.
```
k0nserv, line 77: There's a discrepancy between the time period between egress and ingress loss now. We could address this by making ingress loss also emit every second to match egress.

For egress using data since last event is trickier when the stats interval is small, there's a risk it exceeds RTT and we have no data

PR #199, opened 2023-06-28
vp8: avoid waiting for intentionally unmet NACKs

davibe:
Chrome does not seem to answer nacks for frames on vp8 temporal layer 1 (since they are not vital). This means with some loss on the connection we introduce the highest delay possible (w.r.t. the configured amount of buffered video samples).

Here we aim at avoiding this

### DIFF
--- a/src/packet/buffer_rx.rs
+++ b/src/packet/buffer_rx.rs
@@ -3,7 +3,10 @@ use std::fmt;
 use std::ops::RangeInclusive;
 use std::time::Instant;
 
-use crate::rtp::{ExtensionValues, MediaTime, RtpHeader, SeqNo};
+use crate::{
+    change::DirectApi,
+    rtp::{ExtensionValues, MediaTime, RtpHeader, SeqNo},
+};
 
 use super::{CodecDepacketizer, CodecExtra, Depacketizer, PacketError};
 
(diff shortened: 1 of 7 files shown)


src/packet/buffer_rx.rs
```diff
@@ -146,49 +148,132 @@ impl DepacketizingBuffer {
...
+
+            const VP8_DEPACK_OPTIMIZATION: bool = true;
+
+            match (last_codec_extra, dep.codec_extra) {
+                (CodecExtra::Vp8(prev), CodecExtra::Vp8(next)) => {
+                    if !VP8_DEPACK_OPTIMIZATION {
+                        break 'contig false;
+                    }
+
+                    // In the case of VP8 chrome does answer nacks for frames that are on
```
k0nserv, line 187: does –> doesn't

src/packet/buffer_rx.rs
```diff
@@ -146,49 +148,132 @@ impl DepacketizingBuffer {
...
+                    // missing frame on layer 1 that its contiguous to two frames on layer 0
+
+                    let Some(prev_pid) = prev.picture_id else {
+                        break 'contig false;
+                    };
+                    let Some(next_pid) = next.picture_id else {
+                        break 'contig false;
+                    };
+
+                    let allowed = prev.layer_index == 0
```
k0nserv, line 198: This logic is only correct if we have two temporal layers, right? If we had a single temporal layer we'd be missing a picture from it here

k0nserv: Also we assume that VP8 frames arrive in order always, is that specified in the RTP VP8 RFC?

src/packet/buffer_rx.rs
```diff
@@ -146,49 +148,132 @@ impl DepacketizingBuffer {
...
+
+            let Some((last_seq, last_codec_extra)) = self.last_emitted else {
+                break 'contig true;
+            };
+
+            const VP8_DEPACK_OPTIMIZATION: bool = true;
+
+            match (last_codec_extra, dep.codec_extra) {
+                (CodecExtra::Vp8(prev), CodecExtra::Vp8(next)) => {
+                    if !VP8_DEPACK_OPTIMIZATION {
```
k0nserv, line 183: Maybe we extract this VP8 specific stuff to a function or make this whole thing polymorphic based on the codec, both the cache and this extra parsing is codec specific after all

src/packet/buffer_rx.rs
k0nserv: What happens if we get a subsequent L1 frame here i.e.

```
F0(L0), F2(L0), F3(L1), F4(L0)
```

The consumer will see this as contiguous despite F1 being missing. Presumably the consumer should not use F3 here?

davibe: yes, we only have 2 temporal layer in webrtc. Chrome has experimented 4 but then reverted to 2. I don't remember where this is documented.

I don't think we assume they arrive in order. Our segments are in order.

As far as we know vp8 layers are interleaved L0, L1, L0, L1, .. 

Is the missing F1 on Layer 1 ? In that case the decoder (should) understand that F3 depends on a missing F1 and therefore skip it. If this is not the case, it just means that we'd have to suppress L1 frames until the next L1 marked as sync passes through. Right now latency has improved and the stream is mostly ok considering the amount of loss. I did see minor glitches (but very hard to see and repro) so I may investigate this more when I am able to reproduce better.

src/packet/buffer_rx.rs
davibe: I did think about delegating to isolated codec-specific logic, maybe through a trait .. but to be honest I would do that when we support 2 or 3 codecs like this. At the moment we are not even sure this is a good idea long term

src/packet/buffer_rx.rs
davibe: thx

davibe:
with a few more bytes (i guess) I started triggering this

https://github.com/algesten/str0m/actions/runs/5403814860/jobs/9817154509?pr=199

While it's a good idea to keep things tiny, I guess this disparity between enum cases is a also by design . Should I just suppress ?

src/packet/buffer_rx.rs
k0nserv: > I don't think we assume they arrive in order. Our segments are in order.

I mean after we have reordered them from RTP, are the VP8 frames in order then?

> Is the missing F1 on Layer 1 ? In that case the decoder (should) understand that F3 depends on a missing F1 and therefore skip it. If this is not the case, it just means that we'd have to suppress L1 frames until the next L1 marked as sync passes through. Right now latency has improved and the stream is mostly ok considering the amount of loss I simulate. I did see minor glitches (but very hard to see and repro) so I may investigate this more when I am able to reproduce better.

Cool, that's what I imagined

src/packet/buffer_rx.rs
k0nserv: Perhaps you are right, a function in the interim?

k0nserv (approved):

src/packet/buffer_rx.rs
davibe: okie

PR #201, opened 2023-06-29
API design for RTP mode API

algesten:


src/rtp_api.rs
```diff
@@ -0,0 +1,64 @@
...
+    ///   where the SSRC is not communicated in a side channel.
+    /// * Nackable indicates whether this packet can be nacked by the remote and resent. This is always
+    ///   false for audio, but should also be set to false when the RTP payload is a discardable temporal
+    ///   layer (such as layer L1 in VP8).
+    ///
+    /// Regarding `wallclock` and `rtp_time`, the wallclock is the real world time that corresponds to
+    /// the `MediaTime`. For an SFU, this can be hard to know, since RTP packets typically only
+    /// contain the media time (RTP time). In the simplest SFU setup, the wallclock could simply
+    /// be the arrival time of the incoming RTP data (see
+    /// [`MediaData::network_time`][crate::media::MediaData]). For better synchronization the SFU
```
k0nserv, line 41: To do this you really need the SR reports I think. Maybe we need to consider having str0m emit the SR, or at least data derived from them, as events?

algesten: that's a separate problem which doesn't belong in this api design. that comment was lifted from current docs for writer.

PR #203, opened 2023-06-30
Pull RtxCache out of PacketizingBuffer and make it evict in better ways

pthatcher:


src/media/inner.rs
```diff
@@ -1652,3 +1650,137 @@ fn get_or_create_source_tx<'a>(
...
+    fn find_first_seq_no_thats_too_old(&self, now: Instant) -> Option<SeqNo> {
+        if self.packet_by_seq_no.len() > self.max_packet_count {
+            let first_seq_no = self.packet_by_seq_no.keys().next()?;
+            // Too old because of max_packet_count.
+            return Some(first_seq_no.clone());
+        }
+        // If the max_packet_age is so old that checked_sub returns None, we shouldn't remove based on max_packet_age.
+        let min_queued_at = now.checked_sub(self.max_packet_age)?;
+
+        let (first_seq_no, first_packet) = self.packet_by_seq_no.iter().next()?;
```
algesten, line 1741: This looks a bit like what PacketizingBuffer did before we refactored it. We used to have it as a VecDeque and have SeqNo (extended seq) as the identifier. To find elements, we used `iter()` in various situations, and that's where I see similarities.

The use of `.iter()` on every send, lookup for spurious padding, calculation of queue time for the pacer, cache eviction, became a bottle neck  – it used too much CPU. This is why we changed to a ringbuffer where the `Ident` is a key into the ringbuffer that uses modulo is an instant lookup.

Iterators could work if there are less elements to consider in the first place. @xnorpx  mentioned that we could use TWCC receipts to evict from these buffers, which probably means the amount of elements would be low enough that iterating is not a problem.

algesten (commented):
Let's reason about this PR in three ways.

1. Teasing out the RtxResendBuffer part of the PacketizingBuffer, since resends have nothing to do with packetizing.
2. Limiting the RtxResendBuffer on time as well as amount of packets.
3. Name changes/cleanup etc

On point 1

Overall I think we should change PacketizingBuffer to do what we want instead of introducing a new buffer.

In my mind, PacketizingBuffer has gradually morphed to the point where I think of it as "Send Buffer" – i.e. the "packetizing" part is neither here nor there. Splitting a sample into RTP packets is a single `fn`, for the rest it works on RTP payloads.

I think we should rename `PacketizingBuffer` -> `SendBuffer` and extract the sample splitting business up one level.

The `SendBuffer` serves two purposes, as it already does today, 1) queue of waiting RTP packets to be sent 2) buffer of packets for handling NACKs.

On point 2, limiting buffer on time.

Once we done part 1, we can make the `SendBuffer` have a limit on time as well as length. However I think that needs to built on top of the `RingBuffer` that we got. We switched away from VecDeque (etc), because operations in the "hot path" (every packet sent), required some sort of iter() (keys(), values()) – with our own ring buffer we made those instant lookups. The old structures used too much CPU.

We could explore going back to iterator-style if we combine it with ways of keeping the buffer small (aggressively evicting using TWCC acks and/or keep a low timeout). Though it feels a bit like a step back, I could be convinced.

Performance is also the reason we've made an effort to not clone any packets internally. Omitting `#[Clone]` from `Packetized` is a very deliberate choice (to the point where I should have documented it – my bad). To me, losing `<'a>` on NextPacket, is a regression. But if we want to step away from that principle, we need to have a discussion about why doing so is worth it.

Point 3. Name changes.

I'm don't like "drive by name changes", i.e. where the PR sets out to do _something else_, but also changes variable names of half related things. Name changes are fine, but I'd like that to be separate discussions in separate PRs. This PR is mainly about point 1 and 2.

I'm not arguing the existing names are better, most of these changes make total sense, but I want to deal with that in a separate PR.

pthatcher:
About the name changes and cleanups: yes, those should be in a separate PR.  That why this is a draft PR: just to get an idea if it's the right direction before trying to break things up into manageable pieces.

On cloning a Packet: Sure, I just changed it back.  It no longer clones it.  BTW, I noticed the payloads being copied twice when doing SRTP protect when only one copy is needed.  I think we should fix that.

On PacketizingBuffer vs. SendBuffer: I agree we should tease apart the "sending" part and the "packetizing" part.  But since almost all of the "sending part" is doing RTX (the rest is basically just keeping track of stats of a FIFO), I focused on the RTX part first.

About the VecDeque vs. RingBuffer and "iterator style": if you have a faster FIFO than VecDeque, that's fine with me, the key thing I'm trying to narrow in on with this PR is to not have one buffer for both initial send and RTX purposes.  Have a map cache for the RTX and a FIFO for the initial send.  Honestly, I have a hard time believing that a VecDeque as a FIFO is going to be a critical part of performance when all you do is push on one end and pop off of the other.  I agree avoiding iterations is important, but in my PR, I'm not iterating on the VecDeque.  So I don't think there's any regression into "iterator style".

But all of the above are side issues.  The main issue is that the send buffer is combining two things which should not be combined: the RTX cache and the send queue.  That's the major point of this PR, to split that up. 


> Let's reason about this PR in three ways.
> 
> 1. Teasing out the RtxResendBuffer part of the PacketizingBuffer, since resends have nothing to do with packetizing.
> 2. Limiting the RtxResendBuffer on time as well as amount of packets.
> 3. Name changes/cleanup etc
> 
> On point 1
> 
> Overall I think we should change PacketizingBuffer to do what we want instead of introducing a new buffer.
> 
> In my mind, PacketizingBuffer has gradually morphed to the point where I think of it as "Send Buffer" – i.e. the "packetizing" part is neither here nor there. Splitting a sample into RTP packets is a single `fn`, for the rest it works on RTP payloads.
> 
> I think we should rename `PacketizingBuffer` -> `SendBuffer` and extract the sample splitting business up one level.
> 
> The `SendBuffer` serves two purposes, as it already does today, 1) queue of waiting RTP packets to be sent 2) buffer of packets for handling NACKs.
> 
> On point 2, limiting buffer on time.
> 
> Once we done part 1, we can make the `SendBuffer` have a limit on time as well as length. However I think that needs to built on top of the `RingBuffer` that we got. We switched away from VecDeque (etc), because operations in the "hot path" (every packet sent), required some sort of iter() (keys(), values()) – with our own ring buffer we made those instant lookups. The old structures used too much CPU.
> 
> We could explore going back to iterator-style if we combine it with ways of keeping the buffer small (aggressively evicting using TWCC acks and/or keep a low timeout). Though it feels a bit like a step back, I could be convinced.
> 
> Performance is also the reason we've made an effort to not clone any packets internally. Omitting `#[Clone]` from `Packetized` is a very deliberate choice (to the point where I should have documented it – my bad). To me, losing `<'a>` on NextPacket, is a regression. But if we want to step away from that principle, we need to have a discussion about why doing so is worth it.
> 
> Point 3. Name changes.
> 
> I'm don't like "drive by name changes", i.e. where the PR sets out to do _something else_, but also changes variable names of half related things. Name changes are fine, but I'd like that to be separate discussions in separate PRs. This PR is mainly about point 1 and 2.
> 
> I'm not arguing the existing names are better, most of these changes make total sense, but I want to deal with that in a separate PR.

src/media/inner.rs
pthatcher: Yes, calling .iter() is a bad idea.  In my PR, I do not do any .iter().  Cached packets (for RTX) on done by SeqNo lookup in a map, or (for padding) by a bucket size.  And the FIFO for sending doesn't do any lookups: just push on one end and pop on the other.  

I agree .iter()s should be removed, especially in the hot path.  This PR accomplishes that, but by splitting the roles of FIFO and RTX cache into two separate places instead of them being done in one place.

src/media/inner.rs
```diff
@@ -578,22 +599,24 @@ impl MediaInner {
...
             // a weird situation though, since it means the other side sent a nack for
-            // an SSRC that matched this Media, but didn't match a buffer_tx.
-            let buffer = self.buffers_tx.values().find(|p| p.ssrc() == resend.ssrc)?;
+            // an SSRC that matched this Media, but didn't match n RTX cache
+            let rtx_cache = self.rtx_cache_by_ssrc.get(&resend.ssrc)?;
 
-            let pkt = buffer.get(resend.seq_no);
+            let pkt = rtx_cache
+                .get_cached_packet_by_seq_no(resend.seq_no)
+                .cloned();
```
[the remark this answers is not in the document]
pthatcher: I put this back.  It no longer clones.

src/media/inner.rs
algesten: What are the benefits of splitting FIFO (send queue) and RTX?

src/media/inner.rs
```diff
@@ -1652,3 +1661,137 @@ fn get_or_create_source_tx<'a>(
...
+
+    fn get_cached_packet_smaller_than(&self, max_size: usize) -> Option<(SeqNo, &Packetized)> {
+        let quantized_size = max_size / RTX_CACHE_SIZE_QUANTIZER;
+        let seq_no = self
+            .seq_no_by_quantized_size
+            .range(..=quantized_size)
+            .next_back()?
+            .1
+            .clone();
+        Some((seq_no, self.get_cached_packet_by_seq_no(seq_no)?))
```
k0nserv, line 1717: This will return None if the largest packet found has been evicted already, I think this would be better:


```rust
self
            .seq_no_by_quantized_size
            .range(..=quantized_size)
            .rev()
            .find_map(|(_, seq_no)| self.get_cached_packet_by_seq_no(seq_no))
            .cloned()

```

From our tests it seems preferable to use spurious RTX for padding than generate blank packets.

pthatcher:
I think this is obselete, so I'll close it.

PR #204, opened 2023-06-30
buffer_tx: add configurable duration threshold

davibe:
Allows the buffer to free packets that are older than a configured threshold. This refers to RTP time and acts within the boundaries of the buffer size without changing the size of the buffer itself.

When packets are evicted the Packetized::data Vec<u8> is dropped, therefore freeing up memory. This will be more efficient as we reduce the overhead of the Packetized structure itself.

Having a larger size and limiting by time allows to more efficiently accomodate spikes due to big keyframes (i.e. think of screenshares).

### DIFF
--- a/src/packet/buffer_tx.rs
+++ b/src/packet/buffer_tx.rs
@@ -45,6 +45,7 @@ pub struct PacketizingBuffer {
     emit_next: Ident,
     last_emit: Option<Instant>,
     max_retain: usize,
+    max_retain_time: Duration,
 
     total: TotalQueue,
 
@@ -55,7 +56,7 @@ pub struct PacketizingBuffer {
 const SIZE_BUCKET: usize = 25;
 
 impl PacketizingBuffer {
-    pub(crate) fn new(pack: CodecPacketizer, max_retain: usize) -> Self {
+    pub(crate) fn new(pack: CodecPacketizer, max_retain: usize, max_retain_time: Duration) -> Self {
         PacketizingBuffer {
             pack,
             queue: RingBuf::new(max_retain),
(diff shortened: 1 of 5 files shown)


davibe:
I had this patch laying around. I saw the feature may be of interest from other conversations around #203 so I resumed it. It's a more incremental take to this concept of timed buffer_tx which retains fixed allocation of the "ring buffer".

src/packet/buffer_tx.rs
```diff
@@ -309,6 +315,34 @@ impl PacketizingBuffer {
...
+        let Some(last) = self.queue.last() else { return };
+
+        let Some(last_ident) = self.queue.last_ident() else { return };
+        let Some(mut ident) = self.queue.first_ident() else { return };
+        let recent = last.meta.rtp_time.as_seconds();
+
+        loop {
+            if ident == last_ident {
+                break;
+            }
```
pthatcher, line 329: Nit: could just do "while ident != last_ident"

src/packet/buffer_tx.rs
```diff
@@ -309,6 +315,34 @@ impl PacketizingBuffer {
...
+
+            let Some(packet) = self.queue.get(ident) else { continue};
+            let packet_time = packet.meta.rtp_time.as_seconds();
```
pthatcher, line 332: Don't you want to use the queue_time and not the rtp_time?  If you're writing and SFU, the rtp_time is time set by the original client, and you don't have any idea how that relates to your server's clock.

davibe: We have no idea how RTP relates to the server clock, but since we only care about the `elapsed delta` it is "fine". With this approach time is driven forward by the arrival of new packets, so you get the effect of potentially holding for longer if no new data arrives. I do not see something "wrong" though.

PR #206, opened 2023-06-30
Truncate off SRTP padding after decryption

pthatcher:


algesten (approved):
Great catch! Thanks!

algesten:
@pthatcher I think we need a `cargo fmt` here.

pthatcher:
cargo fmt is done

PR #208, opened 2023-06-30
Implement RtxCache and use it inside of PacketizingBuffer

pthatcher:


src/media/inner.rs
```diff
@@ -543,6 +551,15 @@ impl MediaInner {
                 len
             }
         };
+        let pt = next.pt;
+        let seq_no = next.seq_no;
+        if matches!(next.body, NextPacketBody::Regular { .. }) {
+            if let Some(buffer_tx) = self.buffers_tx.get_mut(&pt) {
+                if let Some(pkt) = buffer_tx.take_next(now) {
```
algesten, line 558: Changing `take_next` from `Packetized` to `Option<Packetized>` introduces the possibility of an incorrect state.

Since `maybe_packetized` returns `Option<&Packetized>`, we get a `Some(packet)` back when there is a packet to send. If `maybe_next` returns `Some` but the following `take_next` returns `None`, something is seriously wrong with the logic. This `if let Some(pkt)` masks that potential problem.

I think we should either continue with `take_next` returning a non-optional `Packetized`, or do an `.expect()` force unwrap here. Either way we want a panic for this inconsistency.

Another solution: Since the `rtx_cache` is an internal detail to `PacketizingBuffer`, we could also move the `packetized` from one buffer to the other inside `take_next`, avoiding this state inconsistency. We would then retain `&Packetized` signature of `take_next`.

algesten (changes requested):
Overall looks good. PacketizingBuffer does look cleaner this way.

One thing I picked up on.

src/packet/buffer_tx.rs
```diff
@@ -396,7 +359,342 @@ impl fmt::Debug for Packetized {
...
+        // There is no packet, so I guess we'll clear it.  But that's a no-op anyway.
+        let Some(first_packet) = self.packet_by_seq_no.values().next() else {
+            return Some(None)
+        };
+
+        if first_packet.queued_at <= min_queued_at {
+            return Some(
+                self.packet_by_seq_no
+                    .iter()
+                    .find(|(_, packet)| packet.queued_at > min_queued_at)
```
davibe, line 487: I believe it’s safer to evict based on media time (rtp time). For example in a SFU scenario It protects us from jitter / latency / drift making the buffer independent from those. A couple of extremes just for reasoning
- 1s worth of media can be queued in 20ms and vice versa
- In a SFU scenario (origin -> SFU -> palyer) if the origin drifts behind, the player slows down, the SFU should not shorten the buffer time, and vice versa
- with screenshares you have large frames (high latency), and very low fps (down to one every few secs). There is a higher chance for the player to nack an evicted buffer which would result in PLIs

src/media/inner.rs
pthatcher: Yes, you have a good point.  I originally tried to do the "take_next moves between buffers" thing, but found it awkward because it provides a seq_no when taking.  But I guess if I called it "cache_next_as_sent" or something like that it could work.

I think a better approach would be to do what I did in #203: have maybe_next be take_next instead.  I was trying to minimize the diff here, but maybe I want too far.  Let me take a shot at going that direction and see how it goes.

src/packet/buffer_tx.rs
pthatcher: For a relatively large time duration (say, 3 seconds), it doesn't really matter if you're using the RTP timestamp or the send time.  After that long of a time, it won't be useful either way. 

If you use the RTP time, then you have to do clock synchronization between the SFU's clock and the RTP timestamps, and that can be difficult.  And even if you get it right, the net result is that you'll be choosing to keep packets in the RTX cache for less time in the case that they arrive late relative to the others.  So the RTX cache gets a little smaller?  If that's the goal, there are many other ways to do it that are better (ACK-based cache eviction, dynamic duration based on a histogram of NACK times, etc).  

So I don't really think it's worth the trouble of the clock synchronization when it's easier to just make the time duration somewhat large and have other mechanisms for keeping the RTX cache small.

src/media/inner.rs
pthatcher: OK, I tried that route and it seems fine, except for a little duplicate code between RTX and non-RTX packets.  Let me know what you think.

pthatcher:
I resolved the merge conflict.  Can I get another round of review?

algesten (approved):

PR #213, opened 2023-07-03
Don't blow up when sending empty packets

pthatcher:


algesten (approved):
Ah yes. Empty packets.

PR #214, opened 2023-07-05
fix detecting opus head / tail for padding

davibe:
Since padding is now truncated, we are getting a depacketising error "Packet too short".


algesten:
There should be no marker in opus.

davibe:
> There should be no marker in opus.

fixed while you were writing :)

algesten (approved):

PR #218, opened 2023-07-21
Allow sending of padding before anything else

pthatcher:


k0nserv:
## Padding
The idea that we don't send padding until we have sent at least some media was ported from libWebRTC, you should be able to find comments that talk about this in there. From observation, using empty padding packets is more expensive(because of per packet overhead) than spurious resends too. 

## Padding rate
The change to the padding rate is not a good idea, at least not under the current system. The idea is that you set two values the bitrate you are currently expecting to send(for pacing) and the desired bitrate you want to send eventually, the system then slowly ramps up to the desired rate using a mix of padding and media. For example, if you have 10 video streams where the highest layer is 1.5Mbit/s the desired bitrate would be 15Mbit/s. With this change someone joining this room would immediately receive 15Mbit/s of(mostly padding), potentially over saturating their link.

What would happen today under the same example is you start out with all the low spatial layers `250kbit/s * 10 = 2.5Mbit/s` and then as the BWE increases the total bitrate is filled in with padding to prove that the BWE is accurate. Eventually BWE is high enough that one or more streams can be upgraded to the next layer, reducing the amount of padding used. My feeling is that we want to optimise the media-to-padding ratio to be low, sending a lot of padding is wasteful and quite CPU intensive(even with recent performance improvements).

All that said, I appreciate your insight and if you think this should work differently I'd love to hear those thoughts

pthatcher:
Just because limiting sending of padding to after there is media to send makes sense for libwebrtc, does not mean it makes sense for all uses of str0m.  For example, when creating an SFU you may want to be able to get the BWE going before you have something to forward.  And libwebrtc wasn't written for SFUs.  And this behavior is completely opt-in.  If the user of str0m doesn't call set_desired_bitrate or sets it to 0, this "padding before sending" doesn't happen.

About the padding rate:  yes, I know that sending lots of padding may be inefficient.  In real world cases, you'd set_desired_bitrate to something relatively small (400kbps?) and only do it for a brief period of time (a few seconds?).   But I think that's actually orthogonal to the change I made.  The change I made just allows padding to be sent even before there is an BWE.  But I suppose we could use the initial bitrate for that instead of the "current rate".

This would be better with probing, but then the same logic applies: you'd want to do probing before you have anything to forward to get the BWE going.

k0nserv:
> Just because limiting sending of padding to after there is media to send makes sense for libwebrtc, does not mean it makes sense for all uses of str0m. For example, when creating an SFU you may want to be able to get the BWE going before you have something to forward. And libwebrtc wasn't written for SFUs. And this behavior is completely opt-in. If the user of str0m doesn't call set_desired_bitrate or sets it to 0, this "padding before sending" doesn't happen.

That depends on why that limit exists in libWebRTC. To be honest I do not know why they have the restriction, my thinking is that it might be more consequential than it just not making sense, i.e. it might break some assumption on the part of the receiver.

> About the padding rate: yes, I know that sending lots of padding may be inefficient. In real world cases, you'd set_desired_bitrate to something relatively small (400kbps?) and only do it for a brief period of time (a few seconds?). But I think that's actually orthogonal to the change I made. The change I made just allows padding to be sent even before there is an BWE. But I suppose we could use the initial bitrate for that instead of the "current rate".

This doesn't really match how the API is intended to be used and how we use it at the moment. Here's how it currently works:

You set **desired_bitrate** to the bitrate you'd need to send all video media at the highest quality. You set **current_bitrate** to the bitrate of the video media you are currently sending. The BWE system makes up the difference between `current_bitrate` and the estimate with padding(to ensure that the estimate is actually valid). 

This change:

```diff
-        let padding_rate = bwe
-            .last_estimate()
-            .map(|estimate| estimate.min(bwe.desired_bitrate))
-            .unwrap_or(Bitrate::ZERO);
+        let padding_rate = bwe.desired_bitrate.min(bwe.current_bitrate);
```

breaks this behaviour. We are no longer padding to the estimate which makes the estimate unproven and invalid. 

What I think we might need here is a two layered API, what you are suggesting is more akin to the initial design of the API where the user had direct control over the padding rate.

pthatcher:
I think this is obsolete, so I'll close it.

PR #223, opened 2023-07-31
.last_key_value() is not in MSRV 1.65

algesten:


src/streams/rtx_cache.rs
```diff
@@ -133,8 +135,8 @@ impl RtxCache {
 
     pub(crate) fn last_packet(&self) -> Option<&[u8]> {
         self.packet_by_seq_no
```
k0nserv, line 137: `self.packet_by_seq_no.iter().rev().next().map(|e| e.1.payload.as_ref())` and you can avoid the new field

algesten: Good point. Let's do that.

k0nserv (approved):
LGTM

PR #226, opened 2023-07-31
Fix incidental packet modification

k0nserv:
Modifying the pt when doing RTX meant subsequent resends of the same packet would fail
to be sent.


k0nserv:
f

algesten (approved):

PR #227, opened 2023-08-01
Fake queue stats when there is padding

algesten:


src/streams/send.rs
```diff
@@ -680,6 +681,28 @@ impl StreamTx {
...
+            } else {
+                QueuePriority::Media
+            };
+
+            let fake_packets = self.padding / AVERAGE_PADDING_PACKET_SIZE;
+
+            snapshot.size += self.padding;
+            snapshot.packet_count += fake_packets as u32;
+
+            // TODO: Should this be faked in relation to size of padding?
```
k0nserv, line 702: My removal of the VecDeque we used for `self.padding` probably doesn't help here. If we still had that we could keep a `queued_at: Instant` field with every padding request and produce a better value here.

To answer the TODO: Shouldn't it be scaled with the number of packets?

src/streams/send.rs
```diff
@@ -680,6 +681,28 @@ impl StreamTx {
         //    first ever poll_packet().
         let use_for_padding = self.is_audio.map(|i| !i).unwrap_or(false);
 
+        let mut snapshot = self.send_queue.snapshot(now);
+
+        // Fake some stats for padding.
+        if self.padding > 0 {
+            // TODO: Be more scientific about this factor.
+            const AVERAGE_PADDING_PACKET_SIZE: usize = 200;
```
k0nserv, line 689: At the very least I think we can make this higher. Even at 0.5Mbit/s the padding burst will be ~3kb and as such we'll probably use a lot of fairly large packets for spurious resends. 700-900 seems like a better first guess than 200

src/streams/send.rs
algesten: Yeah. I was thinking that too. However. Maybe it's an all-or-nothing situation. In the sense that either we make perfect stats (like we did in the past), or we get away with not perfect stats. I do really like the simplification you did.

src/streams/send.rs
algesten: Ok. I change that.

src/streams/send.rs
k0nserv: Yea, this is still an improvement. If we just scale it with the number of packets maybe that'd be okay for now

k0nserv (approved):

PR #228, opened 2023-08-01
pub some things

xnorpx:
We are converting to and from str0m constructs and it would be useful and avoid code duplication if we could pub and add the following.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -135,7 +135,7 @@ pub struct FormatParams {
 /// Session config for all codecs.
 #[derive(Debug, Clone, Default)]
 pub struct CodecConfig {
-    configs: Vec<PayloadParams>,
+    params: Vec<PayloadParams>,
 }
 
 impl PayloadParams {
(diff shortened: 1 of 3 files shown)


src/format.rs
```diff
@@ -181,6 +181,26 @@ impl PayloadParams {
...
+    /// Whether the payload uses the PLI (Picture Loss Indication) mechanic.
+    pub fn fb_pli(&self) -> bool {
+        self.fb_pli
+    }
+
+    /// Whether the payload uses the FIR (Full Intra Request) mechanic.
+    pub fn fb_fir(&self) -> bool {
+        self.fb_fir
+    }
+
```
algesten, line 203: How is this useful without also being able to set these values?

Not saying we shouldn't do it, just want to understand.

src/rtp/ext.rs
```diff
@@ -287,7 +287,8 @@ impl ExtensionMap {
             .map(|p| p as u8 + 1)
     }
 
-    pub(crate) fn iter(&self, audio: bool) -> impl Iterator<Item = (u8, Extension)> + '_ {
+    /// Returns an iterator over the elements of the extension map.
+    pub fn iter(&self, audio: bool) -> impl Iterator<Item = (u8, Extension)> + '_ {
```
algesten, line 291: This needs documentation of what `audio` does.

src/ice/candidate.rs
```diff
@@ -297,10 +298,16 @@ impl Candidate {
         self.component_id
     }
 
-    pub(crate) fn addr(&self) -> SocketAddr {
+    /// Returns the address for the specified ICE candidate.
+    pub fn addr(&self) -> SocketAddr {
         self.addr
     }
 
+    /// Returns a reference to the protocol for the specified ICE candidate.
```
algesten, line 306: Document example values.

src/ice/candidate.rs
```diff
@@ -242,7 +242,8 @@ impl Candidate {
         hash.to_string()
     }
 
-    pub(crate) fn prio(&self) -> u32 {
+    /// Returns the priority for the specified ICE candidate.
```
algesten, line 245: Document this prio is calculated according to spec.

xnorpx:
will fix

src/format.rs
xnorpx: Currently we just generate default codecs and then read out the values and convert to our internal format. But I agree for more flexibility it would be useful to set them so one can customize the codecs fully.

Let me add that.

xnorpx: @algesten probably fine to have a set api for this and not expading the constructor, but let me know what you think.

algesten: Either way it creates a bit of a problem, because we don't expect the sdp API to set any of this (since it is negotiated).

I think ultimately we want a `PayloadParamBuilder` that sits under `rtp` or some such (where I expose lower level stuff). This is to make clear on a code level, that we don't expect people to normally use this.

However for this first iteration, I think setters are enough (I haven't made my mind up on how to deal with this exactly).

src/format.rs
```diff
@@ -285,6 +332,11 @@ impl CodecConfig {
         c
     }
 
+    /// Returns a reference to the vector of payload configurations.
+    pub fn get_configs(&self) -> &Vec<PayloadParams> {
```
algesten, line 336: Let's rename the internal field and this function `params()`. That's the name in the rest of the API.

algesten (commented):
One niggly thing

src/format.rs
xnorpx: will fix

algesten (approved):
LGTM!

PR #234, opened 2023-08-01
Fix incorrect logic in unpacking RTX

algesten:


src/streams/receive.rs
```diff
@@ -214,23 +219,35 @@ impl StreamRx {
...
         clock_rate: u32,
+        is_repair: bool,
     ) -> (SeqNo, MediaTime) {
         self.last_used = now;
 
-        let previous_seq = self.register.as_ref().map(|r| r.max_seq());
+        // Select reference to register to use depending on RTX or not. The RTX has a separate
+        // sequence number series to the main register.
+        let register = if is_repair {
+            &mut self.register_rtx
```
k0nserv, line 229: You might be able to use `get_or_insert_with` here to change the type of `register` from `Option<&mut ReceiverRegister>` to just `&mut ReceiveRegister`.

Something like:

```diff
diff --git a/src/streams/receive.rs b/src/streams/receive.rs
index 129d029..d57be9e 100644
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -226,24 +226,20 @@ impl StreamRx {
         // Select reference to register to use depending on RTX or not. The RTX has a separate
         // sequence number series to the main register.
         let register = if is_repair {
-            &mut self.register_rtx
+            self.register_rtx
+                .get_or_insert_with(|| ReceiverRegister::new(header.sequence_number(None)))
         } else {
-            &mut self.register
+            self.register
+                .get_or_insert_with(|| ReceiverRegister::new(header.sequence_number(None)))
         };
 
-        let previous_seq = register.as_ref().map(|r| r.max_seq());
-        let seq_no = header.sequence_number(previous_seq);
+        let previous_seq = register.max_seq();
+        let seq_no = header.sequence_number(Some(previous_seq));
 
         let previous_time = self.last_time.map(|t| t.numer() as u64);
         let time_u32 = extend_u32(previous_time, header.timestamp);
         let time = MediaTime::new(time_u32 as i64, clock_rate as i64);
 
-        if register.is_none() {
-            *register = Some(ReceiverRegister::new(seq_no));
-        }
-
-        let register = register.as_mut().unwrap();
-
         register.update_seq(seq_no);
         register.update_time(now, header.timestamp, clock_rate);
 ```

k0nserv (approved):
LGTM

src/streams/receive.rs
algesten: Ah yes. Because on first instantiation we don't need the lengthened seq no anyway.

PR #239, opened 2023-08-02
Set StreamTx::rtx_pt on poll_packet

### LINKED ISSUE #236: RTP packet is never resend when received NACK
https://github.com/algesten/str0m/blob/2627fcbfaaa78a925b2a87b695e3a236ef7555b8/src/streams/send.rs#L422-L428

self.rtx_pt is never set then the above logic never passes cause the RTP packet doesn't resend when received NACK

algesten:
Close #236

### DIFF
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -293,6 +293,7 @@ impl StreamTx {
         };
 
         let is_audio = param.spec().codec.is_audio();
+        let mut set_rtx_pt = None;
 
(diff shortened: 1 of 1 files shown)


k0nserv (approved):
Seems reasonable. My only concern is that this is something we should do once, but we end up having this code run for every packet we send. Is there somewhere else that this could live, such that it only happens the one time?

PR #243, opened 2023-08-02
Account for resends in queue state

k0nserv:


src/streams/send.rs
```diff
@@ -696,6 +703,28 @@ impl StreamTx {
 
         let mut snapshot = self.send_queue.snapshot(now);
 
+        // Outstanding resends
+        let mut resend_snapshot =
+            self.resends
+                .iter()
+                .fold(QueueSnapshot::default(), |mut snapshot, r| {
```
k0nserv, line 710: Doing this every time doesn't feel great for performance but neither does exposing `SendQueue::total` so we can call `increase` and `decrease` as the resends change.

We could also move `TotalQueue` out of `SendQueue` to `StreamTx`

src/streams/send.rs
```diff
@@ -696,6 +703,28 @@ impl StreamTx {
...
+                        .first_unsent
+                        .map(|i| i.min(r.queued_at))
+                        .or(Some(r.queued_at));
+
+                    snapshot
+                });
+        resend_snapshot.created_at = now;
+        if resend_snapshot.packet_count > 0 {
+            resend_snapshot.priority = QueuePriority::Media;
+        }
```
algesten, line 724: Can you move this calculation to its own function? Just for tidyness.

algesten (approved):
LGTM

src/streams/send.rs
k0nserv: Yep I was weighing that already

PR #253, opened 2023-08-03
Dev/ci clippy all targets

xnorpx:


xnorpx:
this should fail now

xnorpx:
yep failed, reverting my revert

PR #255, opened 2023-08-03
Add remove media to direct api

xnorpx:


src/change/direct.rs
```diff
@@ -151,6 +151,15 @@ impl<'a> DirectApi<'a> {
...
     }
 
+    /// Remove `Media`.
+    ///
+    /// Removes media and all streams belong to a media identified by a `mid`.
+    pub fn remove_media(&mut self, mid: Mid) {
+        self.rtc.session.medias.retain(|media| media.mid() != mid);
+        self.rtc.session.streams.remove_streams_by_mid(mid);
+        self.rtc.session.source_keys.retain(|_, v| v.0 != mid);
+    }
```
algesten, line 161: Could we maybe move the remove_media call to `self.rtc.session` so that we don't need to `pub(crate)` the source_keys?

tests/rtp-direct-mid-rid.rs
```diff
@@ -127,5 +127,8 @@ pub fn rtp_direct_mid_rid() -> Result<(), RtcError> {
 
     assert_eq!(media.len(), 3);
 
+    l.direct_api().remove_media(mid);
+    r.direct_api().remove_media(mid);
+
```
algesten, line 132: Maybe assert that it's gone? Using the `Rtc::media` call.

algesten (commented):
Two niggly things.

src/session.rs
```diff
@@ -61,7 +61,7 @@ pub(crate) struct Session {
 
     /// Each incoming SSRC is mapped to a Mid/Ssrc. The Ssrc in the value is for the case
     /// where the incoming SSRC is for an RTX and we want the "main".
-    source_keys: HashMap<Ssrc, (Mid, Ssrc)>,
+    pub source_keys: HashMap<Ssrc, (Mid, Ssrc)>,
```
algesten, line 64: Or maybe just not pub? :)

algesten (commented):
One tiny thing

algesten (approved):

Issue #183, opened 2023-06-14
MIDs are not 3 characters


pthatcher:
MIDs can be any length (well... there's probably some limit in the standard... but it's a lot more than 3).  Lengths of 1 or 2 are probably the most common (libwebrtc just increments "0", "1", and it's good to stay small for the header extension value).  So maybe it will never matter in practice, but it doesn't seem to make sense to me to make the MID type a [u8; 3].  It's probably best to use String or Vec<u8>, or if perf around it really did matter somehow, Arc<str> or smallvec or CompactString or something fancy.

algesten:
The reason is capped at 3 is because I want it to be a Copy type. We can increase the length and null-pad. But I want a Copy type, not Arc or something non-std.

k0nserv:
SmolStr isn't copy, but when the data is less than 23 bytes I believe it's Clone implementation is effectively copy anyway

pthatcher:
Why do you want it to be Copy?

algesten:
Well, because they are copied a lot :)

They are used as identifier of media in many situations. Like. On every event out that relates to specific media. I don't want Clone there.

algesten:
This might be superseded elsewhere. But I found this: https://www.rfc-editor.org/rfc/rfc5888#page-4

```
           mid-attribute      = "a=mid:" identification-tag
           identification-tag = token
                                ; token is defined in [RFC 4566](https://www.rfc-editor.org/rfc/rfc4566)
```

And token is defined as: https://www.rfc-editor.org/rfc/rfc4566#page-39

```
   token-char =          %x21 / %x23-27 / %x2A-2B / %x2D-2E / %x30-39
                         / %x41-5A / %x5E-7E

   token =               1*(token-char)
```

So in that respects it seems 3 chars is already "too big", but then there might be standards that comes afterwards or de-facto usage (I think it's Firefox that does 3).

k0nserv:
Relevant WebRTC discussion https://groups.google.com/g/discuss-webrtc/c/VG406JMTBI4/m/MrSex_q7AgAJ

k0nserv:
We could copy libWebRTC and use `[u8; 16]` as the Mid type and reject longer mids when parsing SDPs

xnorpx:
I have seen examples where with mid = "audio", "video". Using what libWebRTC does is probably a good compromise?

xnorpx:
Increased to 16 so closing this issue.

PR #256, opened 2023-08-03
Increase size of Mid

xnorpx:
Based on discussion in #183 increase to 16 similar to libwebrtc.

We use numbers as mid and we did hit 100 in some scenarios so we need increase.


algesten (approved):
Alright. Let's take the plunge.

Issue #260, opened 2023-08-07
Getter function for the send_addr field of the Rtc struct


biceneutron:
First of all, this is a pretty cool project, thanks for all the efforts!

I'm currently using this in our project, and I need to access the `send_addr` field of the `Rtc` struct. Is it possible to add a simple getter like 
```rust
pub fn send_addr(&self) -> Option<(SocketAddr, SocketAddr)> {
    if let Some(send_addr) = &self.send_addr {
        return Some((send_addr.source, send_addr.destination));
    }
    None
}
```
to the `Rtc` struct?

algesten:
Hello! Thanks!

I suppose `current_send_addr()` or something like that.

I guess grabbing it from the outgoing packets means there's risk you get a probe packet for some ICE candidate?

biceneutron:
Yeah, it would be nicer if we can get it directly from `Rtc`.

algesten:
Closing due to inactivity

PR #262, opened 2023-08-07
Add a getter for the send_addr field of the Rtc struct

biceneutron:
#260


algesten (commented):
Hm. This is not just the send address, but both the source and send address. 

Not saying we can't expose this, but i'd like to know more about the use case.

algesten:
Closing due to inactivity

PR #264, opened 2023-08-07
Reset buffers on media direction change  

k0nserv:


src/session.rs
```diff
@@ -860,6 +861,28 @@ impl Session {
 
         Ok(())
     }
+
+    pub(crate) fn set_direction(&mut self, mid: Mid, direction: Direction) -> bool {
+        let Some(media) = self.media_by_mid_mut(mid) else {
+            return false;
```
xnorpx, line 867: Will this ever happen? If it happens does it matter if it returns true or false?

algesten: I guess if someone uses the SDP change to set a direction for a non-existing MID.

But could also be relevant if we expose this in the Direct API.

It could be an error.

algesten: The documentation currently says:

```text
    /// If the direction is set for media that doesn't exist, or if the direction is
    /// the same that's already set [`SdpApi::apply()`] not require a negotiation.
```

xnorpx (approved):
lgtm

PR #269, opened 2023-08-09
Merge stats for a given (mid, rid) pair

k0nserv:
A given (mid, rid) pair can use multiple SSRCs. Not only is there an RTX
SSRC and a main SSRC, which was already handled prior to this commit,
but during the lifetime of an RTP Session new SSRCs (RTX and main) can be
created for the same (mid, rid) pair. In particular, Firefox does this
when a transceiver transitions from send to inactive, and back to send.
When media is resumed the original SSRCs are not used, instead new SSRCs
are created.

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -150,6 +150,47 @@ pub struct MediaIngressStats {
     // pub remote: RemoteEgressStats,
 }
 
+impl MediaIngressStats {
+    /// Merge `other` into `self`, mutating `self`.
+    ///
+    /// **Panics** if called with stats that don't have the same `(mid, rid)` pair.
+    pub(crate) fn merge(&mut self, other: &Self) {
+        *self = self.merged(other);
+    }
+
+    /// Merge `self` and `other` returning a new `Self` of the result.
+    ///
+    /// **Panics** if called with stats that don't have the same `(mid, rid)` pair.
+    fn merged(&self, other: &Self) -> Self {
+        assert!(
+            self.mid == other.mid,
+            "Cannot merge MediaIngressStats for different mids"
+        );
+        assert!(
+            self.rid == other.rid,
+            "Cannot merge MediaIngressStats for different rids"
+        );
+        let (rtt, loss) = if self.timestamp > other.timestamp {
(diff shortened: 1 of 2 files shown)


src/stats.rs
```diff
@@ -150,6 +150,47 @@ pub struct MediaIngressStats {
...
+            bytes: self.bytes + other.bytes,
+            packets: self.packets + other.packets,
+            firs: self.firs + other.firs,
+            plis: self.plis + other.plis,
+            nacks: self.nacks + other.nacks,
+            rtt,
+            loss,
+            timestamp: self.timestamp.max(other.timestamp),
+        }
+    }
```
algesten, line 191: Why is this split into two functions? Seems like we either modify-in-place or we don't. But here we modify in place by using a function that doesn't. Doesn't it also create an extra allocation that we just throw away?

src/stats.rs
```diff
@@ -150,6 +150,47 @@ pub struct MediaIngressStats {
     // pub remote: RemoteEgressStats,
 }
 
+impl MediaIngressStats {
+    /// Merge `other` into `self`, mutating `self`.
+    ///
+    /// **Panics** if called with stats that don't have the same `(mid, rid)` pair.
+    pub(crate) fn merge(&mut self, other: &Self) {
```
algesten, line 157: Maybe something like `merge_by_mid`?

algesten (changes requested):
Some questions

src/stats.rs
k0nserv: `merged_by_mid_rid` then surely?

algesten: Sure

src/stats.rs
algesten: Another way to not forget fields is to deconstruct `other`

```rust
let Self  { _all the fields_ } = other;
```

k0nserv: Yeah I suppose, it occupies some local names that I'd want though.

I got rid of `merged` in the latest commit but kept the `*self = Self { ..}` bit

algesten (approved):

PR #271, opened 2023-08-09
Pub VideoOrientation

xnorpx:
Need access to the VideoOrientation enum to convert to and from with the rtp api.


xnorpx:
ok let me check that

xnorpx:
Seems to be some kind of collision between rtp::Extension::VideoOrientation and rtp::VideoOrientation

xnorpx:
nevermind I get it now.

algesten (approved):

PR #272, opened 2023-08-09
Fix RTX seq num ending up in main receive register

k0nserv:
When we received blank padding packets on the RTX stream those would end up getting
registered in the receive register for the main stream. This would in
turn cause us to NACK packets that never existed in the main stream.

The relevant code looked like this:

```rust
let receipt = if is_repair {
    // Blank padding packets will have empty data.
    if !data.is_empty() { // 1
        // Rewrite the header, and removes the resent seq_no from the body.
        stream.un_rtx(&mut header, &mut data, pt);
    }

    // Now update the "main" register with the repaired packet info.
    // This gives us the extended sequence number of the main stream.
    //
    // We need to do this also for blank padding to schedule the "paused event".
    let receipt = stream.update(now, &header, clock_rate, false); // 2

    // Drop blank padding
    if data.is_empty() {
        return;
    }

    receipt
} else {
  // omitted
}
```

A blank padding packet with seq no **X** would skip the `!data.is_empty()`(1) branch and thus not have the sequence
number in the header rewritten. Then we would call
`StreamRx::update` with the unmodified sequencer number **X**, that of the RTX
stream, and it would be registered as received in the `ReceiveRegister` for the main stream.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -485,24 +485,19 @@ impl Session {
         // the seq_no, however MediaTime might be different when interpreted against the
         // the "main" register.
         let receipt = if is_repair {
-            // Blank padding packets will have empty data.
-            if !data.is_empty() {
-                // Rewrite the header, and removes the resent seq_no from the body.
-                stream.un_rtx(&mut header, &mut data, pt);
-            }
-
-            // Now update the "main" register with the repaired packet info.
-            // This gives us the extended sequence number of the main stream.
-            //
-            // We need to do this also for blank padding to schedule the "paused event".
-            let receipt = stream.update(now, &header, clock_rate, false);
-
-            // Drop blank padding
+            // Drop RTX packets that are just empty padding. The payload here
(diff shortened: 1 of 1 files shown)


k0nserv:
Looking at this a different solve strikes me:

```patch
diff --git a/src/session.rs b/src/session.rs
index 30a2fde..6774229 100644
--- a/src/session.rs
+++ b/src/session.rs
@@ -495,7 +495,7 @@ impl Session {
             // This gives us the extended sequence number of the main stream.
             //
             // We need to do this also for blank padding to schedule the "paused event".
-            let receipt = stream.update(now, &header, clock_rate, false);
+            let receipt = stream.update(now, &header, clock_rate, data.is_empty());
 
             // Drop blank padding
             if data.is_empty() {
```

However, this means we'll register the same `SeqNo` twice for each blank RTX packet, but I guess that's fine 🤔

PR #273, opened 2023-08-09
Less mid/Media-level

algesten:


tests/sdp-negotiation.rs
```diff
@@ -0,0 +1,187 @@
...
+        //
+        &[opus(100, 2)],
+        &[opus(102, 2)],
+        MediaKind::Audio,
+        Direction::SendOnly,
+    );
+
+    // For sendonly only the ANSWER from R is mandatory. The OFFER contains suggestions.
+    // L -> R must be 102
+    // R -> L can be anything, but str0m tries to match the remote, so also 102
```
k0nserv, line 64: Is this true? Doesn't the pt for L remain unbound until an offer is made with a `recvonly` or `sendrecv` direction? The value 100 here is just a suggestion for what it wants to send i.e. `l.codec_config_tx` should be 102 and `r.codec_config_rx` should be `102`.

tests/sdp-negotiation.rs
```diff
@@ -0,0 +1,184 @@
...
+        "L sends to R on PT 102"
+    );
+    assert_eq!(
+        &[opus(100, 2)],
+        &**l.codec_config_rx(),
+        "L receives from R on PT 100"
+    );
+    assert_eq!(
+        &[opus(102, 2)],
+        &**r.codec_config_rx(),
```
k0nserv, line 40: Should say `&**r.codec_config_tx()` and the value above should be 100 i.e. R should respect the value that was present in the offer(100)

tests/sdp-negotiation.rs
k0nserv: I guess 🤔

tests/sdp-negotiation.rs
lolgesten: Yeah. Sigh. Maybe I should remove the text. Too many things that can go wrong when rigging the tests.

k0nserv: Could write a helper to reduce the duplication to

algesten: Done

src/media/writer.rs
```diff
@@ -120,7 +123,7 @@ impl<'a> Writer<'a> {
     /// a=rtcp-fb:96 nack pli
     /// ```
     pub fn is_request_keyframe_possible(&self, kind: KeyframeRequestKind) -> bool {
-        media_by_mid(&self.session.medias, self.mid).is_request_keyframe_possible(kind)
```
k0nserv, line 123: Should we reject sdps that violate these feedbacks being tied to the pt?

```
m=video ...
a=rtpmap:96 VP8/90000
a=rtcp-fb:96 nack
m=video ...
a=rtpmap:96 VP8/90000
// No NACK
```

https://datatracker.ietf.org/doc/html/draft-ietf-mmusic-sdp-mux-attributes-17#section-4.7 says it needs to be identical

src/rtp/ext.rs
```diff
@@ -356,6 +342,70 @@ impl ExtensionMap {
...
+            .map(|(i, _)| i) else {
+                return;
+            };
+
+        if new_index == old_index {
+            return;
+        }
+
+        // swap them
+        let old = self.0[old_index].take();
```
k0nserv, line 405: Can use `swap` here

src/rtp/ext.rs
```diff
@@ -356,6 +342,70 @@ impl ExtensionMap {
 
         orig_len - b.len()
     }
+
+    pub(crate) fn narrow(&mut self, remote_exts: &[(u8, Extension)], is_audio: bool) {
+        // Match remote numbers, this does not remove any configured extension.
```
k0nserv, line 347: This seems too accepting, we should verify that the remote hasn't changed its mind or violated the invariants(same ext id mapping to the same header extension).

src/rtp/ext.rs
```diff
@@ -729,45 +820,82 @@ mod test {
...
+
+        e1.narrow(&e2.iter(false).collect::<Vec<_>>(), false);
+
+        assert_eq!(
+            e1.iter(false).collect::<Vec<_>>(),
+            vec![(12, VideoOrientation), (14, TransportSequenceNumber)]
+        );
+        assert_eq!(
+            e2.iter(false).collect::<Vec<_>>(),
+            vec![(12, VideoOrientation), (14, TransportSequenceNumber)]
```
k0nserv, line 875: This change wouldn't be okay if the remote had already agreed to have `3` mean `TransportSequenceNumber` I think.

It's definitely not okay for one id to map to different extensions and I think it's not okay to map several ids to the same extension

src/rtp/id.rs
```diff
@@ -11,13 +11,17 @@ use rand::random;
...
 macro_rules! str_id {
-    ($id:ident, $name:literal, $num:tt) => {
+    ($id:ident, $name:literal, $num:tt, $new_len:tt) => {
         #[derive(Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
         pub struct $id([u8; $num]);
 
         impl $id {
             pub fn new() -> $id {
-                $id(Id::random().into_array())
+                let mut arr = Id::<$num>::random().into_array();
```
k0nserv, line 20: This seems weird? Why are we removing a bunch of the random data here?

src/session.rs
```diff
@@ -896,6 +899,17 @@ impl Session {
 
         true
     }
+
+    pub fn is_request_keyframe_possible(&self, kind: KeyframeRequestKind) -> bool {
+        // TODO: It's possible to have different set of feedback enabled for different
+        // payload types. I.e. we could have FIR enabled for H264, but not for VP8.
+        // We might want to make this check more fine grained by testing which PT is
+        // in "active use" right now.
```
k0nserv, line 907: Agreed with this `TODO` it should be based on the media so the signature here would change to 

```rust
pub fn is_request_keyframe_possible(&self, kind: KeyframeRequestKind, pt: Pt) -> bool
```

k0nserv (approved):
Looks good overall.

The extension id handling seems much simpler than what I [implemented in webrtc-rs](https://github.com/webrtc-rs/webrtc/pull/154/files), it might just be that you have figured out a better method. In there and in our own media servers I kept proposed and negotiated extension ids separated. Here we for sure don't detect and reject illegal changes by the remote side here, which maybe we should.

src/change/sdp.rs
```diff
@@ -63,6 +64,8 @@ impl<'a> SdpApi<'a> {
     /// let json_answer = serde_json::to_vec(&answer).unwrap();
     /// ```
     pub fn accept_offer(self, offer: SdpOffer) -> Result<SdpAnswer, RtcError> {
+        info!("Accept offer");
```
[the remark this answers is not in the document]
algesten: Will do

src/format.rs
```diff
@@ -663,3 +805,17 @@ impl fmt::Display for Codec {
         }
     }
 }
+
+impl std::ops::Deref for CodecConfig {
```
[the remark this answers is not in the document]
algesten: I think of it like this.

The codec configurations is essentially a list of payload (parameters). We could have had just a `Vec<PayloadParameter>`, but there are a couple of nice functions (`match` etc) that are part of the public API.

If I put it like that, does that make it any better?

src/media/mod.rs
```diff
@@ -27,47 +27,65 @@ pub use crate::rtp_::{Direction, ExtensionValues, MediaTime, Mid, Pt, Rid};
...
-    /// This is for _outgoing_ SDP. Incoming CNAME can be
-    /// found in the `ssrc_info_rx`.
+    /// RTP level.
     cname: String,
 
+    /// Rid that we are expecting to see on incoming RTP packets that map to this mid.
+    /// Once discovered, we make an entry in `stream_rx`.
+    ///
+    /// RTP level.
+    rids: Rids,
```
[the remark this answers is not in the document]
algesten: Alright

src/format.rs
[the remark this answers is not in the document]
k0nserv: Somewhat, but I think it can still be confusing. `Deref` is a bit too magic, I'd prefer explicit methods instead

src/media/mod.rs
```diff
@@ -98,44 +129,75 @@ pub(crate) struct ToPayload {
...
-        self.kind
+    /// Add rid as one we are expecting to receive for this mid.
+    ///
+    /// This is used for situations where we don't know the SSRC upfront, such as not having
+    /// a=ssrc lines in an SDP. Adding a rid means we are dynamically discovering the SSRC from
+    /// a mid/rid combination in the RTP header extensions.
+    ///
+    /// RTP level.
+    pub fn expect_rid(&mut self, rid: Rid) {
+        if matches!(self.rids, Rids::Any) {
```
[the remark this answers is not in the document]
algesten: ok

src/media/mod.rs
```diff
@@ -98,44 +129,75 @@ pub(crate) struct ToPayload {
...
+            Rids::Any => true,
+            Rids::Specific(v) => v.contains(&rid),
+        }
     }
 
-    /// The negotiated payload parameters for this media.
-    pub fn payload_params(&self) -> &[PayloadParams] {
-        &self.params
+    pub(crate) fn expects_specific_rids(&self) -> bool {
+        matches!(self.rids, Rids::Specific(_))
```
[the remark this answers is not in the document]
algesten: Good idea

src/media/mod.rs
```diff
@@ -221,7 +270,7 @@ impl Media {
 
         if !exists {
             // This unwrap is ok because we needed the clock_rate before unpayloading.
-            let params = self.get_params(pt).unwrap();
+            let params = params.iter().find(|p| p.pt == pt).unwrap();
```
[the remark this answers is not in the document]
algesten: Yeah. I make that clearer

src/media/mod.rs
```diff
@@ -388,15 +383,26 @@ impl Media {
...
-            KeyframeRequestKind::Fir => r.fb_fir,
-        })
+    pub(crate) fn set_remote_pts(&mut self, pts: Vec<Pt>) {
+        // Have we already set PTs?
+        if !self.remote_pts.is_empty() {
+            return;
+        }
+
+        // TODO: We should verify the remote peer doesn't suddenly change the PT
+        // order or removes/adds PTs that weren't there from the start.
```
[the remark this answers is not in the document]
algesten: That's effectively what will happen.

* First negotiation. Video OFFER/ANSWER for Mid(2) will narrow Session::codec_config to only those indicated.
* Second negotiation. Video OFFER/ANSWER for Mid(3), with expanded codecs, we will answer with the narrow from first negotiation.

This does make me think we have a bug though if we do it the reverse. If the first negotiation has a broader codec support and then a narrower, we will narrow away codecs that are potentially in use.

src/media/writer.rs
algesten: There's certainly more validation that should be done of the remote SDP. I think we start a TODO list in change/sdp.rs, cause I know there is another one I can't quite remember right now.

src/rtp/ext.rs
algesten: Cool!

src/rtp/ext.rs
algesten: Yeah. None of that would be ok. We're currently a bit better in PT negotiation with the `locked` property. Maybe we should do that here too?

src/rtp/id.rs
algesten: We want 3-letter Mid (like FF). Xnorpx/Peter wanted the possibility to have longer Mid. When we changed to the longer ones, we also made our random Mids very long. I want to bring the default random ones back to 3.

src/session.rs
algesten: Yeah, but that makes for a public API change I don't want. This is what I mean with "active use". We need to somewhere record which PT is in current use. Hence I left it as TODO for now.

src/rtp/ext.rs
algesten: Ah yes. This is where I mean we should add more SDP validation earlier. To detect such shenanigans.

src/rtp/ext.rs
algesten: Yeah. None of that would be ok. We're currently a bit better in PT negotiation with the locked property. Maybe we should do that here too?

k0nserv: Yes, that's why I had the proposed/negotiated distinction in webrtc-rs I think

README.md
```diff
@@ -363,6 +364,12 @@ let rtc = Rtc::builder()
     .build();
 ```
 
+RTP mode gives us some new API points.
+
+1. [`Event::RtpPacket`] emitted for every incoming RTP packet.
```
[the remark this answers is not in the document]
algesten: Sure

src/session.rs
k0nserv: Hmm, does this even make sense `is_request_keyframe_possible` is a `StreamRx` question, does it even make sense to ask at the session level?

src/format.rs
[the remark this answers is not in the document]
algesten: How about I provide that as external API? Leave the deref. Doesn't make much difference if it's possible to Deref _as well as_?

src/rtp/ext.rs
algesten: Yeah, I don't think we need to go that far. It's enough to just lock it ones we got the remote side confirmed.

src/format.rs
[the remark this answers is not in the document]
k0nserv: I'm taking my cues from the Deref docs:

> Implementing Deref for smart pointers makes accessing the data behind them convenient, which is why they implement Deref. On the other hand, the rules regarding Deref and [DerefMut](dfile:///Users/hugotunius/Library/Application%20Support/Dash/DocSets/Rust/Rust.docset/Contents/Resources/Documents/doc.rust-lang.org/1.71.1/std/ops/trait.DerefMut.html) were designed specifically to accommodate smart pointers. Because of this, **Deref should only be implemented for smart pointers** to avoid confusion.

Our use of Deref for newtypes I think is okayish, but it feels like a stretch here

src/session.rs
algesten: It's for `media/Writer`. The high level API.

k0nserv: But then wouldn't you ask as a specific media/Writer about it?

algesten: Not really necessary. The remote side is sending on _some PT_ and the high level API doesn't care what it is. It just want a keyframe. Rid is the only thing we need to provide.

k0nserv: I don't think that makes much sense, but that hasn't changed in this PR so no point in debating that here

PR #275, opened 2023-08-10
assume! macro instead of unwrap

algesten:


src/macros.rs
```diff
@@ -0,0 +1,13 @@
+/// Shorthand for unwrap_or_else to provide context for panics.
+///
+/// The macro should be used similar to `assert!` with an intermediary `maybe_value` or `result_value`.
+///
+/// ```ignore
+/// let maybe_value = my_values.iter().find(|x| == 42);
+/// let value = assume!(maybe_value, "The value should be defined because: {:?}", something);
+/// ```
```
k0nserv, line 8: Worth expanding on why we want to use `unwrap_or_else` and the limitation on only being usable with Option/Result

algesten: True that

PR #279, opened 2023-08-15
Fix RTX

k0nserv:
Packets with incoming RTX payload types were not correctly handled and
went ignored, resulting in no retransmissions.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -451,14 +451,18 @@ impl Session {
 
         // If the header ssrc differs from the main, it's a repair stream.
         let is_repair = header.ssrc != ssrc;
-
-        let clock_rate = match get_params(&self.codec_config, header.payload_type) {
-            Some(v) => v.spec().clock_rate,
+        let params = match main_payload_params(&self.codec_config, header.payload_type) {
+            Some(p) => p,
             None => {
-                trace!("No codec params for {:?}", header.payload_type);
+                trace!(
+                    "No codec params could be found(wether main or RTX) for {:?}",
+                    header.payload_type
+                );
                 return;
             }
         };
+        let clock_rate = params.spec().clock_rate;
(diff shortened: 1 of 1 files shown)


src/session.rs
```diff
@@ -965,7 +965,11 @@ pub struct PacketReceipt {
 }
 
 fn get_params(c: &CodecConfig, pt: Pt) -> Option<&PayloadParams> {
-    c.iter().find(|p| p.pt() == pt)
+    if let p @ Some(_) = c.iter().find(|p| p.pt() == pt) {
+        return p;
+    }
+
+    main_payload_type_for(c, pt).and_then(|pt| c.iter().find(|p| p.pt == pt))
```
algesten, line 972: I think you should remove the `get_params` instead and call `main_payload_type_for` directly from the (only) callsite.

algesten (changes requested):

src/session.rs
k0nserv: See c7506b6, I changed it to both remap and find the PayloadParams in one go. No more double traversals over the codecs

src/session.rs
```diff
@@ -451,14 +451,18 @@ impl Session {
...
         let is_repair = header.ssrc != ssrc;
-
-        let clock_rate = match get_params(&self.codec_config, header.payload_type) {
-            Some(v) => v.spec().clock_rate,
+        let params = match main_payload_params(&self.codec_config, header.payload_type) {
+            Some(p) => p,
             None => {
-                trace!("No codec params for {:?}", header.payload_type);
+                trace!(
+                    "No codec params could be found(wether main or RTX) for {:?}",
```
algesten, line 458:
```suggestion
                    "No payload params could be found (main or RTX) for {:?}",
```

PR #280, opened 2023-08-15
bwe: precalculate mean_max_rtt

algesten:
The calculation is in hot loops and only updates with TWCC.

Also avoid allocation of RTCP VecDeque.


k0nserv (approved):
Perfect

PR #281, opened 2023-08-15
Prefer str0m's locked pts over offer suggestions

k0nserv:
When receiving an offer with a `sendonly` m-line don't remap Pts if they
differ from str0m's locked values. The values from the offer are what
the offerer would like to send, but ultimately the answer dictates this.
If str0m has already establish a pt for the codecs in question reply
with those.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -798,9 +798,17 @@ fn add_new_lines(
         if m.typ.is_media() {
             let mut media = Media::from_remote_media_line(m, idx);
             media.need_open_event = need_open_event;
+            // If this was for a sendonly m-line in the offer the Pts are merely a suggestion, we SHOULD attempt
+            // to match these.
+            // The check here is inverted because `Media::direction` is the direction from str0m's
+            // perspective
+            let params_suggestive =
+                media.direction().is_receiving() && !media.direction().is_sending();
 
             // Match/narrow remote params.
-            session.codec_config.update_params(&m.rtp_params());
+            session
+                .codec_config
+                .update_params(&m.rtp_params(), params_suggestive);
 
             // Narrow the agreed on extmaps.
             session.exts.narrow(&m.extmaps(), media.kind().is_audio());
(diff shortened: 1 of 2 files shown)


src/change/sdp.rs
```diff
@@ -798,9 +798,17 @@ fn add_new_lines(
         if m.typ.is_media() {
             let mut media = Media::from_remote_media_line(m, idx);
             media.need_open_event = need_open_event;
+            // If this was for a sendonly m-line in the offer the Pts are merely a suggestion, we SHOULD attempt
+            // to match these.
+            // The check here is inverted because `Media::direction` is the direction from str0m's
+            // perspective
+            let params_suggestive =
```
k0nserv, line 805: This isn't quite right, there are two conditions we need to consider here:

1. Is this a recvonly m-line(from the perspective of str0m)
2. Was it added in an offer(in which case it express the desire of the offerer).

src/change/sdp.rs
```diff
@@ -879,12 +887,13 @@ fn update_media(media: &mut Media, m: &MediaLine, config: &mut CodecConfig, stre
...
         media.expect_rid(*rid);
     }
+    dbg!(m, &config);
 
     // Narrowing/ordering of of PT
     let pts: Vec<Pt> = m
         .rtp_params()
         .into_iter()
-        .filter(|p| config.matches(p))
+        .filter_map(|p| config.match_params(p))
```
k0nserv, line 896: This also isn't quite right, we should attempt to match the Pt used by the offerer if we haven't already locked down the codec in question.

If the offer specifies `120=VP8/90000` and we have a default of `96=VP8/90000` this will use our default in the case when we haven't previous locked down `VP8/90000`, we should respect the offerer's suggestion

k0nserv:
Closed in favour of #282

src/change/sdp.rs
algesten: Got ya.

PR #282, opened 2023-08-15
sdp: fix FF PT change

algesten:


k0nserv:
Following on my comments in #281, this code only runs when we add a new m-line, but we also need to handle this on direction changes I think? For example, what about an m-line that's initially offered as `inactive` and is later made `recvonly` or `sendonly`

src/change/sdp.rs
```diff
@@ -800,7 +800,9 @@ fn add_new_lines(
             media.need_open_event = need_open_event;
 
             // Match/narrow remote params.
-            session.codec_config.update_params(&m.rtp_params());
+            session
+                .codec_config
+                .update_params(&m.rtp_params(), m.direction());
```
k0nserv, line 805: The direction needs to be inverted here since `Media::direction` returns the local direction. Same in `update_media`

k0nserv (commented):
The logic around `sendonly` and the Pts being suggestions for what the remote wants to send only apply for offers. Do we have a bug here when we have offered a `recvonly` m-line and are processing the answer(in which case remote_dir will be `sendonly`)

src/change/sdp.rs
algesten: This is m (remote m-line). Not media.

src/format.rs
```diff
@@ -587,8 +613,15 @@ impl CodecConfig {
         }
 
         // Now lock potential new parameters to remote.
+        //
+        // If the remote is doing `SendOnly`, the PTs are suggestions, and we are allowed to
+        // ANSWER with our own allocations as overrides. If SendRecv or RecvOnly, the remote
+        // is talking about its own receiving capapbilities and we are not allowed to change it
+        // in the ANSWER.
+        let warn_on_locked = !remote_dir.is_receiving();
```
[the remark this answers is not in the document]
algesten: Inactive _with_ codecs? 🤔

src/change/sdp.rs
k0nserv: Ah, so it is

k0nserv (approved):

PR #287, opened 2023-08-17
Padding before first regular packet

algesten:
- Encapsulation of blank packet sentinel value
- Set blank packet PT before first regular packet

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -221,6 +221,7 @@ impl Session {
             sender_ssrc,
             do_nack,
             &self.medias,
+            &self.codec_config,
             &mut self.feedback_tx,
         );
 
--- a/src/streams/mod.rs
+++ b/src/streams/mod.rs
@@ -89,7 +89,8 @@ pub struct StreamPaused {
     pub paused: bool,
 }
 
-pub const BLANK_PACKET_DEFAULT_PT: Pt = Pt::new_with_value(0);
+/// 255 is out of range for a real PT, which is 7 bit.
+const BLANK_PACKET_DEFAULT_PT: Pt = Pt::new_with_value(255);
 
 impl RtpPacket {
     fn blank() -> RtpPacket {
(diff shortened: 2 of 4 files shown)


src/streams/mod.rs
```diff
@@ -224,15 +231,18 @@ impl Streams {
...
+            stream.handle_timeout(now, get_media);
 
             // This unwrap is OK because a StreamTx can't exist without the
             // corresponding Media (Mid).
-            let cname = medias
-                .iter()
-                .find(|m| m.mid() == stream.mid())
-                .unwrap()
-                .cname();
+            let cname = medias.iter().find(|m| m.mid() == mid).unwrap().cname();
```
algesten, line 245: Lol. Ok. So here we do get the media…. why all the closure bs.

src/session.rs
```diff
@@ -221,6 +221,7 @@ impl Session {
             sender_ssrc,
             do_nack,
             &self.medias,
+            &self.codec_config,
```
k0nserv, line 224: This seems wrong. We need to intersect the session codecs with those negotiated for this media, otherwise we might use an RTX PT that the remote peer has disabled.

algesten: Yeah, if you look at `Media::first_pt_with_rtx` it does that.

src/streams/mod.rs
algesten: That sorted in a newer commit.

src/session.rs
k0nserv: Did you just fix that? I could've sworn I checked `Media::first_pt_with_rtx` and the `self.remote_pts.contains(&p.pt)` bit wasn't there

algesten: The sneaky fox...

k0nserv (approved):
This all looks good.

The pacer will still prevent any padding from being sent on these streams though:

https://github.com/algesten/str0m/blob/a83c618838c2695fc326575d31bbad749f38c8b4/src/packet/pacer.rs#L561-L576

Look at the `last_send_time` and `max_by_key` bits

algesten:
@k0nserv do you have any suggestions on what to do with `last_send_time` and `max_by_key`?

algesten:
The last_send_time and max_by_key is not an issue.

PR #291, opened 2023-08-21
Handle unencrypted RTCP for AEAD_AES128_GCM

k0nserv:


algesten (approved):
Do we have a test for it?

k0nserv:
> Do we have a test for it?

I'm just reversing a test in the RFC atm

PR #295, opened 2023-08-22
Enable (and prefer) AeadAes128Gcm

algesten:


src/dtls/mod.rs
```diff
@@ -96,7 +96,8 @@ pub enum SrtpProfile {
 impl SrtpProfile {
     // All the profiles we support, ordered from most preferred to least.
     // TODO: Enable SrtpProfile::Aes128CmSha1_80
```
k0nserv, line 98: This todo needs to go

k0nserv (approved):

PR #300, opened 2023-08-23
fix quantized_size index of out bounds

davibe:
```
thread 'tokio-runtime-worker' panicked at 'range end index 48 out of range for slice of length 46', /home/ubuntu/work/lookback/str0m/src/streams/rtx_cache.rs:53:22
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
./bin/dev.sh: line 12:  2131 Aborted                 (core dumped) cargo run -- --watch --enable-simulcast
```

### DIFF
--- a/src/streams/rtx_cache.rs
+++ b/src/streams/rtx_cache.rs
@@ -2,13 +2,13 @@ use std::collections::BTreeMap;
 use std::time::Duration;
 use std::time::Instant;
 
-use crate::io::DATAGRAM_MTU;
+use crate::io::DATAGRAM_MAX_PACKET_SIZE;
 use crate::rtp_::SeqNo;
 
 use super::RtpPacket;
 
 const RTX_CACHE_SIZE_QUANTIZER: usize = 25;
-const RTX_CACHE_QUANTIZE_SLOTS: usize = DATAGRAM_MTU / RTX_CACHE_SIZE_QUANTIZER;
+const RTX_CACHE_QUANTIZE_SLOTS: usize = DATAGRAM_MAX_PACKET_SIZE / RTX_CACHE_SIZE_QUANTIZER;
 
 #[derive(Debug)]
 pub(crate) struct RtxCache {
--- a/src/io/mod.rs
+++ b/src/io/mod.rs
@@ -29,6 +29,9 @@ pub(crate) const DATAGRAM_MTU: usize = 1150;
 /// Warn if any packet we are about to send is above this size.
 pub(crate) const DATAGRAM_MTU_WARN: usize = 1280;
 
+/// Max UDP packet size
+pub(crate) const DATAGRAM_MAX_PACKET_SIZE: usize = 2000;
+
 /// Max expected RTP header over, with full extensions etc.
 pub const MAX_RTP_OVERHEAD: usize = 80;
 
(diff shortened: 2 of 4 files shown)


src/streams/rtx_cache.rs
```diff
@@ -48,7 +51,7 @@ impl RtxCache {
     }
 
     pub fn get_cached_packet_smaller_than(&mut self, max_size: usize) -> Option<&mut RtpPacket> {
-        let quantized_size = max_size / RTX_CACHE_SIZE_QUANTIZER;
+        let quantized_size = (max_size / RTX_CACHE_SIZE_QUANTIZER).min(RTX_CACHE_MAX_PACKET_SIZE);
```
algesten, line 54: This is comparing apples and pears. The `quantized_size` is an index, and `RTX_CACHE_MAX_PACKET_SIZE` is a byte size.

algesten (changes requested):

src/streams/rtx_cache.rs
```diff
@@ -48,7 +48,7 @@ impl RtxCache {
     }
 
     pub fn get_cached_packet_smaller_than(&mut self, max_size: usize) -> Option<&mut RtpPacket> {
-        let quantized_size = max_size / RTX_CACHE_SIZE_QUANTIZER;
+        let quantized_size = max_size.min(DATAGRAM_MAX_PACKET_SIZE) / RTX_CACHE_SIZE_QUANTIZER;
```
algesten, line 51: This should be `DATAGRAM_MTU_WARN - MAX_RTP_OVERHEAD`

algesten (changes requested):

src/streams/rtx_cache.rs
davibe: Are you sure ? It's like saying that the cache indexes up to 2000 but you're never ever allowed to request beyond 1200. How does it make sense ?

algesten: str0m itself should always stay below DATAGRAM_MTU_WARN. However I can't control what people do with the RTP API, so they might push in something larger.

I don't want to clamp that larger thing, because then when we ask for something smaller, we might end up with the big one.

So the cache can take larger sizes, but we are not going to produce them for spurious padding.

I also think the `min` clamp should move back to where it was, because this is not an internal concern to RtxCache.

davibe: So when they do push something larger, we do handle that bigger packet, but then we refuse to re use it for padding. Yet we do index it in write-only memory, and we rely on the caller to call the "rtx" cache api with a max_size that does not go out of bounds.

Alternatively, we could simply skip indexing the packet that exceeds the allowed padding size (instead of colliding them in the same slot) and keep the max rtx size a detail that is internal to the *rtx cache*

algesten: Alright. Move the max size to a parameter to `new`. Throw away packages that are larger in `cache_sent_packet` and clamp to that size in `get_cached_packet_smaller_than`. Then set `DATAGRAM_MTU_WARN - MAX_RTP_OVERHEAD` on the outside when calling `new`.

algesten (approved):

PR #301, opened 2023-08-23
Report SR/RR together per Mid

algesten:


k0nserv (approved):
Seems reasonable.

The only concern is that we might end up reporting more frequently than intended now. If we are ultimately unlucky, receiving $N$ simulcast tracks the streams are created at $1s / N$ intervals resulting in $N$ RRs per second instead of 1 per second as we intend

algesten:
Every create_sr/rr will reset the timer for that StreamRx/Tx, so the only result should be that they trigger at the same time.

PR #303, opened 2023-08-24
Docs for 0.2.0

algesten:
- Move BWE to bwe pub mod
- Doc fixes

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,7 +450,18 @@ behavior, at worst a security risk!
 
 Panics are our friends: *panic means bug*
 
-And also: str0m should *never* panic on any user input.
+And also: str0m should *never* panic on any user input. If you encounter a panic,
+please report it!
+
+#### Catching panics
+
+Panics should be incredibly rare, or we have a serious problem as a project. For an SFU,
+it might not be ideal if str0m encounters a bug and brings the entire server down with it.
+
+For those who want an extra level of safety, we recommend looking at [`catch_unwind`][catch]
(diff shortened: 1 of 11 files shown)


src/lib.rs
```diff
@@ -1745,6 +1742,9 @@ impl RtcConfig {
     /// // Defaults to 15.
     /// assert_eq!(config.reordering_size_audio(), 15);
     /// ```
+    ///
+    /// This setting is ignored in [RTP mode][`RtcConfig::set_rtp_mode()`] where RTP
+    /// packets are arrived out of order.
```
algesten, line 1747: are arrived?

README.md
```diff
@@ -439,11 +428,29 @@ references. I.e. `pc.getTranscievers()` returns objects that can be
...
+
+We do not believe it's correct to change every `unwrap()`/`expect()` into
+`unwrap_or_else()`, `if let Some(x) = x { ... }` etc, because doing so
+brushes an actual problem (an incorrect assumption) under the carpet. Trying
+to hobble along with an incorrect state would at best result in broken
+behavior, at worst a security risk!
+
+Panics are our friends: *panic means bug*
+
+And also: str0m should *never* panic on any user input.
```
k0nserv, line 453: Maybe mention catching the panic and converting it to log error here. If the user discards the whole `Rtc` instances all the state that caused the invariant violation should be gone. This gives folks some resilience to avoid bringing down whole server processes in the event of panics

algesten: ok

k0nserv (approved):

PR #307, opened 2023-08-24
Implement ICE restart fully

k0nserv:


src/change/sdp.rs
```diff
@@ -1151,18 +1195,43 @@ impl From<MediaKind> for MediaType {
 }
 
 struct AsSdpParams<'a, 'b> {
-    pub candidates: &'a [Candidate],
-    pub creds: &'a IceCreds,
+    pub candidates: Cow<'a, [Candidate]>,
```
k0nserv, line 1198: @algesten you are going to be sad, but this seemed like the best solution, it still maintains the reference in most cases

k0nserv: We could change our local candidates immediately on creating the offer I suppose

src/change/sdp.rs
```diff
@@ -73,7 +74,14 @@ impl<'a> SdpApi<'a> {
...
 
-        add_ice_details(self.rtc, &offer)?;
+        // When the remote offers, it indicates an ICE restart by changing its ICE credentials
+        let ice_restart = self
+            .rtc
+            .ice
+            .remote_credentials_changed(offer.ice_creds().as_ref());
+        // we always accept the ICE restart by providing new credentials ourselves
+        let maybe_new_ice_creds = ice_restart.then(IceCreds::new);
+        add_ice_details(self.rtc, &offer, ice_restart, maybe_new_ice_creds.as_ref())?;
```
algesten, line 84: We should keep `add_ice_details` as is, and move the logic for detecting ice_restart to inside `add_ice_details`.

src/change/sdp.rs
```diff
@@ -143,7 +151,18 @@ impl<'a> SdpApi<'a> {
...
+            _ => None,
+        });
+
+        // The remote side indicates acceptance by changing its ICE credentials
+        let ice_restart = self
+            .rtc
+            .ice
+            .remote_credentials_changed(answer.ice_creds().as_ref());
+
+        add_ice_details(self.rtc, &answer, ice_restart, new_ice_creds)?;
```
algesten, line 165: Applying new ice creds should move to `apply_offer()` and `apply_answer()`, which is where we effectuate changes of offers/answers.

src/change/sdp.rs
```diff
@@ -335,6 +354,14 @@ impl<'a> SdpApi<'a> {
 
         id
     }
+
+    /// Perform an ICE restart.
```
algesten, line 358: Document that multiple restarts will only do one.

src/change/sdp.rs
```diff
@@ -335,6 +354,14 @@ impl<'a> SdpApi<'a> {
...
+    pub fn ice_restart(&mut self) {
```
algesten, line 359: Return the new `IceCreds` here.

src/change/sdp.rs
```diff
@@ -1151,18 +1195,43 @@ impl From<MediaKind> for MediaType {
...
+                    _ => None,
+                })
+            })
+            .unwrap_or_else(|| {
+                (
+                    rtc.ice.local_credentials().clone(),
+                    Cow::Borrowed(rtc.ice.local_candidates()),
+                )
+            });
+
```
algesten, line 1234: Let's land #308 and do a simple for-loop here with the pending ice restart creds, if there is any.

src/ice/agent.rs
```diff
@@ -660,17 +660,23 @@ impl IceAgent {
...
+
+        // we keep using local candidates, updating the ufrag to match our new credentials.
+        self.local_candidates = self
+            .local_candidates
+            .iter()
+            .map(|c| c.replace_creds(local_creds.ufrag.clone()))
+            .collect();
+        self.local_credentials = local_creds;
         self.remote_credentials = None;
-        self.local_candidates.clear();
```
algesten, line 673: Keep local candidates as is.

src/ice/agent.rs
```diff
@@ -1454,6 +1460,15 @@ impl IceAgent {
...
     }
+
+    pub(crate) fn remote_credentials_changed(&self, new_creds: Option<&IceCreds>) -> bool {
+        // Before we have set remote credentials we don't consider anything a change.
+        let Some(current) = &self.remote_credentials else {
+            return false
+        };
+
+        Some(current) != new_creds
+    }
```
algesten, line 1471: Make a `remote_credentials(&self) -> Option<&IceCreds>` and make the comparison on the sdp.rs side.

algesten (changes requested):
Some things

src/change/sdp.rs
k0nserv: I was trying to do that initially but the logic is different for answers and offers, so I kept as much common as possible. We could pass down the remote ice credentials too making the signature:

```rust
fn add_ice_details(
    rtc: &mut Rtc,
    sdp: &Sdp,
    remote_ice_creds: Option<&IceCreds>,
    new_local_ice_creds: Option<&IceCreds>,
);
```

for a slight improvement tough

src/ice/agent.rs
k0nserv: I did that initially, but felt it was nicer to have this method since it does a bit more than 

```rust
ice.remote_credentials() != credentials
```

In particular there's this special case about credentials not being considered changed if they have never been set

src/change/sdp.rs
algesten: In what way is it different?

src/change/sdp.rs
k0nserv: My thinking was that there was no way to get a hold of the local credentials previously so it felt consistent to keep them hidden. If we return them here shouldn't there be a way to get them via `Rtc`?

src/ice/agent.rs
algesten: I think I prefer that logic not being inside here. Unless we unconditionally set it and detect changes in `set_remote_credentials`

src/change/sdp.rs
k0nserv: Addressed in e794138

src/ice/agent.rs
k0nserv: addressed in e794138

src/change/sdp.rs
```diff
@@ -479,13 +512,38 @@ fn create_offer(rtc: &mut Rtc, changes: &Changes) -> SdpOffer {
...
+    let ice_restart = match rtc.ice.remote_credentials() {
+        Some(v) => *v != creds,
+        None => false,
+    };
+
+    // If we initiated the ICE Restart, the credentials are in the
+    // pending SDP changes. Otherwise we need to create new ones now.
+    let (new_local_creds, keep_local_candidates) = pending
+        .and_then(|p| p.ice_restart())
+        .unwrap_or_else(|| (IceCreds::new(), true));
```
algesten, line 538: This only needs to happen if we are doing `ice_restart`, move to inside the if-statement?

src/change/sdp.rs
algesten: Let's use a bit less combinators here, it gets a bit too muddled what's going on.

We can use the new "SdpPendingOffer::ice_restart()` to make this

```rust
let (creds, candidates) = if let Some(creds) = p.and_then(|p| p.ice_restart()) {
...
} else { 
...
}
```

src/ice/candidate.rs
```diff
@@ -351,6 +351,13 @@ impl Candidate {
     pub(crate) fn clear_ufrag(&mut self) {
         self.ufrag = None;
     }
+
+    pub(crate) fn replace_ufrag(&self, ufrag: &str) -> Candidate {
+        let mut new = self.clone();
+        new.set_ufrag(&ufrag);
+
+        new
+    }
```
algesten, line 360: I think `clone()` + `set_ufrag()` would be clearer and not requiring any new function.

algesten (changes requested):

src/change/sdp.rs
k0nserv: Was just looking at this, we can't do that because `p` isn't `SdpPendingOffer` here, its' `&Changes` but I'll figure something out

src/change/sdp.rs
k0nserv: Good catch, that wasn't intentional

src/ice/candidate.rs
k0nserv: Sure

src/change/sdp.rs
algesten: Oh. Hm… Let's move that ice_creds to `Changes`

`SdpPendingOffer` is after all just a facade against the API user.

src/change/sdp.rs
```diff
@@ -1152,17 +1211,40 @@ impl From<MediaKind> for MediaType {
 
 struct AsSdpParams<'a, 'b> {
     pub candidates: Vec<Candidate>,
-    pub creds: &'a IceCreds,
+    pub creds: IceCreds,
     pub fingerprint: &'a Fingerprint,
```
algesten, line 1220: Maybe clone Fingerprint and drop `'a` altogether? Starting to become a bit dubious :)

src/change/sdp.rs
k0nserv: Did you decide if we should still move this?

algesten (approved):
LGTM

Small question mark, but not necessary to change for merge.

src/change/sdp.rs
algesten: Happy with this for now.

src/change/sdp.rs
k0nserv: Feels like a follow-up PR since it's unrelated to this change

algesten: Sure

PR #308, opened 2023-08-24
ice: blank out ufrag on Candidate once added

algesten:
- ice: blank out ufrag on Candidate once added
- Clippy fixes

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -270,6 +270,8 @@ impl IceAgent {
     }
 
     /// Local ice candidates.
+    ///
+    /// The candidates have their ufrag filled out to the local credentials.
     pub fn local_candidates(&self) -> &[Candidate] {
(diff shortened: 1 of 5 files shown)


src/ice/agent.rs
```diff
@@ -365,6 +367,9 @@ impl IceAgent {
             }
         }
 
+        // "Adopt" any incoming candidate by setting our current ufrag.
+        c.set_ufrag(&self.local_credentials.ufrag);
```
k0nserv, line 371: How was this getting set before?

algesten: Not sure it was other than parsed from the SDP.

k0nserv: From the remote? But it returns its won clients, not ours

k0nserv (approved):
Let's do it

PR #312, opened 2023-08-25
Allow up to 100 consecutive write()

algesten:


davibe (approved):
controversial :) let's see

PR #313, opened 2023-08-26
chore: spelling

xnorpx:


algesten (approved):
Nice!

Didn't realize it was this bad bad :)

PR #314, opened 2023-08-28
bug: Use ssrc_main when calling expect_stream_rx

xnorpx:
When using ssrc we registered rtx as it own mid instead of looking up the main ssrc and register rtx on the main stream.

This lead to problems when looking up stream by mid we could get the rtx ssrc instead of the main one and hence we did not request keyframes on the correct ssrc.


algesten (approved):
Nice catch!

PR #316, opened 2023-08-30
sender: re-evaluate rtx ratio every 50ms

davibe:
`sum_since()` keeps re-evaluating the sum in the last second. Since `now` always move ahead we cannot precalculate the sum easily. Here I am updating it at 20 Hz which should be sufficient. I am getting a hit/miss ratio of 10 logging with println! which causes overhead so i expect it has greater impact with profiling


algesten (approved):
Looks good. Can you move `rtx_ratio_downsampled` to after `poll_packet_resend`?

PR #317, opened 2023-08-30
feature/buffer for rtx cache

algesten:
- EvictingBuffer for better RtxCache performance


src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+        if self.buf.is_empty() || self.max_age.is_zero() {
+            return;
+        }
+
+        if timestamp < self.last_timeout {
+            // Value is already considered evicted.
+            return;
+        }
+
+        // First element.
```
algesten, line 77: Fix comment

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
+use std::mem;
+use std::time::{Duration, Instant};
+
+use crate::util::already_happened;
+
+/// Fixed size growing buffer that evicts the oldest entries based on time.
```
algesten, line 6: Weird to say "fixed size" and "growing"

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+        self.last_timeout = now;
+
+        self.evict(now);
+
+        // If we had any overwrites since the last call to evict_and_maybe_grow(), we
+        // need to grow the buffer.
+        if self.overwrites > 0 {
+            self.grow();
+        }
+        self.overwrites = 0;
```
algesten, line 157: Move into if

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+        for e in old_buffer {
+            if let Some(e) = e {
+                let index = self.index_for_position(e.position);
+                self.buf[index] = Some(e);
+            }
+        }
+    }
+
+    pub fn contains(&self, position: u64) -> bool {
+        if self.buf.is_empty() || self.max_age.is_zero() {
```
algesten, line 220: Make `self.is_inert()` helper instead of repeating this.

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+        self.buf[index] = Some(Entry {
+            position,
+            timestamp,
+            value,
+        });
+    }
+
+    /// Get the entry for the previously inserted position.
+    #[allow(unused)]
+    pub fn get(&self, position: u64) -> Option<&T> {
```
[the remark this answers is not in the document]
algesten: sure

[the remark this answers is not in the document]
algesten: actually. no because `&self` needs to be `&mut self`

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+
+        if position < next_evict {
+            // Do not cache values older than evict position.
+            return;
+        }
+
+        let index = self.index_for_position(position);
+
+        if let Some(entry) = &self.buf[index] {
+            // If the position is exactly the same, we for some reason are overwriting the
```
davibe, line 93: When we cache the first entry, `next_evict` is == `position`. Is there another case for this situation ? Because if there is, we risk overwriting `self.next_evict` with new data, which blocks eviction around line :172 I think

algesten: True. We should probably move next_evict one position up in this case.

algesten: I can make a test for this.

src/streams/rtx_cache_buf.rs
```diff
@@ -0,0 +1,383 @@
...
+        // First element.
+        let next_evict = if let Some(v) = self.next_evict {
+            v
+        } else {
+            self.next_evict = Some(position);
+            position
+        };
+
+        if position < next_evict {
+            // Do not cache values older than evict position.
```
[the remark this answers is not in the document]
davibe: or "before the evict position". Somehow saying "older" seems strange since we also have time here.

src/streams/rtx_cache_buf.rs
algesten: It is a problem. I fixed it.

davibe:
Worth nothing this implementation can "loose packets" before the buffer eventually grows.

i.e. this test fails

```
    #[test]
    fn push_and_get_dade() {
        let mut buf = EvictingBuffer::new(2, Duration::from_secs(10));
        let now = Instant::now();

        buf.push(3, now, 'A');
        buf.push(5, now, 'B');

        assert_eq!(buf.get(3), Some(&'A'));
        assert_eq!(buf.get(5), Some(&'B'));
    }
 ```
 
This is especially true if the user sets a small size counting on the fact that it auto grows. 

I think in case of video the initial size should be something that fits a keyframe (you want to start with a keyframe, and to avoid additional keyframe requests due to lack of retransmission)

**Comment by davibe** [comment]
I think the doc around `RtcConfig.set_send_buffer_*` could not mention this is a min size and the buffer auto-grows

**Comment by algesten** [comment]
I'm expecting the initial buffer size to be big enough to hold the NACK amount required. I.e. if we want 2 seconds of NACK, a good initial buffer size should be able to hold the initial keyframe + more packets up to 2 seconds.

The buffer does grow, but I don't think it's good to rely on that only instead of making a good initial guess.

**Review by davibe — APPROVED** [review]

**Comment by xnorpx** [comment]
Shouldn't the rtt or the ack time decide the rough size of the buffer. Then with a cap of 3 seconds I think they had it in libwebrtc?

File: src/streams/rtx_cache.rs [hunk 8]
```diff
@@ -1,23 +1,19 @@
...
 const RTX_CACHE_QUANTIZE_SLOTS: usize = DATAGRAM_MAX_PACKET_SIZE / RTX_CACHE_SIZE_QUANTIZER;
 
 #[derive(Debug)]
 pub(crate) struct RtxCache {
-    // Config
-    max_packet_count: usize,
-    max_packet_age: Duration,
-
     // Data, new additions here probably need to be cleared in [`clear`].
-    packet_by_seq_no: BTreeMap<SeqNo, RtpPacket>,
```
xnorpx, line 20: Think this is a pretty good candidate for feature flag, or option to pick one packet backend over the other. There might be tradeoffs that make people want to pay more cogs for smaller memory footprint which might be hard to replicate with the buf.

xnorpx: maybe even a trait and have 2 implementations with a feature flag around

algesten: I don't know why people would want the BTreeMap?

PR #320, opened 2023-08-31
Out-of-band initial ROC

algesten:


src/streams/send.rs
```diff
@@ -254,7 +261,15 @@ impl StreamTx {
         nackable: bool,
         payload: Vec<u8>,
     ) -> Result<(), RtcError> {
-        //
+        if self.start_roc.is_none() {
+            // This is the first ever write_rtp. Remember the starting ROC.
+            self.start_roc = Some(*seq_no >> 16);
+        }
+
+        // SeqNo must start from 0.
```
algesten, line 269: Wrong comment

src/streams/receive.rs
```diff
@@ -597,6 +615,23 @@ impl StreamRx {
         self.rtx = Some(rtx);
         self.register_rtx = None;
     }
+
+    /// Reset the current rollover counter (ROC).
+    ///
+    /// This is used in scenarios where we use a single sequence number across all
+    /// receivers of the same stream (as opposed to a sequence number unique per peer).
+    ///
+    /// > Receivers joining an on-going session MUST be given the
```
algesten, line 624: Link to RFC

src/rtp/id.rs
```diff
@@ -189,6 +202,23 @@ impl SeqNo {
...
+    /// // More than 2^16, thus rolled over.
+    /// let seq_no: SeqNo = 95_000.into();
+    ///
+    /// assert_eq!(seq_no.roc(), 1);
+    ///
+    /// // Is the same as shifting 16 bits.
+    /// assert_eq!(seq_no.roc(), 95_000 >> 16);
+    /// ```
+    #[inline(always)]
+    pub fn roc(&self) -> u64 {
```
k0nserv, line 219: Nice, was just thinking that this method would be useful. Can you adopt in `src/rtp/srtp.rs`? too?

src/rtp/id.rs
```diff
@@ -161,6 +161,19 @@ num_id!(SessionId, u64);
...
 /// str0m keeps track of the roll overs and converts the `u16` to `u64` in this `SeqNo`.
+///
+/// To get the RTP u16 value from a `SeqNo`, deref and cast it.
+///
+/// ```
+/// # use str0m::rtp::SeqNo;
+/// let seq_no: SeqNo = 65_537.into();
+///
+/// // Discard upper 48 bits to get RTP u16.
+/// let rtp_seq_no = *seq_no as u16;
```
k0nserv, line 172: For symmetry should we have a method that does this too?

src/streams/receive.rs
```diff
@@ -597,6 +615,23 @@ impl StreamRx {
...
+    /// > current ROC value using out-of-band signaling such as key-management
+    /// > signaling.  Furthermore, the receiver SHALL initialize s_l to the RTP
+    /// > sequence number (SEQ) of the first observed SRTP packet (unless the
+    /// > initial value is provided by out of band signaling such as key
+    /// > management).
+    pub fn reset_roc(&mut self, roc: u64) {
```
k0nserv, line 630: Don't we need to be able to set the RTX ROC too? If a session has been going for long enough the RTX ROC might also be non-zero when the receiver joins

src/rtp/id.rs
algesten: Give me a name. I tried to come up with one :)

src/streams/receive.rs
algesten: The RTX is peer-peer (not entire session), and it's not possible for the user to chose a value for it.

src/rtp/id.rs
algesten: Sure

src/rtp/id.rs
k0nserv: What, you don't like `seq_no.seq_no()`? 😂 

`as_u16`, `truncated_u16`, `truncated`, `without_roc`, `omitting_roc`

src/streams/receive.rs
k0nserv: Cool 👍🏼

k0nserv (approved):

src/rtp/id.rs
k0nserv: Ah, of course.

Issue #166, opened 2023-05-14
Using `str0m` in `rust-libp2p`


thomaseizinger:
Following the advice from [here](https://github.com/algesten/str0m/issues/164#issuecomment-1545855375), I am opening a separate issue!

Here is what we want to do: Have libp2p applications running in the browser connect to server nodes that don't have a valid TLS certificate. Being in the browser, we don't have control over the WebRTC stack but use the `RTCPeerConnection` API.

The entire protocol is described here: https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md#browser-to-public-server

We also don't have separate STUN servers. Instead, the client knows the server's address through an out-of-band mechanism. We have a self-descriptive address format that looks like this:

```
/ip4/172.28.0.1/udp/9999/webrtc/certhash/uEiD8c9GWyEurSjh9UkY-YWXaXBJZIHo179zx9IpgDFFKgw/p2p/12D3KooWDpJ7As7BWAwRMfu1VU2WCqNjvq387JEYKDBj4kx6nXTN
```

The component after the `/certhash` part (`uEiD8c9GWyEurSjh9UkY-YWXaXBJZIHo179zx9IpgDFFKgw`) is the server's certificate fingerprint. This gives us enough information to directly connect to the server.

We want to use `str0m` for the _server_ part of this setup. The node will always be publicly reachable, hence we are in ICE-lite mode. We only ever need data channels, no media.

Concrete questions I have:

- Is there a way how I can disable certificate fingerprint validation?
- The spec describes to perform SDP munging but that doesn't seem to be possible with the `sdp_api`. I _think_ I achieved the same thing with the `direct_api` and adding candidates myself. Is there something I should be aware of when doing it that way?

More general questions:

- Is this possible at all with `str0m`?
- Is this a usecase you want to support?

xnorpx:
@thomaseizinger direct api is probably the way to go here instead of fighting sdp.

regarding At this point B does not know the TLS certificate fingerprint of A. Thus B can not verify A's TLS certificate fingerprint during the DTLS handshake. Instead B needs to disable certificate fingerprint verification (see e.g. [Pion's disableCertificateFingerprintVerification option](https://github.com/pion/webrtc/blob/360b0f1745c7244850ed638f423cda716a81cedf/settingengine.go#L62)).

What is the corresponding setting in webrtc-rs?

thomaseizinger:
[...]

PR #324, opened 2023-09-01
Implement missing features needeed for WebRTC Direct

altonen:
This commit introduces few minor changes to `str0m`:
  * allow user to set local ICE credentials
  * allow user to disable certificate fingerprint verification
  * allow user to fetch remote's certificate fingerprint

These changes allow implementing WebRTC Direct [1] using `str0m`

[1] https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md

---
Related to #166 

cc @thomaseizinger @xnorpx

### DIFF
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -52,6 +52,11 @@ impl<'a> DirectApi<'a> {
         self.rtc.ice.local_credentials().clone()
     }
 
+    /// Sets the local ICE credentials.
+    pub fn set_local_ice_credentials(&mut self, local_ice_credentials: IceCreds) {
+        self.rtc.ice.set_local_credentials(local_ice_credentials);
+    }
+
     /// Sets the remote ICE credentials.
     pub fn set_remote_ice_credentials(&mut self, remote_ice_credentials: IceCreds) {
         self.rtc.ice.set_remote_credentials(remote_ice_credentials);
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -70,6 +70,11 @@ impl<'a> DirectApi<'a> {
         self.rtc.dtls.local_fingerprint().clone()
     }
 
+    /// Returns a reference to the remote DTLS fingerprint used by this peer connection.
+    pub fn remote_dtls_fingerprint(&self) -> Option<Fingerprint> {
+        self.rtc.dtls.remote_fingerprint().clone()
+    }
+
     /// Sets the remote DTLS fingerprint.
     pub fn set_remote_fingerprint(&mut self, dtls_fingerprint: Fingerprint) {
         self.rtc.remote_fingerprint = Some(dtls_fingerprint);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1598,6 +1598,12 @@ impl RtcConfig {
         &self.dtls_cert
     }
 
+    /// Set DTLS certification.
+    pub fn set_dtls_cert(mut self, dtls_cert: DtlsCert) -> Self {
+        self.dtls_cert = dtls_cert;
+        self
+    }
+
     /// Toggle ice lite. Ice lite is a mode for WebRTC servers with public IP address.
     /// An [`Rtc`] instance in ice lite mode will not make STUN binding requests, but only
     /// answer to requests from the remote peer.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -908,7 +908,8 @@ impl Rtc {
         Rtc {
             alive: true,
             ice,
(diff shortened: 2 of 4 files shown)


src/change/direct.rs
```diff
@@ -65,6 +70,11 @@ impl<'a> DirectApi<'a> {
         self.rtc.dtls.local_fingerprint().clone()
     }
 
+    /// Returns a reference to the local DTLS fingerprint used by this peer connection.
```
algesten, line 73: Fix this doc

src/lib.rs
```diff
@@ -1610,6 +1615,20 @@ impl RtcConfig {
         self
     }
 
+    /// Set DTLS certification.
+    pub fn set_dtls_certification(mut self, dtls_cert: DtlsCert) -> Self {
+        self.dtls_cert = dtls_cert;
+        self
+    }
```
algesten, line 1622: Move this to after `dtls_cert` and rename it to `set_dtls_cert`.

src/lib.rs
```diff
@@ -1610,6 +1615,20 @@ impl RtcConfig {
...
+
+    /// Toggle certificate fingerprint verification.
+    ///
+    /// By default the certificate fingerprint is verified.
+    pub fn set_certificate_fingerprint_verification(mut self, enabled: bool) -> Self {
+        self.certificate_fingerprint_verification = enabled;
+        self
+    }
```
algesten, line 1630: 1. For consistent naming, rename this to `set_fingerprint_verification` (in all places).
2. Make a getter `fingerprint_verification` that has a doc test which checks the default is `true`.

algesten (changes requested):
Hi @altonen!

Thanks for this PR. There are a couple of small things to fix. However on an overall level, can you split this into 3 commits, one for each thing changing?

Also, we'd like to understand your use case. Disabling certificate verification is sometimes frowned upon in other contexts. What is the reason you need this? Is it actually a verification callback you are looking for?

altonen:
Hi @algesten, I'll make the necessary changes.

W.r.t. to disabling certificate verification, quoting from the spec:

> **No need for trusted TLS certificates.** Enable browsers to connect to public server nodes without those server nodes providing a TLS certificate within the browser's trustchain. Note that we can not do this today with our Websocket transport as the browser requires the remote to have a trusted TLS certificate. Nor can we establish a plain TCP or QUIC connection from within a browser. We can establish a WebTransport connection from the browser (see [WebTransport specification](https://github.com/libp2p/specs/blob/master/webtransport)).

and
> A and B execute the DTLS handshake as part of the standard WebRTC connection establishment.
At this point B does not know the TLS certificate fingerprint of A. Thus B can not verify A's TLS certificate fingerprint during the DTLS handshake. Instead B needs to disable certificate fingerprint verification (see e.g. [Pion's disableCertificateFingerprintVerification option](https://github.com/pion/webrtc/blob/360b0f1745c7244850ed638f423cda716a81cedf/settingengine.go#L62)).

In case you're interested, here's the full process for establishing browser<->server connection: https://github.com/libp2p/specs/blob/master/webrtc/webrtc-direct.md#browser-to-public-server

This is needed for Polkadot so that light clients can connect to full nodes over WebRTC

src/change/direct.rs
```diff
@@ -57,6 +57,11 @@ impl<'a> DirectApi<'a> {
         self.rtc.ice.set_remote_credentials(remote_ice_credentials);
     }
 
+    /// Sets the local ICE credentials.
+    pub fn set_local_ice_credentials(&mut self, local_ice_credentials: IceCreds) {
+        self.rtc.ice.set_local_credentials(local_ice_credentials);
+    }
+
```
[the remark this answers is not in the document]
altonen: Done

src/change/direct.rs
altonen: Done

src/lib.rs
altonen: Done

src/lib.rs
altonen: Done

algesten (approved):
LGTM!

algesten:
Thanks!

altonen:
Cheers, any chance a new release could be cut some time in the near future?

algesten:
Absolutely!

I haven't made progress on it in a couple of days, but I think we're ready: https://str0m.zulipchat.com/#narrow/stream/377845-general/topic/Release.200.2E2.2E0/near/387221067

algesten:
0.2.0 is released: https://crates.io/crates/str0m/0.2.0

altonen:
thanks!

algesten:
Looking forward to hear how it works out for you!

PR #327, opened 2023-09-04
feature/test loss 2

davibe:
- tests: add util to simulate packet loss
- add test to check retransmissions

### DIFF
--- a/tests/nack.rs
+++ b/tests/nack.rs
@@ -10,6 +10,164 @@ use str0m::{Event, RtcError};
 mod common;
 use common::{connect_l_r, init_log, progress};
 
+use crate::common::progress_with_loss;
+
+#[test]
+pub fn loss_recovery() -> Result<(), RtcError> {
+    init_log();
+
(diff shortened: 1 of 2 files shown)


tests/nack.rs
```diff
@@ -10,6 +10,163 @@ use str0m::{Event, RtcError};
...
+                time,
+                wallclock,
+                false,
+                exts,
+                true,
+                to_write.to_vec(),
+            )
+            .expect("clean write");
+
+        if !(10..=990).contains(&index) {
```
algesten, line 74: In the `src/lib.rs` I turned off this utterly insane lint. I think what you had before was more readable. Though I don't care hugely in tests.

davibe: agree

k0nserv (approved):
I think it might be useful to use a pseudo random generator(`rand` supports this) with a random seed for each test case. This way if we start getting flakiness in the tests we can easily reproduce the failing cases

PR #329, opened 2023-09-06
depack: signal discontinuity

davibe:


src/packet/buffer_rx.rs
```diff
@@ -171,15 +171,26 @@ impl DepacketizingBuffer {
 
         let is_more_than_hold_back = self.segments.len() >= self.hold_back;
 
+        let wait_for_contiguity = !contiguous && !is_more_than_hold_back;
```
algesten, line 174: Why is this necessary?

algesten (approved):

PR #330, opened 2023-09-06
Test contiguous

algesten:


tests/common.rs
```diff
@@ -1,19 +1,25 @@
...
 use str0m::format::PayloadParams;
 use str0m::net::Receive;
+use str0m::rtp::ExtensionMap;
+use str0m::rtp::RtpHeader;
 use str0m::Candidate;
 use str0m::{Event, Input, Output, Rtc, RtcError};
 use tracing::info_span;
 use tracing::Span;
 
+const PCAP_VP8: &[u8] = include_bytes!("data/vp8.pcap");
```
k0nserv, line 21: Including 69kb of data in the test binary seems ill-advised. It's useful to include binary data like this when you want a single binary, for tests we don't care and could just read the file from disk in the test instead, in particular this might slow down test compilation which would be a problem

tests/contiguous.rs
```diff
@@ -0,0 +1,88 @@
...
+    let mut count = 0;
+
+    // Contiguous all the way through.
+    for data in iter {
+        assert!(data.contiguous);
+        count += 1;
+    }
+
+    // TODO: We have 3 continuation flags. This should be 104-3 == 101, no?
+    assert_eq!(count, 102);
```
davibe, line 85: indeed

tests/common.rs
davibe: true if we include many for more tests

algesten: I don't believe it makes any difference whatsoever, since the compilation unit (common.rs) will be compiled once. But also I don't care either way.

tests/contiguous.rs
```diff
@@ -0,0 +1,203 @@
...
+        }
+    });
+
+    let mut count = 0;
+
+    // Contiguous all the way through.
+    for data in iter {
+        count += 1;
+        // We dropped packet 14337, which means 14338 should not be contiguous.
+        let assume_contiguous = !data.seq_range.contains(&14338.into());
```
davibe, line 195: This is wrong, because we rewrite seq numbers its 14337 that is going to be discontinuous

algesten: Where do we rewrite it?

davibe: we depack and repack with new sequence numbers before getting to MediaData

davibe: ah right, hm

PR #333, opened 2023-09-06
Surface SR timing information for ingress media

k0nserv:


src/streams/mod.rs
```diff
@@ -66,6 +66,12 @@ pub struct RtpPacket {
     /// For incoming packets it's the time we received the network packet.
     pub timestamp: Instant,
 
+    /// Timing information from the most recent Sender Report(SR). The first part of the tuple is
+    /// the NTP time, the second part is the RPT timestamp that corresponds to the same timestamp.
+    ///
+    /// If no Sender Report(SR) has been received or this packet is being sent by str0m this is [`None`].
+    pub last_sender_timing: Option<(MediaTime, u32)>,
```
davibe, line 73: I think MediaTime is better suited for RTP time and not for NTP time

algesten: I think it's a doc error.

src/packet/buffer_rx.rs
```diff
@@ -38,6 +43,14 @@ impl Depacketized {
             .expect("a depacketized to consist of at least one packet")
     }
 
+    pub fn first_sender_timings(&self) -> Option<(MediaTime, u32)> {
+        self.meta
+            .iter()
+            .min_by_key(|m| m.received)
+            .map(|m| m.last_sender_timing)
+            .expect("a depacketized to consist of at least one packet")
+    }
```
algesten, line 52: Don't we just want the latest value? Why is this `first` (i.e. `min`)?

src/streams/mod.rs
algesten: But having said that, I think we should surface it as a proper struct in `MediaData` instead of a tuple. I'm thinking we might zhuzh up `SenderInfo` and release that.

The one thing I want to change is to interpret that u32 to _something good_, like `Instant` or `MediaTime`.

davibe: https://github.com/algesten/str0m/pull/333/files#diff-5fac4aafc6e7f613e62fad8a848b9dfd4b5e625f210f7835428533b28015fda1R356

davibe:
`Instant` would not be good for a remote time, it can go backwards.

algesten:
Does it matter if it does?

davibe:
how do you convert the NTP to `Instant` ?

src/packet/buffer_rx.rs
davibe: it's picking the first last_sender_timing of the `first` packet that generated this sample.
There would be no difference if we picked the `last` (but its better than picking `any`).

In both cases it can be slightly wrong because it is possible that
- we receive the 1st packet of frame 3 and we enrich it with the last sender info
- we receive an updated sender info
- we receive the 1st packet of frame 2 and we enrich it with more recent sender info

=> In this case the sample-based api would flush out in-order samples with out-of-order sender infos

It's better to pick up the last value when MediaData is emitted or generated

src/streams/mod.rs
k0nserv: It's not a doc error, we do use a MediaTime for the NTP time in `SenderInfo`. I also thought it was a bit weird, should be converted to `SystemTime`imo.

Here's the parsin code https://github.com/algesten/str0m/blob/9feeb86cae7f59992799f5f00a5a77a42c7080d1/src/rtp/rtcp/sr.rs#L110-L113

src/packet/buffer_rx.rs
k0nserv: My thinking is like this, a frame is packetized into many packets that are typically all sent at the same time. They all have the same RTP timestamp. By picking the sender info based on the first packet that arrived we maximise the chance that we use values from an SR prior to the frame being sent. As Davide outlined it's still possible that this isn't the case, but it's minimised.

For exmaple:

```
    ┌──────┐
    │Frame │
    ├──────┘─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
       ┌──┐  ┌──┐ ┌──┐         ┌──┐      │
    │  │P2│  │P1│ │P3│         │P4│
       └──┘  └──┘ └──┘         └──┘      │
    └ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─
┌───┐           ┌───┐
│SR1│           │SR2│
└───┘           └───┘
```

Here we'd use the information from SR1 because that was the most recent SR when the first packet, P2, was received.

k0nserv:
> But having said that, I think we should surface it as a proper struct in MediaData instead of a tuple. I'm thinking we might zhuzh up SenderInfo and release that.

> The one thing I want to change is to interpret that u32 to something good, like Instant or MediaTime.

Sounds good to me. The `u32` should be a MediaTime I think. The NTP time ought to be SystemTime or Instant(I think converting NTP/unix time to Instants is a bit dubious though)

k0nserv:
Rebased on `main` and actioned

> But having said that, I think we should surface it as a proper struct in MediaData instead of a tuple. I'm thinking we might zhuzh up SenderInfo and release that.

algesten (approved):
LGTM

src/media/mod.rs
```diff
@@ -253,6 +253,7 @@ impl Media {
                         contiguous: dep.contiguous,
                         ext_vals: dep.ext_vals().clone(),
                         codec_extra: dep.codec_extra,
+                        last_sender_info: dep.first_sender_info(),
```
davibe, line 256: Here where you are composing MediaData can you get the last SR from the stream instead getting the potentially outdated one from dep ? (https://github.com/algesten/str0m/pull/333/files#r1318203310)

k0nserv: Yes we could, but I think we want to outdated one i.e the one that applies to the timestamps in this sample

k0nserv: As I explained in https://github.com/algesten/str0m/pull/333#discussion_r1318289998 it's intentionally done like this

PR #334, opened 2023-09-07
Direct api kind by mid

xnorpx:
Allow to lookup kind by mid to avoid cache outside.


xnorpx:
Closing can use: rtc.media(mid).unwrap().kind()

PR #335, opened 2023-09-07
Direct api add close for data channel

xnorpx:
Not sure if I missed something here. Please check that it's sane.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1271,6 +1271,7 @@ impl Rtc {
                         warn!("Drop ChannelClose event for id: {:?}", id);
                         continue;
                     };
+                    self.chan.remove_channel(id);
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -1268,9 +1268,9 @@ impl Rtc {
                 }
                 SctpEvent::Close { id } => {
                     let Some(id) = self.chan.channel_id_by_stream_id(id) else {
-                        warn!("Drop ChannelClose event for id: {:?}", id);
```
algesten, line 1271: Why does this warning go away?

algesten (approved):
LGTM!

One question

src/lib.rs
xnorpx: right you are correct that should stay let me fix.

PR #336, opened 2023-09-07
Time fixes

algesten:
- Expand PAST constant to also have SystemTime
- Consolidate time tricks into util/time_tricks
- Refactor to_unix_duration to use beginning_of_time
- Rename PAST constant to BEGINNING_OF_TIME
- Instant::from_ntp_u64 and as_ntp_u64

### DIFF
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -51,6 +51,9 @@ pub struct StreamRx {
     /// Timestamp when we got some indication of remote using this stream.
     last_used: Instant,
 
+    /// Last seen pt and clock_rate in
+    last_clock_rate: Option<(Pt, i64)>,
+
     /// Last received sender info.
     sender_info: Option<(Instant, SenderInfo)>,
 
@@ -132,6 +135,7 @@ impl StreamRx {
             cname: None,
             suppress_nack: false,
             last_used: already_happened(),
+            last_clock_rate: None,
             sender_info: None,
             reset_roc: None,
             register: None,
@@ -243,7 +247,21 @@ impl StreamRx {
         }
     }
 
-    fn set_sender_info(&mut self, now: Instant, info: SenderInfo) {
+    fn set_sender_info(&mut self, now: Instant, mut info: SenderInfo) {
+        // Extend the incoming time given our knowledge of last time.
+        let extended = {
+            let prev = self.sender_info.map(|(_, sr)| sr.rtp_time.numer() as u64);
+            let r_u32 = info.rtp_time.numer() as u32;
+            extend_u32(prev, r_u32)
+        };
+
+        // The MediaTime has a base 1 after being parsed. At this point
+        // we know whether it's audio or video and set the base accordingly.
+        let clock_rate = self.last_clock_rate.map(|(_, r)| r).unwrap_or(1);
+
+        // Clock rate is that of the last received packet.
+        info.rtp_time = MediaTime::new(extended as i64, clock_rate);
+
         self.sender_info = Some((now, info));
     }
 
@@ -346,6 +364,16 @@ impl StreamRx {
         time: MediaTime,
     ) -> Option<RtpPacket> {
         trace!("Handle RTP: {:?}", header);
+
+        let need_clock_rate = self.last_clock_rate.map(|(pt, _)| pt) != Some(header.payload_type);
+        if need_clock_rate {
+            self.last_clock_rate = Some((header.payload_type, time.denom()));
+
+            // If we get an SR before the first packet, we update the potential clock rate.
+            if let Some(info) = &mut self.sender_info {
+                info.1.rtp_time = MediaTime::new(info.1.rtp_time.numer(), time.denom());
+            }
+        }
 
         let packet = RtpPacket {
             seq_no,
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -2,12 +2,12 @@ use std::collections::VecDeque;
 use std::time::{Duration, Instant};
 
 use crate::media::KeyframeRequestKind;
-use crate::rtp_::{
-    extend_u32, DlrrItem, ExtendedReport, Fir, FirEntry, InstantExt, MediaTime, Mid, Pli, Pt,
-    ReceiverReport, ReportBlock, ReportList, Rid, Rrtr, Rtcp, RtcpFb, RtpHeader, SenderInfo, SeqNo,
-};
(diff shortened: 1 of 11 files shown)


src/util/mod.rs
```diff
@@ -14,10 +15,29 @@ pub(crate) fn not_happening() -> Instant {
...
+// driven from the external API using `Instant`. What works against us is that Instant can't
+// represent things like UNIX EPOCH (but SystemTime can).
+const HOURS_1: Duration = Duration::from_secs(60);
+static PAST: Lazy<(Instant, SystemTime)> = Lazy::new(|| {
+    // These two should be "frozen" the same instant. Hopefully they are not differing too much.
+    let now = Instant::now();
+    let now_sys = SystemTime::now();
+
+    let beginning_of_time = now.checked_sub(HOURS_1).unwrap();
+    // This might be less than 1 hour if the machine uptime is less.
```
k0nserv: I don't think this comment is  true. If the machine is one of the architectures that cannot represent `Instant`s before the machine's boot time the line above would've paniced

src/util/mod.rs
```diff
@@ -14,10 +15,29 @@ pub(crate) fn not_happening() -> Instant {
     *FUTURE
 }
 
+// The goal here is to make a constant "beginning of time" in both Instant and SystemTime
+// that we can use as relative values for the rest of str0m.
+// This is indeed a bit dodgy, but we want str0m's internal idea of time to be completely
+// driven from the external API using `Instant`. What works against us is that Instant can't
+// represent things like UNIX EPOCH (but SystemTime can).
+const HOURS_1: Duration = Duration::from_secs(60);
```
k0nserv: Why is this called `HOURS_1` if the value is 1 minute?

src/util/time_tricks.rs
```diff
@@ -30,6 +30,13 @@ static BEGINNING_OF_TIME: Lazy<(Instant, SystemTime)> = Lazy::new(|| {
     (beginning_of_time, beginning_of_time_sys)
 });
 
+fn beginning_to_epoch() -> Duration {
```
k0nserv, line 33: Since the beginning is after the epoch I'd expect a negative value here, isn't this `epoch_to_beginning` i.e. the duration from the **epoch** _to_ the **beginning**

src/util/time_tricks.rs
```diff
@@ -63,16 +91,52 @@ impl InstantExt for Instant {
...
+        // seconds relative to 0h UTC on 1 January 1900 [4]. The full
+        // resolution NTP timestamp is a 64-bit unsigned fixed-point number with
+        // the integer part in the first 32 bits and the fractional part in the
+        // last 32 bits.
+        let secs_ntp = (v as f64) / F32;
+
+        // Shift to UNIX EPOCH
+        let secs_epoch = secs_ntp - SECS_1900 as f64;
+
+        println!("from secs_epoch: {}", secs_epoch);
```
k0nserv, line 116: This needs to be removed

src/util/time_tricks.rs
```diff
@@ -63,16 +91,52 @@ impl InstantExt for Instant {
...
+        // Wallclock time (absolute date and time) is represented using the
+        // timestamp format of the Network Time Protocol (NTP), which is in
+        // seconds relative to 0h UTC on 1 January 1900 [4]. The full
+        // resolution NTP timestamp is a 64-bit unsigned fixed-point number with
+        // the integer part in the first 32 bits and the fractional part in the
+        // last 32 bits.
+        let secs_ntp = (v as f64) / F32;
+
+        // Shift to UNIX EPOCH
+        let secs_epoch = secs_ntp - SECS_1900 as f64;
```
k0nserv, line 114: This is the code path of user provided values and can panic because of overflow.For example, a sender can send an SR with the second part of the NTP timestamps set to `2_208_988_700` which will result in `2_208_988_700 - 2_208_988_800`, this will then panic on line 119 below

k0nserv (commented):
Some thoughts

src/util/mod.rs
algesten: Right. Should be saturating sub.

src/util/mod.rs
algesten: Very good point.

src/util/time_tricks.rs
algesten: Yep

src/util/time_tricks.rs
algesten: Yes

src/util/time_tricks.rs
algesten: Ok. Saturating it is.

src/util/time_tricks.rs
```diff
@@ -14,13 +14,26 @@ pub(crate) fn not_happening() -> Instant {
...
 // represent things like UNIX EPOCH (but SystemTime can).
-const HOURS_1: Duration = Duration::from_secs(60);
 static BEGINNING_OF_TIME: Lazy<(Instant, SystemTime)> = Lazy::new(|| {
     // These two should be "frozen" the same instant. Hopefully they are not differing too much.
     let now = Instant::now();
     let now_sys = SystemTime::now();
 
-    let beginning_of_time = now.checked_sub(HOURS_1).unwrap();
+    // Find an Instant in the past which is up to an hour back.
+    let beginning_of_time = {
```
k0nserv, line 23: Great! This is just what I was thinking we should do ⭐

k0nserv (approved):

src/streams/receive.rs
```diff
@@ -243,7 +247,20 @@ impl StreamRx {
...
+        // Extend the incoming time given our knowledge of last time.
+        let extended = {
+            let prev = self.sender_info.map(|(_, sr)| sr.rtp_time.numer() as u64);
+            let r_u32 = info.rtp_time.numer() as u32;
+            extend_u32(prev, r_u32)
+        };
+
+        // The MediaTime has a base 1 after being parsed. At this point
+        // we know whether it's audio or video and set the base accordingly.
+        let clock_base = if self.kind.is_audio() { 48_000 } else { 90_000 };
```
k0nserv, line 260: Is this future-proof? I have a vague recollection that all the Pts negotiated for a given mid need to have the same clock rate

k0nserv (approved):

src/streams/receive.rs
algesten: Well. The day we have some other clock base, we deal with it.

k0nserv: Sure, the only issue is we might not remember that we need to update it in this specific place

k0nserv (approved):
Looks good

PR #337, opened 2023-09-08
fix/nack register rollover

davibe:
- rx register: fix start index for wrap reset
- rx register: mark received after wrap reset
- add test for out of order packets on rollover

### DIFF
--- a/src/streams/register.rs
+++ b/src/streams/register.rs
@@ -108,7 +108,9 @@ impl ReceiverRegister {
         if did_wrap {
             // The indices wrapped around the end of `packet_status`, we clear any entries between
             // the current sequence number and nack_check_from.
-            let start = (*seq - self.packet_index(*seq) as u64).into();
+            let start = (*seq - self.packet_index(*seq) as u64)
+                .saturating_sub(self.packet_status.len() as u64)
+                .into();
             let end = self.nack_check_from;
             trace!(
                 "ReceiveRegister wrapped, clearing all entries from {start} to {end} on receiving {seq}"
--- a/src/streams/register.rs
+++ b/src/streams/register.rs
@@ -897,6 +897,31 @@ mod test {
     }
 
     #[test]
+    fn out_of_order_and_rollover() {
+        let mut reg = ReceiverRegister::new(0.into());
+
+        reg.update_seq(2998.into());
+        reg.update_seq(2999.into());
+
+        // receive older packet
+        reg.update_seq(2995.into());
+
+        // wrap
+        for i in 3000..5995 {
+            reg.update_seq(i.into());
+        }
+
+        // 5995 is missing
+
+        reg.update_seq(5996.into());
+        reg.update_seq(5997.into());
+
+        let reports = reg.create_nack_reports();
+        assert_eq!(reports.len(), 1);
+        assert_eq!(reports[0].reports[0].pid, 5995);
+    }
+
+    #[test]
     fn nack_check_on_seq_rollover() {
         let range = 65530..65541 + MISORDER_DELAY;
         let missing = [65535_u64, 65536_u64, 65537_u64];
--- a/src/streams/register.rs
+++ b/src/streams/register.rs
@@ -94,17 +94,6 @@ impl ReceiverRegister {
         let did_wrap = (*seq / self.packet_status.len() as u64)
             != (seq.saturating_sub(forward_delta) / self.packet_status.len() as u64);
 
-        let pos = self.packet_index(*seq);
-        let was_set = self.packet_status[pos].received;
-        self.packet_status[pos].received = true;
-
-        if self.packet_status[pos].received && self.packet_status[pos].nack_count > 0 {
-            debug!(
-                "Received packet {} after {} NACKs",
-                seq, self.packet_status[pos].nack_count
-            );
-        }
-
         if did_wrap {
             // The indices wrapped around the end of `packet_status`, we clear any entries between
             // the current sequence number and nack_check_from.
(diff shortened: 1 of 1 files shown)


k0nserv (approved):
Nice find!

src/streams/register.rs
```diff
@@ -94,21 +94,12 @@ impl ReceiverRegister {
...
-            );
-        }
-
         if did_wrap {
             // The indices wrapped around the end of `packet_status`, we clear any entries between
             // the current sequence number and nack_check_from.
-            let start = (*seq - self.packet_index(*seq) as u64).into();
+            let start = (*seq - self.packet_index(*seq) as u64)
+                .saturating_sub(self.packet_status.len() as u64)
+                .into();
```
algesten, line 102: There is a problem here, but this is not the fix. The bug appeared when we changed "probation", you can check that this works without the bug fix if you set MIN_SEQUENTIAL to 5.

```rust
        reg.update_seq(2998.into());
        reg.update_seq(2999.into());

        // receive older packet
        reg.update_seq(2995.into());
```

I believe the problem you're seeing is how the test starts. I don't think we ever move `nack_check_from` backwards, but that 2995 means we're marking packet as received that should been cleared when we roll over.

```diff
diff --git a/src/streams/register.rs b/src/streams/register.rs
index cc2d779..ddfa1ef 100644
--- a/src/streams/register.rs
+++ b/src/streams/register.rs
@@ -91,6 +91,11 @@ impl ReceiverRegister {
     ///
     /// Returns true if the packet received wasn't received before.
     fn record_received(&mut self, seq: SeqNo, forward_delta: u64) -> bool {
+        if seq < self.nack_check_from {
+            // just say no
+            return false;
+        }
+
         let did_wrap = (*seq / self.packet_status.len() as u64)
             != (seq.saturating_sub(forward_delta) / self.packet_status.len() as u64);
 
@@ -98,7 +103,7 @@ impl ReceiverRegister {
         let was_set = self.packet_status[pos].received;
         self.packet_status[pos].received = true;
 
-        if self.packet_status[pos].received && self.packet_status[pos].nack_count > 0 {
+        if self.packet_status[pos].nack_count > 0 {
             debug!(
                 "Received packet {} after {} NACKs",
                 seq, self.packet_status[pos].nack_count
@@ -108,7 +113,7 @@ impl ReceiverRegister {
         if did_wrap {
             // The indices wrapped around the end of `packet_status`, we clear any entries between
             // the current sequence number and nack_check_from.
-            let start = (*seq - self.packet_index(*seq) as u64).into();
+            let start = (*self.nack_check_from - self.packet_index(*seq) as u64).into();
             let end = self.nack_check_from;
             trace!(
                 "ReceiveRegister wrapped, clearing all entries from {start} to {end} on receiving {seq}"
@@ -894,6 +899,32 @@ mod test {
         assert_eq!(reports[0].reports[0].blp, 4);
     }
```

The clearing point in `did_wrap` is wrong, but this PR is not fixing it. I believe it should be "some offset" before `nack_check_from`.

davibe: I am not implying we ever move `nack_check_from` backwards.
In a real scenario it does happen that older packets arrive in a position before `nack_check_from", and it does happen that we mark them as received. You are right that they should be cleared on rollover.

The issue is that they are not cleared because the start seq number for `reset_received` is wrong.
When we rollover with seq (or max_seq), `nack_check_from` is still in the previous rollover window (it has not rolled over yet). Because `reset_received` works by sequence number, we need to move the start position back to the previous rollover window otherwise [*start..*end](https://github.com/algesten/str0m/blob/887aa5b7c185872c8a9a4d69af41cfecdf035924/src/streams/register.rs#L175) would be empty and no reset happens unless we have never rolled over.

The fix you are proposing is to "avoid to mark as received if it's before nack_check_from". It is ok but it does not fix the start index at rollover and it is not needed if we do fix it.

davibe: > The clearing point in did_wrap is wrong, but this PR is not fixing it. I believe it should be "some offset" before nack_check_from.

No, It must be from the beginning of the rolling window and it must happen at the first `seq` causing the rollover. This guarantees we clear the "future". If we leave out any `index` marked as received we will fail nacking it in the next rollover cycle which is preciseliy what happened.

davibe: If we introduce
```
        if seq < self.nack_check_from {
            // just say no
            return false;
        }
```
Then we can reset from self.nack_check_from - MAX_MISORDER. I did not do this because I thought there may be other edge cases leaving positions not cleared and there was no need to have more troubles here. If you prefer that, I can do.

I tried this approach here, I noticed it violates one of the prev test cases I had to change https://github.com/algesten/str0m/pull/343

algesten: Here's what I mean

`c` = check from
`s` = inserted seq

before roll over:
```
----------------c--------------s----------->
```

after roll over
```
-----s__________c-------------------------->
```

The area in underscore is the are we're trying to blank, since it might hold values from a previous roll-over.

That area is `next_check_from - pos(s)`. So I maintain the fix is:

* Don't mark entries before `next_check_from`.
* Correct the blanking range.

davibe: @algesten consider your example. 
```
d = dirty bites that have not been cleared

----------------c--------------s----------->

ROLLOVER

ddddds__________c-------------------------->
```
The issue from here onwards is that IF some dirty bits are marked as received, `nack_check_from` will skip them at the first opportunity (as soon as it wrongly detect some contiguity). Therefore we would not nack them.

So for example it may be that those 5 dirty bites before S are: missing, received, received, received, received.
When receiving the first missing, the subsequent ones will be skipped (nack_check_from will move past them)

davibe: In order to avoid those dirty bits, both strategies are valid
- clearing them on rollover (this patch)
- skip handling incoming seq numbers before `nack_check_from` (#343)

The latter is a tiny bit more risky in case there are other sources of dirty bits that we do not foresee, but should be fine as well.

algesten: Rather than arguing the finer points of this bug (I'm still not convinced), do you want to rewrite this thing?

I would either look at crates.io for a simple ring buffer, or write one myself.

Issue #339, opened 2023-09-08
Use `Bytes` instead of `Vec<u8>` for working with payload


evdokimovs:
We are integrating `str0m` to our media server, and we have architecture which requires cloning of `MediaData` structure. So we're wondering why you're not used [`bytes`][1] crate instead of `Vec<u8>` for working with payload. 

This crate will give us opportunity to cheaply clone `MediaData` (which stores `Vec<u8>` payload). Also it will be optimization for `str0m` because as I see you clone this payload too in [your code][2].

Also it will be more handy if we will add `Clone` implementation directly to `MediaData` structure, because `bytes` usage will make it cheaply cloneable, so there will be no problem with performance.

If there is no problem, I can implement it by myself.

[1]: https://crates.io/crates/bytes
[2]: https://github.com/algesten/str0m/blob/6f5694de9bbd479ee5e7280b347310a3f6160c44/src/media/writer.rs#L132

algesten:
Hi @evdokimovs, welcome to str0m!

I'm not a huge fan of `Bytes`, here's why:

1. It is "contagious", i.e. it's hard for a crate not wanting `Bytes` to interoperate with a crate that has `Bytes` as part of the public API without being forced to clone. The opposite is not true. An API designed around `&[u8]` and `Vec<u8>` can easily be used by someone preferring `Bytes`.
2. It has internal locks. str0m is (pretty much) lock free. The nature of `Bytes` means it does reference counting (similar to `Arc`). To `Drop`/deallocate a shared `Bytes`, the library internally requires thread synchronizations (it's done with atomics). This makes working with `Bytes` unpredictable – when a `Bytes` goes out of scope, also inside str0m, there may or may not be a lock wait.
3. `Vec<u8>` and `&[u8]` is the "Rust native" way. I like APIs to require as little knowledge as possible of the rest of the eco-system of crates.

You have correctly pointed out a place where we do a clone that maybe shouldn't be there. We think this should probably be `Into<Vec<u8>>` allowing the caller to pass either an uncloned `Vec<u8>` or something else. That would keep the caller in control of how this clone happens.

That change would be a welcome PR.

Also. I encourage finding other API points where you see unnecessary cloning, or where we can shift control of the cloning to the caller.

k0nserv:
[...]

PR #340, opened 2023-09-08
Use `Into<Vec<u8>>` instead of `&[u8]` in `Writer::write`

evdokimovs:
In #339 was decided to change `Writer::write` `data` parameter to be `Into<Vec<u8>>` instead of `&[u8]`, because it will allow the caller to pass either a `Vec<u8>` (without cloning) or something else. 

Also it will be more obvious that this method needs owned `data`.

### DIFF
--- a/src/media/writer.rs
+++ b/src/media/writer.rs
@@ -96,12 +96,12 @@ impl<'a> Writer<'a> {
     /// If you write media before `IceConnectionState` is `Connected` it will be dropped.
     ///
     /// Panics if [`RtcConfig::set_rtp_mode()`][crate::RtcConfig::set_rtp_mode] is `true`.
-    pub fn write<T: Into<Vec<u8>>>(
+    pub fn write(
         self,
         pt: Pt,
         wallclock: Instant,
         rtp_time: MediaTime,
-        data: T,
+        data: impl Into<Vec<u8>>,
     ) -> Result<(), RtcError> {
         let media = media_by_mid_mut(&mut self.session.medias, self.mid);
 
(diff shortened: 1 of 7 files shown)


src/media/writer.rs
```diff
@@ -96,12 +96,12 @@ impl<'a> Writer<'a> {
...
     ///
     /// Panics if [`RtcConfig::set_rtp_mode()`][crate::RtcConfig::set_rtp_mode] is `true`.
-    pub fn write(
+    pub fn write<T: Into<Vec<u8>>>(
         self,
         pt: Pt,
         wallclock: Instant,
         rtp_time: MediaTime,
-        data: &[u8],
+        data: T,
```
k0nserv, line 104: This can just be `impl Into<Vec<u8>>` which is slightly more ergonomic I think

algesten (commented):
Yep. Apart from that function signature, this looks great. Thanks!

src/media/writer.rs
evdokimovs: @k0nserv , @algesten ,

Fixed!

algesten (approved):
Excellent!

evdokimovs:
@algesten ,

Oops, sorry. I just couldn't run tests on my machine, because `cargo test` eats all available RAM (32GB) at build stage and hangs up the system. Maybe I missing something?

k0nserv:
> Oops, sorry. I just couldn't run tests on my machine, because cargo test eats all available RAM (32GB) at build stage and hangs up the system. Maybe I missing something?

I pushed a fix

k0nserv:
Thanks @evdokimovs

PR #342, opened 2023-09-10
Remove DtlsCert::new() on RtcConfig::new()

xnorpx:
Creating new dtlscert takes time and is expensive, in a server scenario one might want to reuse the certs.

This change allow the user to create it's own cert and then clone it for new instances instead of str0m creating a new one on every configure.

If a user have not created a cert then the build will generate one.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -917,8 +917,11 @@ impl Rtc {
         Rtc {
             alive: true,
             ice,
-            dtls: Dtls::new(config.dtls_cert, config.fingerprint_verification)
-                .expect("DTLS to init without problem"),
+            dtls: Dtls::new(
+                config.dtls_cert.unwrap_or_else(DtlsCert::new),
+                config.fingerprint_verification,
+            )
+            .expect("DTLS to init without problem"),
             session,
             sctp: RtcSctp::new(),
             chan: ChannelHandler::default(),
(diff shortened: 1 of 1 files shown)


xnorpx:
Numbers from this commit:

2023-09-09T23:52:22.560765Z  INFO rtc_config: Duration of RtcConfig and build: 92.2353ms
2023-09-09T23:52:22.644839Z  INFO rtc_config: Duration of new cert: 83.5913ms
2023-09-09T23:52:22.645563Z  INFO rtc_config: Duration of rtcconfig and build with passing in dtls cert: 343.5µs

src/lib.rs
```diff
@@ -1603,15 +1606,13 @@ impl RtcConfig {
     }
 
     /// The configured DtlsCert.
-    ///
-    /// The certificate is uniquely created per new RtcConfig.
-    pub fn dtls_cert(&self) -> &DtlsCert {
-        &self.dtls_cert
+    pub fn dtls_cert(&self) -> Option<&DtlsCert> {
+        self.dtls_cert.as_ref()
```
algesten, line 1610: I think it be worth to say that this is a _previously set_ dtls cert and that if you need the cert in use, go to direct api (and link to the call).

src/lib.rs
```diff
@@ -1603,15 +1606,13 @@ impl RtcConfig {
...
     }
 
     /// Set DTLS certification.
     pub fn set_dtls_cert(mut self, dtls_cert: DtlsCert) -> Self {
-        self.dtls_cert = dtls_cert;
+        self.dtls_cert = Some(dtls_cert);
```
algesten, line 1615: Some doc here be good too. That unless you set something, it will be automatically created for you.

algesten (approved):
LGTM. Some small doc changes.

PR #343, opened 2023-09-11
alt fix for nack register rollover

davibe:
- add test for out of order packets on rollover
- rx buffer: skip seqs before nack_check_from

### DIFF
--- a/src/streams/register.rs
+++ b/src/streams/register.rs
@@ -895,6 +895,31 @@ mod test {
     }
 
     #[test]
+    fn out_of_order_and_rollover() {
+        let mut reg = ReceiverRegister::new(0.into());
+
+        reg.update_seq(2998.into());
+        reg.update_seq(2999.into());
+
+        // receive older packet
+        reg.update_seq(2995.into());
+
+        // wrap
+        for i in 3000..5995 {
+            reg.update_seq(i.into());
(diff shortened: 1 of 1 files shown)


src/streams/register.rs
```diff
@@ -108,7 +112,11 @@ impl ReceiverRegister {
         if did_wrap {
             // The indices wrapped around the end of `packet_status`, we clear any entries between
```
davibe, line 113: now I am thinking maybe we don't need to clear the entries. Since we already clear them as we move `nack_check_from` ahead, and we don't allow updating positions before..

src/streams/register.rs
```diff
@@ -108,7 +112,11 @@ impl ReceiverRegister {
...
             // the current sequence number and nack_check_from.
-            let start = (*seq - self.packet_index(*seq) as u64).into();
+
+            let start = (*self.nack_check_from - MISORDER_DELAY)
+                .max(seq.saturating_sub(self.packet_status.len() as u64 + 1))
+                .into();
```
algesten, line 118: Can you talk me through this start value like I'm 5 years old.

davibe: I addressed my other comment removing this

algesten (approved):

PR #347, opened 2023-09-13
Remove `SimulcastGroup` and `SimulcastOption`

k0nserv:
`SimulcastOption` contained an `Ssrc` variant that wasn't used, when
removed we were only left with `Rid` so the whole thing goes.

`SimulcastGroup` incorrectly attempted to handle rid alternatives, the
support for this was not complete so it has been removed entirely.


algesten (approved):
Nice!

PR #348, opened 2023-09-13
rx register rewrite

davibe:
- separate nack and RR generation
- remove misorder delay
- remove probation period
- simplify handing nack to avoiding unnecessary operations

### DIFF
--- a/src/streams/register_nack.rs
+++ b/src/streams/register_nack.rs
@@ -2,17 +2,16 @@ use std::ops::Range;
 
 use crate::rtp_::{Nack, NackEntry, ReportList, SeqNo};
 
-/// How many sequence numbers we keep track of
-const MAX_DROPOUT: u64 = 3000;
-
 /// Number of out of order packets we keep track of for reports
-/// A suggested maximum value min is MAX_DROPOUT / 2
 const MAX_MISORDER: u64 = 100;
 
 const U16_MAX: u64 = u16::MAX as u64 + 1_u64;
 
 /// The max number of NACKs we perform for a single packet
 const MAX_NACKS: u8 = 5;
+
+/// Circular buffer size
+const BUFFER_SIZE: u64 = MAX_MISORDER + 1;
 
 #[derive(Debug)]
 pub struct NackRegister {
@@ -49,7 +48,7 @@ impl PacketStatus {
 impl NackRegister {
     pub fn new() -> Self {
         NackRegister {
-            packets: vec![PacketStatus::default(); MAX_DROPOUT as usize],
+            packets: vec![PacketStatus::default(); BUFFER_SIZE as usize],
             active: None,
         }
     }
@@ -67,15 +66,15 @@ impl NackRegister {
             return false;
         }
 
-        let mut next_active = active.start..active.end.max(seq);
+        let new = !self.packet(seq).received || seq > active.end;
 
-        let new = self.packet(seq).mark_received();
+        let end = active.end.max(seq);
 
-        next_active.start = {
-            let min = next_active.end.saturating_sub(MAX_MISORDER);
-            let mut start = (*next_active.start).max(min);
-            while start < *next_active.end {
-                if !self.packet(start.into()).received {
+        let start: SeqNo = {
+            let min = end.saturating_sub(MAX_MISORDER);
+            let mut start = (*active.start).max(min);
+            while start < *end {
+                if !self.packet(start.into()).received && start != *seq {
                     break;
                 }
                 start += 1;
@@ -84,7 +83,7 @@ impl NackRegister {
         };
 
         // reset packets that are rolling our of the nack window
-        for s in *active.start..*next_active.start {
+        for s in *active.start..*start {
             let p = self.packet(s.into());
             if !p.received {
                 debug!("Seq no {} missing after {} attempts", seq, p.nack_count);
@@ -92,7 +91,11 @@ impl NackRegister {
             self.packet(s.into()).reset();
         }
 
-        self.active = Some(next_active);
+        if (start..=end).contains(&seq) {
+            self.packet(seq).mark_received();
+        }
+
+        self.active = Some(start..end);
 
         new
     }
@@ -245,7 +248,6 @@ mod test {
         // future packet accepted, sliding window
         let next = 12 + MAX_MISORDER;
         assert_update(&mut reg, next, true, true, 12..next);
-        assert!(!reg.packet(11.into()).received);
 
         // older packet received within window
         let next = 13;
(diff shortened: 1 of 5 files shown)


src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+pub struct NackRegister {
+    /// Status of packets indexed by wrapping SeqNo.
+    packets: Vec<PacketStatus>,
+
+    /// Range of seq numbers considered NACK reporting.
+    active: Option<Range<SeqNo>>,
+}
+
+#[derive(Debug, Default, Clone, Copy)]
+struct PacketStatus {
```
k0nserv, line 27: Nitpic: This ends up being 2 bytes in size because of the `received` `bool`. We could cut that in half by storing the received flag in the most significant bit. So instead of being 6KB in memory it would be 3KB.

src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+impl NackRegister {
+    pub fn new() -> Self {
+        NackRegister {
+            packets: vec![PacketStatus::default(); MAX_DROPOUT as usize],
+            active: None,
+        }
+    }
+
+    pub fn update(&mut self, seq: SeqNo) -> bool {
+        let Some(active) = self.active.clone() else {
```
k0nserv, line 58: `Range` not being `Copy` when `Idx` is is one of those weird quirks of Rust. If we aren't heavily relying on the Range functionality we could create our own type that could be `Copy` instead

src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+            self.packet(s.into()).reset();
+        }
+
+        self.active = Some(next_active);
+
+        new
+    }
+
+    pub fn max_seq(&self) -> Option<SeqNo> {
+        Some(self.active.as_ref()?.end)
```
k0nserv, line 101: nit: 

```rust
self.active.as_ref().map(|a| a.end)
```

Feels a bit easier to read I think

src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+    ///
+    /// This modifies the state as it counts how many times packets have been nacked
+    pub fn nack_report(&mut self) -> Option<Vec<Nack>> {
+        let active = self.active.as_ref()?;
+
+        if active.is_empty() {
+            return None;
+        }
+
+        let mut nacks = vec![];
```
k0nserv, line 114: `ReportList` is designed to have a fixed size so it seems wasteful to allocate here. Instead we can use `ReportList::new` and then `ReportList::push`.

Also the result of this method gets pushed onto a `&mut VecDeque<Rtcp>` that is allocated a single time in `Session`. We could entirely avoid allocating here we changed the signature to

```rust
pub fn nack_report(&mut self, on_nack: F)
where 
  F: FnMut(nack: Nack)
```

src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+                        sender_ssrc: 0.into(),
+                        ssrc: 0.into(), // changed when sending
+                        reports,
+                    }
+                })
+                .collect(),
+        )
+    }
+
+    fn as_index(&self, seq: SeqNo) -> usize {
```
k0nserv, line 162: nit: Could also implement `IndexMut<SeqNo>` for `NackRegister`

src/streams/register_nack.rs
```diff
@@ -0,0 +1,560 @@
...
+    }
+
+    #[test]
+    fn active_window_sliding() {
+        let mut reg = NackRegister::new();
+
+        assert_update(&mut reg, 10, true, true, 10..10);
+
+        // packet before window start is ignored
+        assert_update(&mut reg, 9, false, false, 10..10);
```
k0nserv, line 234: Does this means we could lose out on some packets if we get unlucky with re-ordering at the start of an RTP stream? i.e. if packets 11, 10, 9 arrive in quick succession in that order we'd ignore 10 and 9. At the start of a video stream these are also likely to be part of a keyframe so their loss is particularly problematic.

k0nserv (approved):
Some assorted thoughts, mostly just nitpicks and potential performance improvements

src/streams/register_nack.rs
algesten: Maybe make it work before optimizing like that?

Will make it harder to read.

src/streams/register_nack.rs
algesten: Since this is tightly coupled to the RTCP type, I don't mind making it produce the specific type directly.

algesten (commented):
If we are merging everything together again, I want the name `register.rs` :)

src/streams/register_nack.rs
k0nserv: So

```rust
pub fn nack_report(&mut self, on_nack: F)
where 
  F: FnMut(rtcp: Rtcp)
```
?

algesten: Either the way we collect feedback through the rest of handle_timeout

```rust
pub fn nack_report(&mut self, feedback: &mut VecDeque<Rtcp>) { … }
```

or

```rust
pub fn nack_report(&mut self) -> Option<Rtcp> { … }
```

src/streams/register_nack.rs
davibe: Correct

algesten: This was already the case, so it's not better or worse.

src/streams/register_nack.rs
k0nserv: Aight, I guess the former is slightly more flexible since `nack_report` could, in theory, produce two instances of `Rtcp`. But I guess we might not want to cause that much NACKing anyway

src/streams/register_nack.rs
k0nserv: Is it something we want to address long term?

algesten: Could we detect and `warn!` to see how much better it'd be?

src/streams/register_nack.rs
k0nserv: Not needing to clone here and also not needing to do the `as_ref` business elsewhere

src/streams/register_nack.rs
davibe: @k0nserv we would not ignore packets as you wrote. The comment there is from the perspective of nacking. We would not consider them for nacking. 
So `11, 10, 9` is fine (no loss). An issue would be if we receive `11, 9` at the start because we would not nack `10`. However even if we received `11, 10`, we would not nack `9` because .. how far back should we go ?

davibe: yes. also what i said before is not 100% accurate, because we also don't flush them out in RTP mode (based on the fact that the register does not register them as new). 

Needs more thought.

davibe: I have made it so that the window extends back if the packet is within` MAX_MISORDER`, or `warn!` otherwise

davibe: There are issues with this, I have ideas how to solve but I put it in a separate PR.

algesten (approved):
LGTM! ✨

PR #350, opened 2023-09-13
Simplify logging on SDP a=ssrc change

algesten:
The `change_ssrc` and `maybe_reset_rtx` already do logging. This
removes some extra logging that is already captured.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1,7 +1,6 @@
 //! Strategy that amends the [`Rtc`] via SDP OFFER/ANSWER negotiation.
 
 use std::fmt;
-use std::fmt::Write as _;
 use std::ops::{Deref, DerefMut};
 
 use crate::channel::ChannelId;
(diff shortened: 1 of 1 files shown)


src/change/sdp.rs
```diff
@@ -1045,36 +1032,20 @@ fn update_media(
...
+            let from = stream.ssrc();
+            streams.change_stream_rx_ssrc(from, i.ssrc);
         } else {
-            log_change(
-                format!("Adding pre-communicated SSRC: {}", i.ssrc),
-                repair_ssrc,
-                mid,
-                rid,
+            info!(
+                "Adding pre-communicated SSRC: {:?} RTX: {:?} MID: {}",
```
k0nserv, line 1046: This no longer logs rid, I was trying to be consistent with the other places were we log these SSRC association events

algesten: Good point. Let's add that.

PR #352, opened 2023-09-17
receive: allow nacks so we discover rtx

davibe:
I can reproduce this with Safari. Safari sometimes sends rtx probes before we discover the corresponding "main ssrc" (so we can't pick it up). If we don't send nacks there will never be an RTX active and the stream gets corrupted with minimal loss

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -318,6 +318,15 @@ impl Session {
             return Some((mid, ssrc_main));
         }
 
+        // Based on the pt, we may expect this stream to have an RTX ssrc.
+        // Using a fake rtx ssrc value we allow nacking and we will pick up the
+        // correct ssrc later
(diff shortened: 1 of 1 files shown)


src/streams/receive.rs
```diff
@@ -524,7 +524,7 @@ impl StreamRx {
     }
 
     fn nack_enabled(&self) -> bool {
-        self.rtx.is_some() && !self.suppress_nack
+        !self.suppress_nack
```
algesten, line 527: This is not correct, because it would enable nack also for streams that we don't want nack for.

algesten (changes requested):

src/streams/receive.rs
davibe: audio ?

algesten (approved):
LGTM!

PR #353, opened 2023-09-18
Avoid overwriting RTX SSRC with 0-hack

algesten:


src/session.rs
```diff
@@ -318,14 +318,27 @@ impl Session {
...
+            // If the remote does NOT want RTX, this will have been negotiated away.
+            let is_rtx_negotiated = self
+                .codec_config
+                .iter()
+                .find_map(|c| (c.pt == header.payload_type).then_some(c.resend().is_some()))
+                .unwrap_or(false);
+
+            // We will only try this hack the first time the stream is declared, so not override a good
+            // RTX SSRC with 0 later.
+            let stream_not_declared = maybe_stream.is_none();
```
davibe, line 336: we cannot get here if maybe_stream is some

davibe (approved):

PR #359, opened 2023-09-19
Payloader: rebase rtp_time to the codec's clockrate

david-flok:
Previously, it was assumed that the denominator of RTP time passed to media writers was the clock rate of the media but this is not necessarily the case. As the RTP packet is prepared for writing, this adjustment (if any) should be done. To accomplish this, `Payloader` now takes a whole `CodecSpec` and produces the packetizer and a copy of the clock_rate parameter from this. This will allow the `MediaTime` rational number abstraction to function in the API as expected and make working with, e.g., pre-recorded source streams easier.


algesten (approved):
LGTM!

This won't break the API for people that already used he RTP time, so it's a nice improvement!

PR #360, opened 2023-09-20
VP8 L1T3 compat

davibe:
This PR improves how we wait for retransmissions and detect discontinuities in our sample-based api (MediaData). This is relevant for VP8 when temporal layers are enabled (simulcast) in case of network loss.

 Fixes issue with minor corruption on Chrome improves major corruption on Safari (it’s less frequent and when it happens the discontinuity is reported).

### DIFF
--- a/src/packet/buffer_rx.rs
+++ b/src/packet/buffer_rx.rs
@@ -1,3 +1,4 @@
+use core::panic;
 use std::collections::VecDeque;
 use std::fmt;
 use std::ops::RangeInclusive;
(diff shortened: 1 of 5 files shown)


src/packet/buffer_rx.rs
```diff
@@ -167,80 +174,45 @@ impl DepacketizingBuffer {
...
-
-        let is_more_than_hold_back = self.segments.len() >= self.hold_back;
+        // If we have contiguity of seq numbers we emit right away,
+        // Otherwise, we wait for retransmissions up to `hold_back` frames
+        // and re-evaluate contiguity based on codec specific information
 
-        let wait_for_contiguity = !contiguous && !is_more_than_hold_back;
+        let more_than_hold_back = self.segments.len() >= self.hold_back;
+        let contiguous_seq = self.is_following_last(start);
+        let wait_for_contiguity = !contiguous_seq && !more_than_hold_back;
```
algesten, line 183: The amount of changes here seem a bit "not invented here" – unnecessary name change etc.

algesten (approved):
Seems ok

src/packet/buffer_rx.rs
davibe: Anything specific I can address ?

PR #361, opened 2023-09-20
Move Ssrc -> (Mid, Ssrc) mappping to `Streams`

k0nserv:


src/streams/mod.rs
```diff
@@ -441,15 +484,109 @@ impl Streams {
...
+        // Add mapping for new main SSRC
+        self.associate_ssrc_mid(ssrc_to, mid, ssrc_to, None);
+
+        // Map the RTX to the new main SSRC
+        if let Some(rtx) = rtx {
+            self.associate_ssrc_mid(rtx, mid, ssrc_to, None);
+        }
     }
+
+    pub(crate) fn debug(&self) {
```
k0nserv, line 516: Hmm this was supposed to be excluded...

src/streams/mod.rs
```diff
@@ -441,15 +484,109 @@ impl Streams {
         }
     }
 
-    pub(crate) fn change_stream_rx_ssrc(&mut self, ssrc_from: Ssrc, ssrc_to: Ssrc) {
+    pub(crate) fn change_stream_rx_ssrc(
+        &mut self,
+        ssrc_from: Ssrc,
+        ssrc_to: Ssrc,
+        rtx_to: Option<Ssrc>,
+    ) {
```
algesten, line 492: I prefer having a separate function for changing RTX, this overloads the purpose a bit too much.

algesten (approved):
Just one thing

src/streams/mod.rs
```diff
@@ -198,16 +214,42 @@ impl Streams {
...
+
+            // SSRC is mapped to a Sender/Receiver in this media. Make an entry for it.
+            associate_ssrc_mid(
+                &mut self.source_keys_rx,
+                ssrc,
+                mid,
+                ssrc_main,
+                Some("Known SSRC"),
+            );
+            self.debug();
```
k0nserv, line 245: Also should go

src/streams/mod.rs
k0nserv: Aight, we have a quite a lot of methods that take an optional `rtx: Option<Ssrc>` so I tried to mirror that

algesten (approved):

PR #364, opened 2023-09-21
Enhance SDP Generation Flexibility

evdokimovs:
This pull request addresses limitations in the current SDP API. While it's secure, it lacks flexibility in certain cases. `str0m` also have a Direct API for media source creation and etc, but it doesn't support manual SDP generation. To improve this, we propose adding raw APIs for SDP generation and applying.

While Direct API may not be the ideal place for SDP offers, I'm open to alternative suggestions for implementation. Your input is welcome!

### DIFF
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -1,14 +1,14 @@
-use crate::change::sdp::{accept_answer, Changes, create_offer};
+use crate::change::sdp::{accept_answer, create_offer, Changes};
 use crate::channel::ChannelId;
 use crate::dtls::Fingerprint;
 use crate::ice::IceCreds;
 use crate::media::{Media, MediaKind};
 use crate::rtp_::{Mid, Rid, Ssrc};
 use crate::sctp::ChannelConfig;
+use crate::sdp::{SdpAnswer, SdpOffer};
 use crate::streams::{StreamRx, StreamTx, DEFAULT_RTX_CACHE_DURATION};
 use crate::Rtc;
 use crate::RtcError;
-use crate::sdp::{SdpAnswer, SdpOffer};
 
 /// Direct change strategy.
 ///
@@ -149,7 +149,13 @@ impl<'a> DirectApi<'a> {
         };
 
         let exts = self.rtc.session.exts.cloned_with_type(kind.is_audio());
-        let pts = self.rtc.session.codec_config.all_for_kind(kind).map(|p| p.pt()).collect();
+        let pts = self
+            .rtc
+            .session
+            .codec_config
+            .all_for_kind(kind)
+            .map(|p| p.pt())
+            .collect();
         let m = Media::from_direct_api(mid, next_index, kind, exts, pts);
 
         self.rtc.session.medias.push(m);
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -244,10 +244,12 @@ impl<'a> DirectApi<'a> {
         stream
     }
 
+    /// Creates new [`SdpOffer`] from the `str0m`'s state.
     pub fn create_offer(&mut self) -> SdpOffer {
         create_offer(self.rtc, &Changes(Vec::new()))
     }
 
+    /// Accept an answer to a previously created [`SdpOffer`].
     pub fn accept_answer(&mut self, answer: SdpAnswer) -> Result<(), RtcError> {
         accept_answer(self.rtc, answer)
     }
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -1,3 +1,4 @@
+use crate::change::sdp::{accept_answer, Changes, create_offer};
 use crate::channel::ChannelId;
 use crate::dtls::Fingerprint;
 use crate::ice::IceCreds;
@@ -7,6 +8,7 @@ use crate::sctp::ChannelConfig;
 use crate::streams::{StreamRx, StreamTx, DEFAULT_RTX_CACHE_DURATION};
 use crate::Rtc;
 use crate::RtcError;
+use crate::sdp::{SdpAnswer, SdpOffer};
 
 /// Direct change strategy.
 ///
@@ -147,7 +149,8 @@ impl<'a> DirectApi<'a> {
         };
 
         let exts = self.rtc.session.exts.cloned_with_type(kind.is_audio());
-        let m = Media::from_direct_api(mid, next_index, kind, exts);
+        let pts = self.rtc.session.codec_config.all_for_kind(kind).map(|p| p.pt()).collect();
+        let m = Media::from_direct_api(mid, next_index, kind, exts, pts);
 
         self.rtc.session.medias.push(m);
         self.rtc.session.medias.last_mut().unwrap()
@@ -239,6 +242,14 @@ impl<'a> DirectApi<'a> {
         stream.set_rtx_cache(size, DEFAULT_RTX_CACHE_DURATION);
 
         stream
+    }
+
+    pub fn create_offer(&mut self) -> SdpOffer {
+        create_offer(self.rtc, &Changes(Vec::new()))
+    }
+
+    pub fn accept_answer(&mut self, answer: SdpAnswer) -> Result<(), RtcError> {
+        accept_answer(self.rtc, answer)
     }
 
     /// Remove the transmit stream for the given SSRC.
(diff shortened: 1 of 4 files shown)


algesten:
Why is this better? You need to go into more detail about the problem you're trying to solve.

evdokimovs:
@algesten 

## Uselessness of Direct API at this moment

The Direct API allows you to add media sources, but it lacks any means for negotiation if you have added media sources through it. Consequently, it seems that there are no practical use cases for this API at this moment.

## Race conditions

In WebRTC, it's more common to create `Transceiver`s immediately rather than waiting until after the negotiation is complete. This is because delaying the creation of `Transceiver`s can lead to race conditions. For instance, consider a scenario where you attempt to achieve something like perfect negotiation:

1. SdpApi.addMedia()
2. SdpApi.apply() // <- negotiation has started
3. SdpApi.acceptOffer // <- all changes from first negotation are dropped
4. As a result, you lose all the changes you made to the media

To address this issue, the process of creating Transceivers should be separated from the SDP offer creation process.

## More freedom for `str0m` user

It's adds more flexibility to working with offers. You can't predict all use cases of you library, so it's better to add raw APIs, so the users can bypass safe APIs on their own risk. As I understand idea of Direct API, it was created to give users that raw API, so they can implement their safe wrappers if their vision on working with something is not the same.

algesten:
> Uselessness of Direct API at this moment

Not sure I agree it's "useless". The Direct API has one primary purpose: to configure a WebRTC session _without using SDP_. This is exactly the use case xnorpx (and his company) are using it for. A secondary use case is SDP + `write_rtp`, but the SDP part would be through the normal SDP API.

> Race conditions

str0m has no equivalent of "transceivers" (bidirectional things) and using `sdp_api.add_media` does not alter any state inside str0m. The state changes will only happen on a completed negotiation, it should not be possible to create a race condition using the SDP API. If you found one, that's a bug.

I think you're trying to use the Direct API followed by SDP negotiation to make both sides have the same state. This is not the intended use case of direct API. If you use SDP between str0m and the client, the SDP API should give you all control you need. There might be functions missing, that would be useful to know so we can address it.

> More freedom for str0m user
> It's adds more flexibility to working with offers. You can't predict all use cases of you library, so it's better to add raw APIs, so the users can bypass safe APIs on their own risk.

Bypassing the safe API is kinda the issue here. SDP is a terrible thing. Superficially it might look like it represents the complete setup of the session, but it actually doesn't. Each side have their own view of it [read more here](https://github.com/algesten/str0m/blob/2deb5f08fc864f64764c23e9d3d397f0f2596d72/SDP.md), and there are also other properties in the SDP that are similar.

RTCPeerConnection lets you make changes to properties of transceivers, senders, receivers etc. Those changes are then compared with current values and if a delta is discovered, the `negotiationneeded` event is triggered. str0m has no intention of working like that. str0m does not diff an old state with a new state to discover a required SDP negotiation.

Instead, str0m encapsulates the necessary operations in the safe SDP API. Those operations can be thought of as delta modifications that are applied upon successful negotiation. Thus there is always "perfect negotiation", since it's not possible (using the SDP API) to make any other state.

I haven't thought through the consequences of your proposed changes, but I think they will lead to situations where we must compare old/new values to create the correct SDP. It's not the architecture I want for str0m.

### Conclusion

It would be good to know what functions of Direct API you're reaching for that you can't make through the SDP API. We're very happy to make the SDP API better.

algesten:
I've tried to clarify this in the rust doc: https://github.com/algesten/str0m/blob/main/src/change/direct.rs#L11-L31

evdokimovs:
@algesten ,

> Not sure I agree it's "useless".

By "useless" I mean that if you add media through Direct API, you can't generate any SDP out of it. I understand that this is useful in some cases, but it's just very unintuitive when you see API which adds media, but you can't generate SDP after it. And SDP API will require you to add at least one `Change` or it will not generate SDP offer.


> str0m has no equivalent of "transceivers" (bidirectional things) and using sdp_api.add_media does not alter any state inside str0m. The state changes will only happen on a completed negotiation

Yep, I understand that these changes are not applied on the `str0m` state until negotiation is finished. Basically, it is where our problems are coming. We're migrating from `webrtc-rs` and we're thinking more in WebRTC way (like `PeerConnection`, `Transceiver`, `MediaTrack` etc), so we want behavior in which we create `Transceiver` before negotiation start and can store it and know that it will be alive all the time (even if first negotiation in my example will be overlapped by second negotiation, `Transceiver` should be alive).

I know that this is not a very good argument, but I think that many people who will migrate from `webrtc-rs` (and I think that many people will do it eventually, because architecture of `str0m` is way better) will encounter this problem.

> create a race condition

I used "race condition" not in meaning that something brokes in `str0m`. Here is more detailed example:

1. Server adds media with `sdp_api.add_media`
2. Server sends offer generated by `sdp_api.apply` to a client
3. At the same time, client also wants to do negotiation and sends SDP offer to the server, before receiving SDP offer from server
4. Server received offer and starts new process of negotiation
5. All media which should be added by the first negotiation is now lost, because all `Changes` are now outdated

It's synthetic example, we don't have exact same problem, but it shows problem with your approach to aborting negotiation. It's just inconvenient that you added some `Media` but you can't be sure that it's added.

> I think they will lead to situations where we must compare old/new values to create the correct SDP.

My idea was to add more low level APIs for SDP based on which user will manage these kind of things on their own. But maybe I just missing some problems. Also I understand that making low level APIs is not a very good thing, and you want your library to be very safe, so the user can't broke anything, but problem is that you can't do safe API which will be usable for anyone. 

So idea is to make something like `unsafe` module (like in Rust). Maybe you see problems here which I'm not seeing because I don't know all code base of `str0m`, but based on what I've seen, it shouldn't bring problems to `str0m`.

xnorpx:
As @algesten said we are very happy with the Direct API. Our servers does not use SDP at all so I don't agree with your statement about useless.

algesten:
Ah. I see the problem. 

What if `PendingOffer` had a call `retry()` that would reapply the changes you're trying to make?

evdokimovs:
@algesten ,

Yep, it should work. But also in this case, it will be more convenient for our use case if `PendingOffer` will provide some methods for getting list of newly created `Media`s which are waiting for negotiation end. Basically at this moment we only need this things: `mid`, `kind` and `direction`. I know that we can get this things from `SdpOffer` but then it will be more complicated because we will need somehow to filter out old `Media`s.

algesten:
Part of this might be addressed by #370

algesten:
@evdokimovs at this point I don't want to make an API for inspecting the `SdpPendingOffer`. The reason is that it would be a lot of API surface area for something that could be handled by the caller remembering what changes have been made.

In code terms, I don't want to make the `Changes` struct with the `Change` enum into public API.

evdokimovs:
@algesten ,

`SdpPendingOffer::retry` fixes previous scenario, but we still have another problem:

1. Server adds media with `sdp_api.add_media`
2. Server sends offer generated by `sdp_api.apply` to a client
4. Server adds another media with `sdp_api.add_media`
5. Server sends offer generated by `sdp_api.apply` to a client
6. Client receives first SDP offer and sends answer back to server
7. Client receives second SDP offer and brokes, because second SDP offer doesn't have `Transceiver` which was added by the first `add_media`

algesten:
It's by design that only one single change -> `SdpOffer` + `SdpPendingOffer` is valid at a time. That means step 4 above invalidates step 2.

evdokimovs:
@algesten ,

So, are you saying that the SDP API should be "locked" when negotiation has already begun? Right now, it seems like the SDP API is designed to handle such cases as they come up. From what I can tell, it looks like you were trying to implement something similar to an [SDP offer rollback][1], but without actually changing the state. However, the way it currently works might not be correct, because it can remove `Transceiver`s when there are overlapping changes. Removing (and reordering) `Transceiver`s from an SDP offer is a bit like undefined behavior, and it's generally recommended not to do that, as it could potentially break the client side.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription/type#rollback

algesten:
The way I see it: `SdpOffer` + `SdpPendingOffer` are only valid as long as this holds true:

1. No incoming OFFER from the remote peer has been applied to the session (though we have the `retry()` function here to retry relevant changes).
2. No new `SdpOffer` + `SdpPendingOffer` is created.

The API tries to "lock down" in that `SdpPendingOffer` has an internal number that must be correct compared to a change number inside `Rtc`.

PR #370, opened 2023-09-24
SdpPendingOffer::retry to attempt changes again

algesten:


src/change/sdp.rs
```diff
@@ -429,6 +429,47 @@ pub struct SdpPendingOffer {
     changes: Changes,
 }
 
+impl SdpPendingOffer {
+    /// Retry this `SdpPendingOffer`.
+    ///
+    /// This is useful for this sequence of events.
+    ///
+    /// 1. str0m creates an OFFER with some changes (this results in `SdpPendingOffer`).
+    /// 2. Client creates an OFFER with other changes.
```
xnorpx, line 438: Client in this case means the other peer? Maybe use Alice and Bob :)

algesten: Other peer is good. Not sure I'm ready for Alice and Bob :)

algesten:
Use `merge` from #371 instead.

PR #371, opened 2023-09-25
SdpPendingOffer merging

evdokimovs:
Related to #364

## Synopsis

Sometimes, when using `str0m`, users may need to run a a second negotiation while first one is still in progress. 

## Implementation

To make this possible, I suggest adding an API that lets us combine a `SdpPendingOffer` with new changes from `SdpApi`. This combination will produce a new `SdpOffer` and a fresh `SdpPendingOffer`. These new offer will include changes from the previous `SdpPendingOffer` as well as new changes from `SdpApi`. The original `SdpPendingOffer` will still be available for users, allowing them to cancel any `SdpAnswer` related to the previous `SdpOffers`. This can be done using the `ChangeId` stored in the original `SdpPendingOffer`.

It's important to note that these older `SdpPendingOffers` will be empty and won't make any changes. Only the latest `SdpPendingOffer` generated by the `SdpApi::merge_and_apply` method can be used to update the `str0m` state and will be accepted in the `SdpApi::accept_answer` process.

The ordering of `m-lines` in the SDP offer will also be accurate.

Example of usage of this new API can be found in the `SdpApi::merge_and_apply` doc.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -411,10 +411,8 @@ impl<'a> SdpApi<'a> {
     /// ## Example
     ///
     /// ```no_run
-    ///
     /// # use str0m::media::{Direction, MediaKind};
     /// # use str0m::Rtc;
-    ///
     /// let mut rtc = Rtc::new();
     /// let mut changes = rtc.sdp_api();
     /// changes.add_media(MediaKind::Audio, Direction::SendOnly, None, None);
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -475,7 +475,7 @@ impl<'a> SdpApi<'a> {
         mut self,
         pending_offer: &mut SdpPendingOffer,
     ) -> Option<(SdpOffer, SdpPendingOffer)> {
-        pending_offer.retain_relevant(&self.rtc);
+        pending_offer.retain_relevant(self.rtc);
 
         pending_offer.changes.extend(self.changes.drain(..));
         swap(&mut self.changes, &mut pending_offer.changes);
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1,6 +1,7 @@
 //! Strategy that amends the [`Rtc`] via SDP OFFER/ANSWER negotiation.
 
 use std::fmt;
+use std::mem::swap;
 use std::ops::{Deref, DerefMut};
 
 use crate::channel::ChannelId;
@@ -401,6 +402,86 @@ impl<'a> SdpApi<'a> {
             None
         }
     }
+
+    /// Combines the modifications made in [`SdpApi`] with those in [`SdpPendingOffer`], generating
+    /// a new [`SdpOffer`] and [`SdpPendingOffer`].
+    ///
+    /// This function merges the changes present in [`SdpApi`] with the changes
+    /// in [`SdpPendingOffer`]. The resulting [`SdpOffer`] incorporates modifications
+    /// from both the previous [`SdpPendingOffer`] and any newly added changes.
+    ///
+    /// The original [`SdpPendingOffer`] is intentionally not consumed by this function to
+    /// allow users the flexibility to retain it for case when [`SdpAnswer`] for the
+    /// old [`SdpPendingOffer`] is received. It will not affect any state, but it can be used
+    /// to discard outdated [`SdpAnswer`]s.
+    ///
+    /// ## Example
+    ///
+    /// ```no_run
+    /// # use std::collections::VecDeque;
+    /// # use str0m::{Rtc, RtcError};
+    /// # use str0m::media::{MediaKind, Direction};
+    /// # use str0m::change::{SdpAnswer, SdpOffer, SdpPendingOffer};
+    ///
+    ///
+    /// struct MediaController {
+    ///     rtc: Rtc,
+    ///     pending_offers: VecDeque<SdpPendingOffer>,
+    /// }
+    ///
+    /// impl MediaController {
+    ///     pub fn new() -> Self {
+    ///         Self {
+    ///             rtc: Rtc::new(),
+    ///             pending_offers: VecDeque::new(),
+    ///         }
+    ///     }
+    ///
+    ///     pub fn add_media(&mut self) -> SdpOffer {
+    ///         let mut changes = self.rtc.sdp_api();
+    ///         changes.add_media(MediaKind::Audio, Direction::SendOnly, None, None);
+    ///         if self.pending_offers.is_empty() {
+    ///             let (offer, pending) = changes.apply().unwrap();
+    ///             self.pending_offers.push_front(pending);
+    ///             offer
+    ///         } else {
+    ///             let last_pending_offer = self.pending_offers.back_mut().unwrap();
+    ///             let (offer, pending) = changes.merge_and_apply(last_pending_offer).unwrap();
+    ///             self.pending_offers.push_back(pending);
+    ///             offer
+    ///         }
+    ///     }
+    ///
+    ///     pub fn accept_answer(&mut self, answer: SdpAnswer) -> Result<(), RtcError> {
+    ///         let pending = self.pending_offers.pop_front().unwrap();
+    ///         self.rtc.sdp_api().accept_answer(pending, answer)
+    ///     }
+    /// }
+    ///
+    /// let mut media_controller = MediaController::new();
+    ///
+    /// let first_offer = media_controller.add_media();
+    /// let first_answer: SdpAnswer = todo!();
+    /// let second_offer = media_controller.add_media();
+    /// let second_answer: SdpAnswer = todo!();
+    /// let third_offer = media_controller.add_media();
+    /// let third_answer: SdpAnswer = todo!();
+    ///
+    /// media_controller.accept_answer(first_answer).unwrap_err();
+    /// media_controller.accept_answer(second_answer).unwrap_err();
+    /// media_controller.accept_answer(third_answer).unwrap();
+    /// ```
+    pub fn merge_and_apply(
+        mut self,
+        pending_offer: &mut SdpPendingOffer,
+    ) -> Option<(SdpOffer, SdpPendingOffer)> {
+        pending_offer.retain_relevant(&self.rtc);
+
+        pending_offer.changes.extend(self.changes.drain(..));
+        swap(&mut self.changes, &mut pending_offer.changes);
+
+        self.apply()
+    }
 }
 
 /// Pending offer from a previous [`Rtc::sdp_api()`] call.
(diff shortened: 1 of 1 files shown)


algesten:
I like the direction in which this is going. This idea replaces #370. I think the signature we want is one of:

* `SdpApi::add_pending(&mut self, pending: SdpPendingOffer)`
* `SdpApi::merge(&mut self, pending: SdpPendingOffer)`
* `SdpApi::append(&mut self, pending: SdpPendingOffer)`

Crucially I think the function should not automatically apply, but just extend. I also think it should take `SdpPendingOffer` by value, not reference to not allow the same to be applied multiple times.

evdokimovs:
@algesten ,

> Crucially I think the function should not automatically apply, but just extend.

I also wanted this function to extend `SdpApi` instead of applying, and I've done it at first, but then I realized that we need to take into account `m-lines` ordering after extend, so if we will allow to add more `m-line`s to SDP (e.g. by `SdpApi::add_media` call) after extending, then we can't guarantee that `m-line`s ordering will be correct at the end.

> I also think it should take SdpPendingOffer by value, not reference to not allow the same to be applied multiple times.

Yep, this is also was my first taught but then user will not have any mechanism to discard `SdpAnswer`s for the old values. Those `SdpPendingOffer`s is empty by design, so they can't make any changes. Also they can't make any chages because they will have outdated `ChangeId`s, and `accept_answer` for example will return error (and this is useful for filtering SDP answers for the old SDP offers). I know that this is not the best approach, but I can't come up with a better idea for filtering old SDP answers. I appreciate any ideas. 

Also you can see my example use case in the docs of `SdpApi::merge_and_apply` function for more details. Maybe it will clarify some things for you.

algesten:
> I also wanted this function to extend SdpApi instead of applying, and I've done it at first, but then I realized that we need to take into account m-lines ordering after extend, so if we will allow to add more m-lines to SDP (e.g. by SdpApi::add_media call) after extending, then we can't guarantee that m-lines ordering will be correct at the end.

The changes in `SdpApi` are _not yet applied_. The order of m-lines is only relevant for already negotiated. Hence the order of these "floating changes" doesn't matter.

> Yep, this is also was my first taught but then user will not have any mechanism to discard SdpAnswers for the old values.

Not sure I follow. What do you mean?

evdokimovs:
@algesten ,

> The changes in `SdpApi` are not yet applied

Changes from `SdpPendingChange`s can be already applied on client side because of SDP offer. So if we will not honor `m-lines` ordering then client can crash on the newly generated `SdpOffer`.

> Not sure I follow. What do you mean?

`SdpOffer`s may be already sent to the client side, so client side will apply them before `str0m` does. So when `str0m` will generate new `SdpOffer` with the same changes, `m-lines` ordering should be the same as it was in the first `SdpOffer`.

Here is example implementation of this case:

```rust
impl MediaController {
    pub fn new() -> Self {
        Self {
            rtc: Rtc::new(),
            pending_offers: VecDeque::new(),
        }
    }

    pub fn add_media(&mut self) -> SdpOffer {
        let mut changes = self.rtc.sdp_api();
        changes.add_media(MediaKind::Audio, Direction::SendOnly, None, None);
        if self.pending_offers.is_empty() {
            let (offer, pending) = changes.apply().unwrap();
            self.pending_offers.push_front(pending);
            offer
        } else {
            let last_pending_offer = self.pending_offers.back_mut().unwrap();
            let (offer, pending) = changes.merge_and_apply(last_pending_offer).unwrap();
            self.pending_offers.push_back(pending);
            offer
        }
    }

    pub fn accept_answer(&mut self, answer: SdpAnswer) -> Result<(), RtcError> {
        let pending = self.pending_offers.pop_front().unwrap();
        self.rtc.sdp_api().accept_answer(pending, answer)
    }
}

let mut media_controller = MediaController::new();

let first_offer = media_controller.add_media();
let first_answer = SdpAnswer::from((*first_offer).clone());
let second_offer = media_controller.add_media();
let second_answer = SdpAnswer::from((*second_offer).clone());
let third_offer = media_controller.add_media();
let third_answer = SdpAnswer::from((*third_offer).clone());

media_controller.accept_answer(first_answer).unwrap_err();
media_controller.accept_answer(second_answer).unwrap_err();
media_controller.accept_answer(third_answer).unwrap();
```

algesten:
This isn't how the API is supposed to be used. You can't have multiple pending offers. You must wait for the answer to the first offer before making the next change. The sequence is:

1. change
2. offer
3. answer

or

1. change
2. offer
3. reject/abort

I.e. the API is not meant to support:

1. change -> offer
2. change -> offer
3. change -> offer
4. answer
5. answer
6. answer

evdokimovs:
@algesten ,

> You can't have multiple pending offers.

I understand it, but it causes too many problems while working with this API. We don't want to lock `Media` changes, because this lock will depend on the client and this is not a good thing. I can provide you more real life example:

1. Member `Carol` connects to the call
1.1 Server creates `SendOnly` `Media`s for the `Carol` 
1.2 Server starts negotiation of `Carol`
3. Member `Carlos` connects to the call
3.1 Server creates `SendOnly` `Media`s for the `Carlos`
3.2 Server creates `RecvOnly` `Medias` for the `Carol`
3.3 Server waits for `Carol` negotiation because offer is already sent
3.4 Server starts REnegotation of `Carol`
3.5 Server starts negotiation of `Carlos`
4. Member `Charlie` connects to the call
4.1 Server creates `SendOnly` `Media`s for the `Carol` and `Carlos`
4.2 But offers are already sent to this `Member`s
4.3 Everything the same as for member `Carlos`, but in this case we will also need to REnegotiate `Carlos` and wait for every negotiation which `Carlos` waits

All this members connects to the server simultaneously, so in this example `Charlie` will also wait for negotiation of `Carol`, then for REnegotiation for `Carol` and for negotiation of `Carlos`. And then multiply it on number of `Member`. Waiting time of the last connected member will grow exponentially.

Also it can cause vulnerabilities. If, for example `Carol` will be attacker then she can just not send `SdpAnswer` and then `Charlie` will not connect at all. 

Also long `await`ing on the adding media action can cause many problems on server because of "locking" some important loops.

algesten:
We are running one of these style WebRTC implementations on str0m, in production, and we don't have problems with large meetings. We have a different way we do the negotiations than you outline above. It's pretty much exactly like in this example: https://github.com/algesten/str0m/blob/main/examples/chat.rs#L423-L440

1. Keep track of outgoing tracks needing negotiation. 
2. If there are tracks needing negotiation, make OFFER.

3a. If peer sends OFFER before we get ANSWER, abort negotiation and retry from 1.
3b If peer sends ANSWER, apply it and mark negotiated tracks as ready to use.

We do not await any changes or stall the clients while this is going on.

evdokimovs:
@algesten ,

> 3a. If peer sends OFFER before we get ANSWER, abort negotiation and retry from 1.

1. Server sends OFFER to the client
2. Client receives OFFER and generates ANSWER
3. Server wants to make renegotiation before receiving ANSWER

What will happen in this case?

evdokimovs:
@algesten ,

We just can't use this approach. It's good, but we already have too many layers of code which requires SDP offer at the time when we add some new medias. If we're writing all this from scratch then there will be no problems, but we already have P2P and SFU, we just migrating to `str0m` and don't want to break it all.

I've thought about this approach many times, but with this approach we have only options to `await` on `add_media`s calls until negotiation finishes. But this will definitely cause many bugs and performance issues. Or we can make crutches to somehow bypass your safe SDP API. But this will cause many issues in the future when `str0m` will make changes in how things works in SDP API.

evdokimovs:
@algesten ,

I've reworked this PR as you proposed. You can check it out.

evdokimovs:
@algesten ,

Also what you think about adding ability to set `mid` and `ssrc` of `Media` when adding it with `add_media`? So signature of `SdpApi::add_media` function will be:

```rust
    pub fn add_media(
        &mut self,
        kind: MediaKind,
        dir: Direction,
        stream_id: Option<String>,
        track_id: Option<String>,
        mid: Option<Mid>,
        ssrc: Option<Ssrc>,
    ) -> Mid
```

Or we can do it even better with builder pattern for creating `Media`. Something like this:

```rust
struct MediaBuilder {
  // ...
}

impl MediaBuilder {
  pub fn new(kind: MediaKind, direction: Direction) -> Self { /* ... */ }
  pub fn ssrc(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn stream_id(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn ssrc(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn track_id(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
}

impl SdpApi {
  pub fn add_media_with_builder(&mut self, media_builder: MediaBuilder) -> Mid;
}
```

Or maybe like this: 

```rust
struct MediaBuilder {
  // ...
}

impl MediaBuilder {
  pub fn new() -> Self { /* ... */ }
  pub fn ssrc(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn stream_id(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn ssrc(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
  pub fn track_id(self, ssrc: Ssrc) -> MediaBuilder { /* ... */ }
}

impl SdpApi {
  pub fn add_media_with_builder(
    &mut self,
    kind: MediaKind,
    direction: Direction,
    media_builder: MediaBuilder
  ) -> Mid { /* ... */ }
}
```

algesten:
@evdokimovs Thanks! I'll check the PR.

As for your question, I think we should have that discussion elsehwere. Either raise a new issue or join us on Zulip https://str0m.zulipchat.com

src/change/sdp.rs
```diff
@@ -401,6 +401,37 @@ impl<'a> SdpApi<'a> {
...
+    /// from both the previous [`SdpPendingOffer`] and any newly added changes.
+    ///
+    /// ## Example
+    ///
+    /// ```no_run
+    ///
+    /// # use str0m::media::{Direction, MediaKind};
+    /// # use str0m::Rtc;
+    ///
+    /// let mut rtc = Rtc::new();
```
algesten, line 418: Remove the whitespace lines before and after the import lines, or we get weird gaps in the generated doc:

image

algesten (changes requested):
Looks great! Just a minor doc change and we can land it.

src/change/sdp.rs
evdokimovs: @algesten ,

Fixed!

algesten: Great!

algesten (approved):

PR #372, opened 2023-09-26
Rework Firefox SSRC workaround

k0nserv:
In #346 we attempted to workaround some bugs in Firefox to do with establishing the SSRC mappings for simulcast media. 
We took the approach of always relying on the information communicated via SDP i.e. `a=ssrc-group`, however it turns out 
this information is not reliable and what we should have relied on is the information communicated via the `rid` and `repaired-rid`
RTP headers exclusively.

This PR also fixes an issue where hangover state in `DepacketizingBuffer` would cause it to not emit samples when the SSRC for a given 
(Rid, Mid) pair, or just Mid in the case of non-simulcast, changes.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -337,6 +337,10 @@ impl Session {
                             // We got a change in main SSRC for this stream.
                             let ssrc_from = stream.ssrc();
                             self.streams.change_stream_rx_ssrc(ssrc_from, ssrc);
+                            // When the SSRCs changes the sequence number typically also does, the
+                            // depayloader(if in use) relies on sequence numbers and will not handle a
+                            // large jump corretly, reset it.
+                            media.reset_depayloader(header.payload_type, None);
                         }
                     }
 
@@ -381,6 +385,10 @@ impl Session {
                         // We got a change in main SSRC for this stream.
                         let ssrc_from = stream.ssrc();
                         self.streams.change_stream_rx_ssrc(ssrc_from, ssrc);
+                        // When the SSRCs changes the sequence number typically also does, the
+                        // depayloader(if in use) relies on sequence numbers and will not handle a
+                        // large jump corretly, reset it.
+                        media.reset_depayloader(header.payload_type, Some(rid));
                     }
                 }
                 ssrc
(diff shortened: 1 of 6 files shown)


src/session.rs
```diff
@@ -446,6 +464,8 @@ impl Session {
         let media = self.medias.iter_mut().find(|m| m.mid() == mid).unwrap();
         let stream = self.streams.stream_rx(&ssrc).unwrap();
 
+        // If the header ssrc differs from the main, it's a repair stream.
+        let is_repair = header.ssrc != ssrc;
```
k0nserv, line 468: This should not have been reverted

src/session.rs
```diff
@@ -458,7 +478,6 @@ impl Session {
         };
         let clock_rate = params.spec().clock_rate;
         let pt = params.pt();
-        let is_repair = pt != header.payload_type;
```
k0nserv, line 461: Also this

src/session.rs
```diff
@@ -374,13 +378,27 @@ impl Session {
             let stream_mid_rid = self.streams.stream_rx_by_mid_rid(mid, Some(rid));
 
             let ssrc_main = if is_repair {
-                stream_mid_rid?.ssrc()
+                let Some(stream) = stream_mid_rid else {
+                    return None;
+                };
```
algesten, line 383: Why not keep the `?`

src/session.rs
```diff
@@ -374,13 +378,27 @@ impl Session {
...
+                let main = stream.ssrc();
+
+                if let Some(rtx) = stream.rtx() {
+                    // We got a change in the RTX SSRC
```
algesten, line 387: Is this comment right?

src/session.rs
algesten, line 461: Why does this move up?

algesten (approved):

src/session.rs
k0nserv: The comment is right, but the condition isn't

src/session.rs
k0nserv: Because I didn't need just ssrc anymore so it felt cleaner to check once if the stream existed but it could just be

```
let stream = stream_mid_rid?;
```

src/session.rs
k0nserv: It was part of the revert, it shouldn't have changed

davibe (approved):

PR #373, opened 2023-09-26
chat example: satisfy str0m contract

davibe:


examples/chat.rs
```diff
@@ -124,25 +126,21 @@ fn run(socket: UdpSocket, rx: Receiver<Rtc>) -> Result<(), RtcError> {
...
+                let t = propagate_one(&propagated, client, &mut propagation, &socket);
+                timeout = timeout.min(t);
+                timeouts += 1
+            }
+            if propagation.is_empty() && timeouts == clients.len() {
+                // we have progated everything in the propagation queue and all
+                // clients have returned a timeout
+                break timeout;
+            }
+        };
```
algesten, line 143: This is a bit messy with how that timeout is created every loop, and why is the timeout in this loop at all there is no other timeout that it can end with. Please revise.

algesten (changes requested):

examples/chat.rs
algesten: Right. There's a second loop.

I don't like that second loop. Polling output has nothing to do with propagation, so it shouldn't be inside something called `propagate_one`.

This whole PR is taking the pattern too far away from the documented patterns we have in the README.md etc.

algesten: I would start over, think "minimal changes" and make sure it follows the README.md in terms of structure.

davibe: If I am not wrong, the contract is to 
- send input
- poll all output until you get a timeout

At this point there is no change in state until either time passes (timeout) or there is new input.

propagate_one does exactly that

davibe: the example in the readme does the same, but it's just inverted.

- first it polls everything until there is a timeout
- then reads an input from the network and sends it through

From a state-machine perspective there is no difference.

However in a typical application you "wait" the network and then react, so I find this sequence more natural.

examples/chat.rs
```diff
@@ -107,6 +108,7 @@ fn web_request(request: &Request, addr: SocketAddr, tx: SyncSender<Rtc>) -> Resp
 /// and forwards media data between clients.
 fn run(socket: UdpSocket, rx: Receiver<Rtc>) -> Result<(), RtcError> {
     let mut clients: Vec<Client> = vec![];
+    let mut propagation: VecDeque<Propagated> = VecDeque::new();
```
algesten, line 111: Let's keep calling this `to_propagate` (or alternatively `propagated`).

examples/chat.rs
```diff
@@ -124,26 +126,19 @@ fn run(socket: UdpSocket, rx: Receiver<Rtc>) -> Result<(), RtcError> {
...
-        // We keep propagating client events until all clients respond with a timeout.
-        if to_propagate.len() > timeouts.len() {
-            propagate(&mut clients, to_propagate);
-            // Start over to propagate more client data until all are timeouts.
-            continue;
+        // Poll clients until they return timeout
+        let mut timeout = Instant::now() + Duration::from_millis(100);
+        for client in clients.iter_mut() {
+            let t = poll_until_timeout(client, &mut propagation, &socket);
+            timeout = timeout.min(t);
```
algesten, line 133: Propagation must happen before `poll_until_timeout` or we might get starvation.

algesten (approved):
Some small things

examples/chat.rs
davibe: Can you explain this ? We discover stuff to propagate while polling

algesten: If polling creates more Propagated than we manage to send, we are starving the sending.

The sending side discards stuff, so we need to ensure we are keeping the propagation clear.

davibe: Yes. I expect polling to create more propagated, and propagation to sometimes result in more stuff to "poll and propagate" again. 
In this phase we do not drive time forward and we do not check network input. 
Everything must settle (and it does)

PR #374, opened 2023-09-27
Clean up dynamic SSRC mapping

algesten:


src/session.rs
```diff
@@ -291,144 +291,63 @@ impl Session {
...
-                );
-
-                return Some((mid, ssrc_main));
-            };
+        // The dynamic mapping might have added an entry by now.
+        self.streams.mid_ssrc_rx_by_ssrc_or_rtx(ssrc_header)
+    }
 
-            let is_repair = header.ext_vals.rid_repair.is_some();
+    fn map_dynamic(&mut self, header: &RtpHeader) {
```
davibe, line 307: this could return whether something changed so we can avoid the second attempt

src/session.rs
```diff
@@ -291,144 +291,63 @@ impl Session {
...
-                        // When the SSRCs changes the sequence number typically also does, the
-                        // depayloader(if in use) relies on sequence numbers and will not handle a
-                        // large jump corretly, reset it.
-                        media.reset_depayloader(header.payload_type, Some(rid));
-                    }
-                }
-                ssrc
-            };
+        if let Some(rid) = rid {
+            // Case A - use the rid_repair header to identify RTX.
```
davibe, line 339: where do we use the rid_repair ?

davibe: ah, i see what you mean

PR #375, opened 2023-09-27
fix/vp8 overflow

davibe:
- nack register: don't log RTX activity
- vp8: handle overflow on encoder restarts


src/packet/buffer_rx/vp8_contiguity.rs
```diff
@@ -50,7 +50,10 @@ impl Vp8Contiguity {
 
             // Frame on layer 0: always emit and report discontinuity if not subsequent
             let emit = true;
-            let contiguous = tl0_picture_id == last_tl0_picture_id + 1;
+            // note: we use overflowing add because the encoder can restart if the
+            // camera is closed / reopened and here we only care about
+            // contiguity
+            let contiguous = tl0_picture_id == last_tl0_picture_id.overflowing_add(1).0;
```
k0nserv, line 56: Why not use `wrapping_add` if we aren't looking at the second value in the tuple?

davibe: you are right, I think that's what i meant to do in reality

davibe: :(

davibe: it matters when we cross the 0 again, we'd detect it wrong!

PR #377, opened 2023-09-28
Event has Box<RawPacket> to reduce stack size

algesten:


src/media/event.rs
```diff
@@ -122,7 +122,7 @@ pub struct MediaData {
 
     /// RTP header extensions for this media data. This is taken from the
     /// first RTP header.
-    pub ext_vals: ExtensionValues,
```
k0nserv, line 125: This is "only" 184 bytes large and boxing it adds another allocation to our hottest path(parsing incoming RTP packets). Is it worth it?

davibe (approved):

src/media/event.rs
algesten: Agree. Not worth it. I remove that.

PR #379, opened 2023-09-29
add DatagramSend into Vec

davibe:


src/io/mod.rs
```diff
@@ -71,6 +71,12 @@ impl From<Vec<u8>> for DatagramSend {
     }
 }
 
+impl Into<Vec<u8>> for DatagramSend {
```
k0nserv, line 74: Instead of implementing `Into<Vec<u8>> for DatagramSend`, implement `From<DatagramSend> for Vec<u8>`, quoting from the docs for `Into`

> One should avoid implementing Into and implement From instead. Implementing From automatically provides one with an implementation of Into thanks to the blanket implementation in the standard library.

algesten (approved):

PR #380, opened 2023-09-30
Add cargo-deny check to CI

algesten:
This config protects from:

  * duplicate crates
  * copy-left licenses in the dependency tree
  * RUSTSEC advisories

Note we need to check in Cargo.lock for this to work.


algesten:
Thanks to @emilk for the help in ureq https://github.com/algesten/ureq/pull/661

davibe (approved):

PR #381, opened 2023-10-04
send stats performances

davibe:
Based on this capture ValueHistory contributes to the heaviness of str0m::session::Session::poll_packet

Screenshot 2023-10-04 at 19 12 55

Tested locally these 2 alternatives

```
+        for _ in 0..1_000_000 {
+            h.push(now, 22);
+
+            // 1
+            h.sum_since(now - Duration::from_secs(1));
+            // 2
+            h.sum();
+
+            now += inc;
+        }
```

This goes from 2s to 0.05s

### DIFF
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -493,8 +493,7 @@ impl StreamTx {
     }
 
     fn poll_packet_resend(&mut self, now: Instant) -> Option<NextPacket<'_>> {
-        let from = now.checked_sub(Duration::from_secs(1)).unwrap_or(now);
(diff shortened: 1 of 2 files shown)


algesten:
Did you run a test with proper profiling?

These kinds of microbenchmarks tend to be wildly wrong because the compiler optimizes away something that it won't be able to optimize away in the real usage.

davibe:
Screenshot 2023-10-16 at 08 33 54


in the same capture the sum_since / sum is gone completely

algesten (approved):
Lovely jubbly, bottle of bubbly!

PR #384, opened 2023-10-11
Up max blank padding size to 240

### LINKED ISSUE #266: Consider increasing blank padding packet size
We [set the maximum blank padding size to 224](https://github.com/algesten/str0m/blob/6d317bb6084351215c15cf5e652748cf056e9218/src/rtp/mod.rs#L37-L38). This decision was based on libWebRTC, but they have now [upped this limit](https://webrtc-review.googlesource.com/c/src/+/305642) and we should be able to follow suite.

algesten:
Modulus 16 for SRTP.

Close #266

### DIFF
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::fmt;
 use std::time::{Duration, Instant};
 
+use crate::rtp_::MAX_BLANK_PADDING_PAYLOAD_SIZE;
 use crate::rtp_::{Bitrate, DataSize, Mid};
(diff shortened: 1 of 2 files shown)


k0nserv:
From the [changeset](https://webrtc-review.googlesource.com/c/src/+/305642) in libWebRTC
> SRTP encryption does not need to be taken into account since none of
the cipher suites used by WebRTC require padding:

From the str0m source I was under the impression that `SRTP_AES128_CM_SHA1_80` did require padding, if that's not the case str0m could probably be optimised a bit by making the current padding optional(based on SRTP context)

k0nserv (approved):

PR #385, opened 2023-10-12
sdp: expose MediaType and MediaLine to crate users

david-flok:
This enables users to easily inspect str0m's SDP negotiation outcome for, e.g., codec preferences.

I appreciate this is probably not currently to your standards for merging e.g. regarding docs or policies about type aliases or something but I thought I'd get the conversation going.

The use case is basically one where a client makes an SDP offer, str0m generates an SDP answer, and the application would like to know what str0m is negotiating and which codecs are expected in what preferred order.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -575,6 +575,12 @@ use sctp::{RtcSctp, SctpEvent};
 
 mod sdp;
 
+/// type of media
+pub type MediaType = crate::sdp::MediaType;
+
+/// SDP media line
+pub type MediaLine = crate::sdp::MediaLine;
+
 pub mod format;
 use format::CodecConfig;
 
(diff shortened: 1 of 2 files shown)


algesten:
Hi @david-flok!

In my mind, SDP is a transfer mechanism (and a terrible such at that!). Since each SDP dance is a negotiation, that can be rejected/fail, I'm wondering in what situation we would get good insights into the session from that? 

To me the API that would be more appropriate is str0m's `Media` API, where ready negotiated codecs and configs are all available. However there might be some need you have that I'm not considering. Hence I'm asking for a bit more context.

david-flok:
Aha, no, I think you are right that `Media` actually satisfies my use case. I was expecting to extract the information as soon as it was available rather than after I get `MediaAdded` events but the `Media` API is probably better, anyway. Thanks and sorry to bother you!

algesten:
No problem!

PR #388, opened 2023-10-13
Dev/high first rtt

### LINKED ISSUE #386: First egress rtt value is wrong.
First egress rtt value is wrong rather than None or correct.

Here is the first 2 instances when reporting every 10s and the testcase is 25s long

```
egress_stats_l: [
    MediaEgressStats {
        mid: Mid(IER),
        rid: None,
        bytes: 196320,
        packets: 2454,
        firs: 0,
        plis: 0,
        nacks: 0,
        rtt: Some(
            3605042.0,
        ),
        loss: None,
        timestamp: Instant {
            tv_sec: 19007,
            tv_nsec: 122291193,
        },
    },
    MediaEgressStats {
        mid: Mid(IER),
        rid: None,
        bytes: 395760,
        packets: 4947,
        firs: 0,
        plis: 0,
        nacks: 0,
        rtt: Some(
            0.0,
        ),
        loss: Some(
            0.0,
        ),
        timestamp: Instant {
            tv_sec: 19017,
            tv_nsec: 122291193,
        },
    },
]
```

Testcase can be found here:

https://github.com/pthatcher/str0m/commit/1b068f4b7dfeec8b4d5927aef3772ecf70956c1b

xnorpx:
Fixes #386 

- Add tests for stats
- Check RTT value
- Return None and set RTT to None instead of large number.

### DIFF
--- a/src/util/mod.rs
+++ b/src/util/mod.rs
@@ -61,6 +61,11 @@ pub(crate) fn calculate_rtt_ms(ntp_time: Duration, delay: u32, last_report: u32)
     // [32 bit seconds].[32 bit fractions]
     //         [16 bit].[16 bit]
 
+    // As per RFC delay is 0 in case no SR packet has been received yet.
+    if delay == 0 {
+        return None
+    }
+
     let now_secs = ntp_time.as_secs();
     let now_fract_ns = ntp_time.subsec_nanos() as u64;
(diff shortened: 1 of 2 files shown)


algesten (approved):
This looks good to me. @davibe since you're closer to the stats code, what do you think?

davibe (approved):
LGTM

PR #389, opened 2023-10-17
Pass Candidate Protocol arounds along with Address(es) so that routing to the correct ICE route can be done cleanly.

efer-ms:
With this change it is possible to have UDP and TCP candidates that use the same port number, as well as, the ability to create non-UDP host candidates manually (via the Candidate::host function.

Synopsis of the changes:
- proto on Candidate has been change from a String to an enum CandidateProtocol
- Functions that create Candidates now take the CandidateProtocol, rather than sometimes simply using "udp"
- do_prio() in Candidate now takes into account protocol as well as candidate type. This matches libwebtc values.
- Candidates are now considered redundant only if the old rules that were applied are still true, and the protos match
- Candidates being placed in pairs, must have the same protocol
- StunRequest, NominatedSend, DiscoveredRecv, Receive, Transmit all now includes protocol the Candidate used in addition to the address(es).

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,7 +58,7 @@ let mut rtc = Rtc::new();
 
 //  Add some ICE candidate such as a locally bound UDP port.
 let addr = "1.2.3.4:5000".parse().unwrap();
-let candidate = Candidate::host(addr).unwrap();
+let candidate = Candidate::host(addr, CandidateProtocol::Udp).unwrap();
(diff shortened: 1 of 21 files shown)


algesten:
Wow! This looks amazing! I'll review tomorrow.

Thanks!

algesten:
This is landed in #391 with some additional tweaks. Thanks!!!

PR #393, opened 2023-11-01
Add support for 2-byte header extensions

pthatcher:
While in there, also refactor ExtensionMap.iter() to allow for iterating over all extensions.


algesten (approved):
I never seen the 2-byte in the wild, but the PR looks good. Some nice cleanups too.

Thanks!

PR #394, opened 2023-11-01
Fix the RID when receiving RTX packets

pthatcher:


algesten (approved):
Cool.

I was wondering why this doesn't bite our own server, but when using the Media API, the rid for the layer comes from another place. So this is a concern for RTP API.

PR #395, opened 2023-11-01
Send unpaced media as padding.

pthatcher:
This prevents a hot loop when you enable RTX for audio and BWE/pacing at the same time without having to know to call .set_unpaced(false) for the audio.

### DIFF
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -431,7 +431,7 @@ impl LeakyBucketPacer {
             .filter_map(|qs| (qs.snapshot.first_unsent.map(|t| (t, qs))))
             .min_by_key(|(t, _)| *t);
 
-        // Unpaced packets (such as audio buy default) are sent immediately.
+        // Unpaced packets (such as audio by default) are sent immediately.
         if let Some((queued_at, qs)) = unpaced {
             return Some((queued_at, Some(qs)));
         }
--- a/src/packet/pacer.rs
+++ b/src/packet/pacer.rs
@@ -431,30 +431,32 @@ impl LeakyBucketPacer {
             .filter_map(|qs| (qs.snapshot.first_unsent.map(|t| (t, qs))))
             .min_by_key(|(t, _)| *t);
 
-        // Unpaced packets (such as audio by default) are sent immediately.
+        // Unpaced packets (such as audio buy default) are sent immediately.
         if let Some((queued_at, qs)) = unpaced {
             return Some((queued_at, Some(qs)));
         }
 
         // "Media queue" as opposed to a "padding queue".
-        let non_empty_media_queue = {
-            let non_empty_media_queues = self.queue_states.iter().filter(|q| {
-                let is_none_empty_media_queue = match q.snapshot.priority {
+        let non_empty_queue = {
+            let non_empty_queues = self.queue_states.iter().filter(|q| {
+                let is_none_empty_queue = match q.snapshot.priority {
                     QueuePriority::Media => true,
-                    QueuePriority::Padding => false,
-                    QueuePriority::Empty => false,
+                    // TODO: Try changing this to false.  Currently, if it's false, the test_realistic test fails.
+                    QueuePriority::Padding => true,
+                    // TODO: Try removing "q.snapshot.packet_count > 0" and using "QueuePriority::Empty => false" instead
+                    QueuePriority::Empty => true,
                 };
-                q.snapshot.packet_count > 0 && is_none_empty_media_queue
+                q.snapshot.packet_count > 0 && is_none_empty_queue
             });
 
-            // Send on the non-empty media queue with the lowest priority that, was least recently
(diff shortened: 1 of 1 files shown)


src/packet/pacer.rs
```diff
@@ -431,25 +431,32 @@ impl LeakyBucketPacer {
             .filter_map(|qs| (qs.snapshot.first_unsent.map(|t| (t, qs))))
             .min_by_key(|(t, _)| *t);
 
-        // Unpaced packets (such as audio by default) are sent immediately.
+        // Unpaced packets (such as audio buy default) are sent immediately.
```
xnorpx, line 434: typo?

pthatcher: Well that is strange.   I didn't even try and change that line.  Oh well.  I reverted that line.

src/packet/pacer.rs
```diff
@@ -436,20 +436,27 @@ impl LeakyBucketPacer {
...
+            let non_empty_queues = self.queue_states.iter().filter(|q| {
+                let is_none_empty_queue = match q.snapshot.priority {
+                    QueuePriority::Media => true,
+                    // TODO: Try changing this to false.  Currently, if it's false, the test_realistic test fails.
+                    QueuePriority::Padding => true,
+                    // TODO: Try removing "q.snapshot.packet_count > 0" and using "QueuePriority::Empty => false" instead
+                    QueuePriority::Empty => true,
+                };
+                q.snapshot.packet_count > 0 && is_none_empty_queue
+            });
```
algesten, line 450: At the moment this change looks a bit like it does nothing. We remove `!q.unpaced`, but then, if there were any content in an unpaced queue, we would have returned earlier. The `is_none_empty_queue` bool is always true, should we maybe consider this PR a draft given the comment?

pthatcher: The key to the PR is that "&& !q.unpaced" is removed.  That's what fixes that hot loop that's triggered by enabling audio RTX with BWE.  But Hugo suggested I also filter out queues with QueuePriority::Padding.  But that caused the BWE test called test_realistic to fail.

I would be happy to remove the QueuePriority stuff and just reduce this PR down to removing "&& !q.unpaced", because that's what's important to the fix, and perhaps leave a TODO to look into filtering by QueuePriority.

algesten: > This prevents a hot loop when you enable RTX for audio and BWE/pacing at the same time without having to know to call .set_unpaced(false) for the audio.

Is the starting point here right? Is it really meaning to change `!q.unpaced`? 

* RTX is required for BWE 
* `.set_unpaced(false)` is required for BWE

Should the default `unpaced` flag logic be different for this scenario?

We of course want no hot looping, but is the solve here the right one?

k0nserv: The current behaviour is:

1. We only do BWE calculations on video media
2. We only pace video media
3. We only do RTX on video media



> At the moment this change looks a bit like it does nothing. We remove !q.unpaced, but then, if there were any content in an unpaced queue, we would have returned earlier. The is_none_empty_queue bool is always true, should we maybe consider this PR a draft given the comment?

The root cause that @pthatcher identified is that we don't set a reasonable fake value for `first_unsent` when an audio queue only contains padding. We could fix it that way, but that would introduce a new problem(we would prioritize queued padding on an audio queue over queued media on a video queue). This is what we need to remove the `!q.unpaced` condition.

The priority stuff is not correct I think, just remove that, we already have code that will select the queue with the highest priority.

lolgesten: > 1. We only do BWE calculations on video media
> 2. We only pace video media
> 3. We only do RTX on video media

It is possible to enable RTX for audio (point 3) and it's possible to turn on pacing for audio (point 2). The only point I'm unsure about is 1 – you mean we don't produce queue stats for audio?

algesten: > 1. We only do BWE calculations on video media
> 2. We only pace video media
> 3. We only do RTX on video media

It is possible to enable RTX for audio (point 3) and it's possible to turn on pacing for audio (point 2). The only point I'm unsure about is 1 – you mean we don't produce queue stats for audio?

k0nserv: > The only point I'm unsure about is 1 – you mean we don't produce queue stats for audio?

No, I was under the impression that we didn't enable TWCC for audio and it therefore wasn't counted, seems we do and I was wrong. It might have been that browsers themselves didn't use to negotiate TWCC for audio

k0nserv: > The real question is: if audio RTX and audio TWCC are enabled, can audio be unpaced? If so, then it shouldn't result in a hot loop. If not, then we need a way of making sure audio is paced when audio TWCC is enabled.

Yes it can be, but in theory it shouldn't be. I think the bandwidth of audio might be low enough that it does not matter. The pacing is mostly about avoiding overloading the link when sending large keyframes after all

PR #398, opened 2023-11-03
Serialize/Deserialize Candidate

gak:
* When used with serde_json, it will generate:
```json
{
  "candidate": "candidate:12044049749558888150 1 udp 2130706175 1.2.3.4 9876 typ host ufrag ufrag",
  "sdpMid": null,
  "sdpMLineIndex": 0,
  "usernameFragment": "ufrag"
}
```
* It will ignore `sdpMid` and `sdpMLineIndex` and set them to null and 0 respectively, because 1) `Candidate` does not store these values, and 2) ICE candidates require them to [not be null](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/sdpMLineIndex#value)).
* Also add `pub fn from_sdp_string(s: &str)` and `pub fn to_sdp_string(&self) -> String`
* Display/from_str now doesn't return an sdp m-line.

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -531,18 +531,20 @@ impl<'de> Deserialize<'de> for Candidate {
     {
         #[derive(Deserialize)]
         #[serde(rename_all = "camelCase")]
-        struct CandidateJson {
+        struct CandidateDeserialized {
             candidate: String,
             username_fragment: Option<String>,
         }
 
-        let CandidateJson {
+        let CandidateDeserialized {
             candidate,
             username_fragment,
-        } = CandidateJson::deserialize(deserializer)?;
+        } = CandidateDeserialized::deserialize(deserializer)?;
+
         let (mut c, _) = trickle_candidate_parser()
             .parse(candidate.as_str())
             .map_err(|e| serde::de::Error::custom(e))?;
+
         if let Some(ufrag) = username_fragment {
             c.set_ufrag(&ufrag);
         }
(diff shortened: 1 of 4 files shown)


algesten:
@gak Thank you! Looks good! I have a couple of thoughts around naming and package structure.

1. "ice string" doesn't sound right. In the RTCPeerConnection API, the serialized form of the entire container object is called ["candidate info"](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate#candidateinfo). This could be a good name for that internal structure in the deserialization, and be used in documentation. However the actual string form is [from SDP](https://datatracker.ietf.org/doc/html/rfc5245#section-15.1) and is called a "candidate attribute" (or maybe ["candidate SDP attribute"](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate#candidate)).  I think we should echo the existing API and rename the "ice string" to `to_sdp_string` and `from_sdp_string`.
2. Regarding the `a=` prefixing, I think we can just leave not bring that out like this. That is an internal detail to the SDP parser and serializer. 
3. Leading us to the parser. I want the SDP combine parser to stay in one place. Which leads us to two actions:
  a. Refactor in the `parser.rs` so that you get the non-`a=` part of `fn candidate()` into a separate function.
  b. Expose this function in `sdp/mod.rs` and give it some non-combine signature (like working off `&str`). 
4. Finally `"sdpMid": ""` I think is incorrect. They are both marked as optional in the javascript API, but they [cannot be null at the same time](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate#exceptions). We set non-null for `sdpMidLineIndex` and null for `sdpMid`.

Let me know if anything is wrong/unclear or whatever.

src/sdp/parser.rs
```diff
@@ -201,16 +202,22 @@ where
     ))
 }
 
-#[doc(hidden)]
-pub fn candidate<Input>() -> impl Parser<Input, Output = Candidate>
+/// Parse a candidate string into a [Candidate].
+///
+/// Does not parse an `a=` prefix.
+pub fn parse_candidate_attribute(s: &str) -> Result<Candidate, SdpError> {
```
algesten, line 208: This name is a bit strange. The `a=` bit is literally the SDP attribute, so technically this function parses everything BUT the attribute.

I think we should turn it around so `candidate` and `candidate_attribute` change name with each other. That also drops the "attribute" bit from the module export which I think is correct.

src/sdp/parser.rs
```diff
@@ -232,55 +239,71 @@ where
...
+where
+    Input: Stream<Token = char>,
+    Input::Error: ParseError<Input::Token, Input::Range, Input::Position>,
+{
+    // a=candidate:1 1 udp 2113929471 203.0.113.100 10100 typ host
+    // a=candidate:1 2 udp 2113929470 203.0.113.100 10101 typ host
+    // a=candidate:1 1 udp 1845494015 198.51.100.100 11100 typ srflx raddr 203.0.113.100 rport 10100
+    // a=candidate:1 1 udp 255 192.0.2.100 12100 typ relay raddr 198.51.100.100 rport 11100
+
+    (string("a="), candidate_attribute()).map(|(_, c)| c)
```
algesten, line 306: I believe you can still reuse the `attribute_line` parser function outermost here.

algesten (changes requested):
Looks great! A couple of tiny name changes and we ship it.

src/sdp/parser.rs
gak: You beat me to it. I was going to reconsider these names but didn't get around to it yet.

src/sdp/parser.rs
gak: Unfortunately `attribute_line` still needs the name of the attribute which is already handled.

src/sdp/parser.rs
gak: fn's renamed. Hopefully it makes sense now.

src/ice/candidate.rs
```diff
@@ -481,13 +487,6 @@ fn is_valid_ip(ip: IpAddr) -> bool {
     }
 }
 
-// TODO: maybe a bit strange this is used for SDP serializing?
```
gak, line 457: I moved this function closer to the struct, and removed the comment as it isn't being used by sdp now.

algesten (approved):
Looks really nice. Let's merge!

algesten:
Thanks for doing this!

gak:
Happy to contribute! 👍

morajabi:
Now I can remove codes from `webrtc::ice::candidate` 🫡

PR #399, opened 2023-11-13
Fix SDP for inactive `rid`s

logist322:
This PR adds ability to receive offer with simulcast including some inactive `rid`s ([marked with "~" prefix](https://datatracker.ietf.org/doc/html/rfc8853#name-simulcast-attribute)) and to generate correct answer.

### DIFF
--- a/src/sdp/data.rs
+++ b/src/sdp/data.rs
@@ -667,14 +667,16 @@ impl From<Direction> for MediaAttribute {
 pub struct RestrictionId(pub String, pub bool);
 
 impl RestrictionId {
-    pub fn new(v: String) -> Self {
-        let rid = v.trim_start_matches("~");  
-        
-        RestrictionId(rid.to_string(), v == rid)
+    pub fn new(rid: String, active: bool) -> Self {
+        RestrictionId(rid, active)
+    }
+
+    pub fn new_active(rid: String) -> Self {
+        RestrictionId(rid, true)
     }
 
     pub fn to_sdp(&self) -> String {
-        format!("{}{}", if self.1 {""} else {"~"}, self.0)
+        format!("{}{}", if self.1 { "" } else { "~" }, self.0)
     }
 }
 
--- a/src/sdp/data.rs
+++ b/src/sdp/data.rs
@@ -664,17 +664,23 @@ impl From<Direction> for MediaAttribute {
 ///
 /// Defined in https://tools.ietf.org/html/draft-ietf-avtext-rid-09
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


algesten:
Hi! Welcome to str0m!

Looking good. I don't actually know the spec that underpins this change, but I will look into it.

xnorpx:
~ is libwebrtc specific?

lolgesten:
I'm curious too. Can @pthatcher shed some light?

The `~` prefix on the RID seems to indicate that the stream might be paused using "CCM TMMBR and TMMBN" RTCP messages. Following the links to https://datatracker.ietf.org/doc/html/rfc5104 – which largely seems to concern H271.

@logist322 could you maybe describe how you encountered these `~`?

logist322:
@algesten Sure! I have faced it after adding a video transceiver with several  `RTCRtpEncodingParameters` some of them have `active: false`. That's the way how Chrome and Safari works (@xnorpx in pure libWebRTC too). 
You can reproduce it with this simple code in Chrome or Safari:
```
(async () => {
    let pc = new RTCPeerConnection();
    await pc.addTransceiver("video", {
        direction: "sendrecv",
        sendEncodings: [{
            rid: "h",
            active: true
        }, {
            rid: "m",
            active: false
        }, {
            rid: "l",
            active: false
        }]
    });
    let offer = await pc.createOffer();
    console.log(offer.sdp);
})();
```
You should see `a=simulcast:send h;~m;~l` line in the printed `sdp`.

pthatcher:
That's accurate.  It just means that the stream starts as if an TMMBR pause
message had been received at the same time.  Nothing should be sent until a
TMMBR unpause is sent.

Honestly, I would do all of this over a data channel and not use RTCP, but
maybe someone would want to use RTCP for this.  But I also don't think
anyone should use SDP :).

logist322:
> It sounds like `~` means a RID starts paused, but that might change later _without SDP negotiation_ (whether there are RTCP ways for us to discover when something becomes unpaused, I don't know). We already have a paused state on the receive streams here: https://github.com/algesten/str0m/blob/main/src/streams/receive.rs#L95-L99

@algesten You are absolutely right! We do use it for changing `RTCRtpEncodingParameters` dynamically _without SDP negotiation_.  

> Does it sound right that this ~ state should have an effect for a newly created StreamRx?

I don't think so. Especially, if a newly created `StreamRx`s already have `paused: true`.

algesten:
Alright. Let's merge it. At this point the `~` has no implication for str0m. This PR will fix so it doesn't fall over.

algesten (approved):

PR #400, opened 2023-11-14
MediaTime: fix negative value into Duration bug

david-flok:
`as u64` of `i64` is dangerous. `Duration` is non-negative but `MediaTime` is not necessarily. This fixes what is effectively an integer underflow.

### DIFF
--- a/src/rtp/mtime.rs
+++ b/src/rtp/mtime.rs
@@ -1,5 +1,3 @@
-#![allow(missing_docs)]
-
 use std::cmp::Ordering;
 use std::fmt::Display;
 use std::num::{NonZeroU32, TryFromIntError};
@@ -10,8 +8,9 @@ use std::time::Instant;
 
 use serde::{Deserialize, Serialize};
 
-use crate::util::InstantExt;
-
+/// Media timeline frequency as represented by a non-zero unsigned integer.
+///
+///
 #[derive(Debug, Clone, Copy, Serialize)]
 pub struct Frequency(NonZeroU32);
 
@@ -37,7 +36,7 @@ impl Frequency {
     /// Seconds in a second.
     pub const SECONDS: Frequency = Self::make(1);
 
-    /// Private unconditional non-zero constructor for use with constants
+    /// Private unconditional non-zero constructor for use with constants.
     const fn make(v: u32) -> Frequency {
         match NonZeroU32::new(v) {
             Some(v) => Self(v),
@@ -45,18 +44,22 @@ impl Frequency {
         }
     }
 
+    /// Any non-zero u32 is a valid media timeline frequency.
     pub const fn from_nonzero(v: NonZeroU32) -> Self {
         Self(v)
     }
 
+    /// Every u32 is a valid media timeline frequency except zero.
     pub fn new(v: u32) -> Option<Self> {
         NonZeroU32::new(v).map(Self)
     }
 
+    /// The frequency as a u32.
     pub const fn get(&self) -> u32 {
         self.0.get()
     }
 
+    /// The frequency as a [`std::num::NonZeroU32`] i.e. including a positivity proof.
     pub const fn nonzero(&self) -> NonZeroU32 {
         self.0
     }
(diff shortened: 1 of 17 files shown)


david-flok:
Actually, I'm not sure that `MediaTime` should have `From<MediaTime>` for `Duration` at all -- this conversion is not lossless. I think I am in favor of an explicit conversion instead or `TryFrom` or something.

lolgesten:
I don't feel strongly one way or the other.

Have you experienced a bug due to this conversion?

david-flok:
Yes. I had two `MediaTime` values that I expected to almost always be ordered tsA < tsB and so I used `Duration::max(Duration::ZERO, (tsB - tsA).into())` but this results in underflow. Instead, I must use `MediaTime::max(MediaTime::ZERO, tsB - tsA).into()` so the timestamp doesn't get corrupted. I would expect `_.into()` and `max` to commute.

I will prepare a proposed `MediaTime` patch to catch this issue statically.

algesten:
Sounds good!

david-flok:
@algesten I've revised the patch. While I was there, I also fixed some defects regarding handling of negative and zero denominators. Those cases are now statically disallowed. This should eliminate a possible DoS vulnerability resulting from SDP parsing of streams with 0 frequency resulting in division by zero panics.

You can, of course, cherry-pick the commit(s) you want or ask me to prepare a reduced changeset but I would very much like this basic data type to be as safe as possible.

lolgesten:
Thanks!

I'm in two minds about this `Frequency` newtype. On the one hand, it provides some really nice guarantees in the SDP parsing, on the other, it is Another Thing a user of the library need to get their head around. It doesn't look like `Frequency` is visible in that many places in the public API though. Could you maybe try making it `pub(crate)` and see what compilation errors you get?

For now we could potentially keep it internal and decide later if we want to have it public.

david-flok:
Unfortunately, it is indeed rather heavyweight compared to just plugging in literals but I have tried to make it as straightforward as possible to use with frequency-specific constructors and constants.

As far as I can see, the design space for this division by zero restriction is:

1. Do nothing and expect the user to ensure non-zero frequency.
2. Check at construction time and panic.
3. Check at construction time and return a `Result`/`Option`.
4. Check at construction time and map 0 to 1 silently.
5. Use `NonZeroU32` directly in interfaces.
6. Wrap `NonZeroU32` with a newtype (`Frequency`).

Perhaps you'd be happier with 4?

I have added `AddAssign` and `SubAssign` implementations because doing `t = t + x` instead of `t += x` was bothering me. I've also rebased onto `main` as there were some conflicts.

david-flok:
Another option would be to do both the `Frequency` newtype and the 0 to 1 mapping with the larger type on input and the smaller type on output (Postel style).

The `Frequency` type only shows up in signatures for things like `clock_rate` and the denominator of these `MediaTime` terms -- it's there entirely to guarantee that there are only 2^32 - 1 values.

algesten:
@xnorpx you had thoughts on MediaTime (passing it to `write_rtp`, rather than just the RTP u32 value).

My idea for API ergonomics here is that

1. Get the `Format` for whatever you're sending (`Rtc -> CodecConfig -> PayloadParam -> CodecSpec -> Format`)
2. Grab the `Frequency` from that `Format`.
3. Make a `MediaTime` using the `Frequency` (this could potentially be expanded with more helpers)

Since you need to know the negotiated `PT`, you already need to dig out the `PayloadParam`.

xnorpx:
It feels a little clunky to need to create a MediaTime everytime I call write. Pass in Pt and RtpTimestamp and MediaTime could be created inside. But it could just be the way I use it.

We probably need to add 8000 and 16000 at some point to the frequency struct, but I like having the new type.

src/rtp/mtime.rs
```diff
@@ -1,36 +1,127 @@
 #![allow(missing_docs)]
 
 use std::cmp::Ordering;
-use std::ops::{Add, Sub};
+use std::fmt::Display;
```
algesten, line 4: This is not your problem, really, but it's expanding with the increased amount of code in this file.

Could you remove the `#![allow(missing_docs)]` at the top and document the new and old code?

Also, In the `Frequency` doc on `struct`, and `MediaTime` struct, I'd like to see some code examples how to get the `Frequency` for a certain `Format` from the chain `Rtc -> CodecConfig -> PayloadParam -> CodecSpec -> Format -> Frequency` and how to subsequenty use that to construct a `MediaTime`. You can duplicate the code example.

algesten (changes requested):
Ok, let's land this. I want some documentation is all.

algesten:
@david-flok also, note #338 which I want to land before this, which will clean up in `MediaTime` a bit.

algesten:
Hi @david-flok, I know I've created some more work in this PR.

Would you prefer if I took the PR to completion, or do you want to finish it?

david-flok:
Hi @algesten, no worries. I'll get polish this now, sorry for the delay.

src/rtp/mtime.rs
david-flok: I'm not sure what you have in mind for this example so I've not guessed, sorry.

david-flok:
Ok, I've rebased and resolved the conflicts with #338. I also did some documentation but I haven't provided the example(s) you wanted, sorry.

algesten (approved):
From my private user

algesten:
@david-flok I added some code example: https://github.com/algesten/str0m/commit/a319da9d0a6b776b979893fb31ad1a184eee8ac8

Issue #261, opened 2023-08-07
Dtls cannot establish with libdatachannel


giangndm-bluesea:
When we work with OBS broadcast (which uses libdatachannel internally), DTLS connection cannot be established.

```
[2023-08-07T10:12:18.325Z INFO  str0m::ice::agent] Add local candidate: Candidate(host=192.168.66.223:64225 prio=2130706175)
[2023-08-07T10:12:18.325Z TRACE str0m::ice::agent] Calculated local preference: 65534
[2023-08-07T10:12:18.327Z INFO  str0m::ice::agent] Set remote credentials: IceCreds { ufrag: "p5IV", pass: "MeL2sgkoqVHqORLqUmVP9W" }
[2023-08-07T10:12:18.327Z INFO  str0m::ice::agent] Add remote candidate: Candidate(host=192.168.66.223:53735 prio=2122317567)
[2023-08-07T10:12:18.327Z TRACE str0m::ice::agent] Form pair local: Candidate(host=192.168.66.223:64225 prio=2130706175) remote: Candidate(host=192.168.66.223:53735 prio=2122317567)
[2023-08-07T10:12:18.327Z DEBUG str0m::ice::agent] Add new pair CandidatePair(0-0 prio=9115284546252701182 state=Waiting attempts=0 unanswered=0 remote=0 last=None nom=None)
[2023-08-07T10:12:18.327Z INFO  str0m::ice::agent] Add remote candidate: Candidate(host=10.243.178.47:53735 prio=2122317823)
[2023-08-07T10:12:18.327Z TRACE str0m::ice::agent] Form pair local: Candidate(host=192.168.66.223:64225 prio=2130706175) remote: Candidate(host=10.243.178.47:53735 prio=2122317823)
[2023-08-07T10:12:18.327Z DEBUG str0m::ice::agent] Add new pair CandidatePair(0-1 prio=9115285645764328958 state=Waiting attempts=0 unanswered=0 remote=0 last=None nom=None)
[2023-08-07T10:12:18.327Z INFO  str0m] DTLS setup is: false
[2023-08-07T10:12:18.327Z DEBUG str0m::session] Enable TWCC feedback
[2023-08-07T10:12:18.327Z INFO  str0m::media] Set Mid(0) extension map: Extensions()
[...]

PR #402, opened 2023-11-20
Set DTLS cert issuer/subject

algesten:


k0nserv (approved):
Do we not need to set more than `CN`? The example from @jellevd246 in #261 did after all

lolgesten:
@k0nserv it doesn't look like libWebRTC does. https://webrtc.googlesource.com/src/+/1568f1b1330f94494197696fe235094e6293b258/rtc_base/openssl_certificate.cc#58

davibe (approved):

Issue #404, opened 2023-11-24
No way to choose `VP9 SVC layer` 


logist322:
I'm trying to use **str0m** with **VP9 SVC** and I have found that there is no opportunity to choose **SVC layers** from the emitted `MediaData`. 

I have some propositions how to solve it:
- add a scheme of `MediaData::data` in `Vp9CodecExtra` for manual constructing preferred **SVC layers**,
- add some methods on `MediaData` that configure `MediaData::data` for preferred **SVC layer**,
- emit `MediaData` not for the all **SVC layers**,  but for the very **SVC layer**.

Let`s choose what to do and I will implement it!

davibe:
I think the current approach would be the 1st one you mentioned. Unless you see issues with it

logist322:
@davibe Thanks for your quick answer!

algesten:
@davibe do we have any example of how this would look?

logist322:
@algesten If you want you can take a look at [_my drafts_](https://github.com/algesten/str0m/compare/main...logist322:str0m:svc-mapping?diff=unified). Of course, it is minimal dirty prototype, but it has been already working with our media server.

logist322:
@algesten The max size is 9 (9 layers for S3T3/L3T3 cases).

davibe:
@logist322 this can be closed, right ?

PR #407, opened 2023-11-30
VP9 SVC layers mapping

logist322:
This PR implements solution discussed in the issue #404.

### DIFF
--- a/src/packet/vp9.rs
+++ b/src/packet/vp9.rs
@@ -13,6 +13,89 @@ const MAX_VP9REF_PICS: usize = 3;
 /// InitialPictureIDFn is a function that returns random initial picture ID.
 pub type InitialPictureIDFn = Arc<dyn (Fn() -> u16) + Send + Sync + UnwindSafe + RefUnwindSafe>;
 
+/// Vp9 information describing the depacketized/packetized data.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct Vp9CodecExtra {
+    /// Map of the SVC layers.
+    ///
+    /// Index of each element corresponds to `spatial layer` index.
+    ///  
+    /// Each element represents the end (not including) in [`MediaData::data`]
+    /// for `spatial layer` it belongs.
+    ///
+    /// The all `spatial layer`s (up to 3) in [`MediaData::data`] are sorted by
+    /// `spatial layer` index from lower to greater. So, the beginning of some
+    /// `spatial layer` is the end (including) of the previous `spatial layer`
+    /// (for the `spatial layer` with index 0 the beginning is 0).
+    ///
+    /// ```text
+    /// [MediaData::data]:
+    /// +-----+----------------------+---------------------------+
+    /// | SL0 |         SL1          |            SL2            |
+    /// +-----+----------------------+---------------------------+
+    /// |     |                      |                           |
+    /// V     V                      V                           V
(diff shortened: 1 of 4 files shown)


src/packet/vp9.rs
```diff
@@ -13,6 +13,83 @@ const MAX_VP9REF_PICS: usize = 3;
...
+    ///         if let (true, Some(end_of_target_layer)) = (
+    ///             // Check whether we need this `temporal layer`.
+    ///             tid <= target_temporal,
+    ///             // Check whether we have information about the right
+    ///             // border of data for the target `spatial layer`.
+    ///             vp9_extra.layers_scheme[target_spatial as usize],
+    ///         ) {
+    ///             // Cut off unwanted data of `spatial layer`s with spatial
+    ///             // indeces higher than `target_spatial`.
+    ///             media_data.data.drain(end_of_target_layer..);
```
davibe, line 69: This is a great example!

If someone truly drains the data out he should also update the layer_scheme accordingly ?

logist322:
@algesten Done!
Also created a new issue #411 after testing this task

src/packet/vp9.rs
```diff
@@ -13,6 +13,83 @@ const MAX_VP9REF_PICS: usize = 3;
...
+    pub tid: Option<u8>,
+
+    /// Map of the SVC layers widths.
+    ///
+    /// Specified for every spatial layer.
+    pub layers_widths: [Option<u16>; 3],
+
+    /// Map of the SVC layers heights.
+    ///
+    /// Specified for every spatial layer.
```
davibe, line 86: I wonder if it would be nicer to group `layer_*` into a struct by layer

src/packet/vp9.rs
logist322: As far as changing `layers_scheme` does not have any affect on any mechanism and nothing relays on this `layers_scheme` in **str0m** (at least now), we can do anything with it after `MediaData` emitted.
In general, `layers_scheme` field has been added only to let **str0m** clients be able to select SVC layers they need. So, they should decide by themselves what to do with `MediaData::data` and `layers_scheme`.

logist322: I got you. You mean we should implement some synchronization between `MediaData::data` and `layers_scheme`, right?

src/packet/vp9.rs
logist322: Or to join all information specific for the layer under the one `struct LayerInfo {position: usize, width: u16, height: u16}`?

davibe: Yes, I meant joining them. I am trying to imagine how it gets used

```
struct SpatialLayerInfo {
    start: usize, end: usize, width: u16, height: u16
}
```

One can use it like

```
let info = media_data.extra.layers.find(/* by resolution */);
let data_slice = media_data.data[info.start..info.end];
_ = decode(data_slice);
```

or

```
// remove layer 3
let info = media_data.extra.layers[3];
media_data.data.drain(info.start..);
media_data.extra.layers.drain(2..);
```

logist322: Yes, I got your idea. I tried to do the same at the beginning. But there is a problem with huge `MediaData` size. We can not add start field. In my current configuration it is already 440 bytes size.

davibe: I see

davibe (approved):

algesten:
Perfect! Thanks!

PR #409, opened 2023-12-03
Simplify edge case handling in the chat example

davibe:


algesten:
So why are we keeping the noop?

davibe:
No, because `self.rtc.disconnect()` will make it str0m not alive

davibe:
I have tested and I have thought about that. This is what I answered myself

- if there is no timeout to wait for (= the internal str0m state machine has nothing do to next), str0m should not return an immediate timeout (= it should not pretend there is something immediate to do). It would be “violating the contract”.
- [demo.rs](http://demo.rs/) showcases the usage of str0m. It must be correct and simple when possible.

In any case, it’s a tiny aspect. Based on your decision feel free to close this PR.

PR #410, opened 2023-12-06
chat example: request keyframe on discontinuity

davibe:


algesten (approved):
Just some code org and then good to go!

examples/chat.rs
```diff
@@ -409,19 +415,61 @@ impl Client {
...
     }
+    
+    fn handle_media_data_in(&mut self, data: MediaData) -> Propagated {
+        if !data.contiguous {
+            self.request_keyframe_throttled(
+                data.mid,
+                data.rid,
+                KeyframeRequestKind::Fir,
+            );
+        }
```
algesten, line 442: If we are requesting a keyframe, should we really propagate the data?

k0nserv: I think so, with VP9 and VP8 for example, it's possible that the decoder can still do something reasonable because of temporal scalability. Also, if we don't send anything then we'll trigger `onmute` on the `MediaStreamTrack` instead

algesten: Isn't that the wishful thinking we had in our internal service (derp)?

When using media level, the remote side can't see the discontinuity in RTP sequence, so we're effectively hoping the VP9 packet is enough for it to "do something reasonable" with it?

I know that both VP8 and VP9 has an internal sequence number, but didn't we stop doing this for derp?

k0nserv: We still forward, in addition to requesting a PLI

Issue #411, opened 2023-12-06
`DepacketizingBuffer::depack_cache` is not working well


logist322:
While working on #404 found out some strange behavior with `DepacketizingBuffer::depack_cache`.

Once we have `contiguous = false` inside the one `picture_id` (I have been working and testing in context of VP9 codec, but I guess in other codec we can face with something like that), then the whole `picture_id` will be corrupted, even if  `contiguous` becomes `true`. It means when `contiguous` becomes `true` `DepacketizingBuffer::pop` will return data for that `picture_id`, but there will be only data from **RTP packet**s that have sequence numbers greater than the sequence number of the  **RTP packet** that have caused `contiguous = false`.

For example:
1. We have received packets with sequence numbers `[200, 201, 202, 204, 205]` for the same `picture_id = 100`. So we have `contiguous = false` and `DepacketizingBuffer::pop` returns `None`.
2. We have received one more packet with sequence number `203` for the `picture_id = 100`. `contiguous` becomes `true`.
3. Call `DepacketizingBuffer::pop`, it returns `Some`. But in the returned data we have data only from the **RTP packet**s with sequence numbers `[204, 205]`. 

-------------

To make it reproducible I have sniffed some real **VP9 RTP packet**s (RTP and VP9 headers are original, but VP9 payload have been shorted to make it more human readable) and make a test.

To reproduce it by yourself you can run `cargo test --lib -- packet::buffer_rx::test::check_reorder_rtp` on  [**this branch**](https://github.com/logist322/str0m/tree/test-depack-cache). The test  will fail.

But if you comment [**this line**](https://github.com/logist322/str0m/blob/0d6eace9ede71f911ce80d3df399920eb972e09a/src/packet/buffer_rx.rs#L184C16-L184C16), the test will be passed.

> `184 self.depack_cache = Some((seq, dep));`

algesten:
Fixed by #414

PR #414, opened 2023-12-07
Depacketization cache based on sequence numbers

davibe:
Fixes depacketization issues on transient discontinuities explained
in #411


algesten (approved):
A mi me gusta!

PR #415, opened 2023-12-12
DtlsCert not implement Clone

algesten:


algesten:
@xnorpx I don't know if you relied on cloning the `DtlsCert` in your things. There is an escape hatch.

morajabi (approved):

k0nserv:
Why is it a security risk to reuse the certificate? DTLS is just stream oriented TLS, and under TLS certificates are reused all the time after all.

lolgesten:
Hm. You might have a point here. 🤔

lolgesten:
Tried to researching if there been any recommendations on this. Can't find any.

Slack had a problem of a fixed well known private key, which is tangential but not the same case: https://hackerone.com/reports/531032

xnorpx:
Afaik reusing cert on server for multiple connections is standard practice. But you should generate new certs for each connection on the clients.

This is something we do rely on, creation of str0m essentially becomes blocking call if we can't clone it on the server.

Remove Clone and have specific API is probably fine. But I would probably name it more clear.

pub fn clone_for_multiple_connections_on_server(&self) -> DtlsCert {
    DtlsCert {
        pkey: self.pkey.clone(),
        x509: self.x509.clone(),
    }
}

PR #417, opened 2023-12-20
VP8 VP9 H264 allow to detect keyframes

geekylthyosaur:
This PR adds `is_keyframe` field to `Vp8CodecExtra`, `Vp9CodecExtra`, `H264CodecExtra` which indicates that there is an individual frame containing complete and independent visual information. This frame serves as a reference point for other frames in the video sequence.

The only problem that I could not solve is testing of this algorithm for VP9 codec, because `Vp9Depacketizer` is throwing errors on some packets of sample `pcap` file, exported from `Wireshark` by intercepting packets from chat example.

The issue could be that support for VP9 in `Wireshark` was just added in the latest version and maybe it is not feature-full/stable enough. Any ideas?


algesten:
Hi @geekylthyosaur, welcome to str0m!

This looks nice, thanks for doing it! @davibe is our guy for Vp8CodecExtra, I'll let him weigh in.

In terms of problems for Vp9, I'm afraid I haven't used that codec (yet), so I don't really know. This `Vp9CodecExtra` is  very new (see #407), there could be bugs. @logist322 worked on that PR and maybe have some additional information?

src/packet/vp9.rs
```diff
@@ -371,6 +375,7 @@ impl Vp9Depacketizer {
         vp9_extra.pid = self.picture_id;
         vp9_extra.layers_widths.copy_from_slice(&self.width[..3]);
         vp9_extra.layers_heights.copy_from_slice(&self.height[..3]);
+        vp9_extra.is_keyframe |= !self.p && (self.sid == 0 || !self.l) && self.b;
```
davibe, line 378: I don't understand, why this is `|=` ?

davibe:
@algesten I think the idea is useful.

For example if you want to know whether a PLI request actually resulted in a keyframe, you currently have to look into the payload and doing so in a way that is codec-specific (we do this in our internal product). This is simpler.

src/packet/vp9.rs
geekylthyosaur: [In this loop](https://github.com/algesten/str0m/blob/c76993cf7b0c69d6b221979bffa4a73c6f88ccab/src/packet/buffer_rx.rs#L235-L244) the `CodecExtra` can be overwritten, and we do not want `is_keyframe: true` to be replaced with `is_keyframe: false`.

geekylthyosaur:
@logist322 Nope, no artifacts and no errors while capturing packets.

logist322:
@geekylthyosaur I guess in this test you try to parse encrypted payload.

Your VP9 `pcap`:
Screenshot from 2023-12-21 13-02-36

Another VP9 `pcap`:

geekylthyosaur:
@davibe @algesten @logist322 Yeah, the issue was that `pcap` files contained encrypted data and I managed to obtain unencrypted ones. Now this is fixed and all tests are green. Thasnks!

algesten:
@geekylthyosaur great work!

Would you mind squashing the commits and then we merge?

src/packet/vp9.rs
davibe: Those in the loop are all packets of the same "picture".

I think this means we are reporting this media sample as a keyframe if any of the inner spatial layers contains a keyframe, but it's unclear which one. Correct me if I am wrong. I think, either this flag should refer to spatial layer 0 alone, or maybe we should state which spatial layers contains the keyframe ?

davibe: Ah, I got confused :) I see why this works now.

algesten (approved):
Excellent!

Thanks!

PR #418, opened 2023-12-22
De-multiplex STUN replies based on transaction ID

thomaseizinger:
Previously, `accept_message` would validate that incoming STUN requests match the credentials we have locally. This allows a server to de-multiplex, which agent should handle an particular STUN request.

In a peer-to-peer setting, both parties will make STUN requests and thus, we also need to de-multiplexing incoming STUN replies. Luckily this is quite easy because the candidate pairs that we have formed locally keep track of the transaction IDs that we've used so far. If an inbound message is a STUN reply but we don't recognise the transaction ID, we now return `false` from `accepts_message`.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1781,4 +1781,41 @@ mod test {
             }
         }
     }
+
+    #[test]
+    fn does_not_accept_response_with_unknown_transaction_id() {
+        let mut agent = IceAgent::new();
+        let remote_creds = IceCreds::new();
+        let mut remote_candidate = Candidate::host(ipv4_3(), "udp").unwrap();
+        remote_candidate.set_ufrag(&remote_creds.ufrag);
+
+        agent.set_remote_credentials(remote_creds.clone());
+        agent.add_local_candidate(Candidate::host(ipv4_1(), "udp").unwrap());
+        agent.add_remote_candidate(remote_candidate);
+        agent.handle_timeout(Instant::now());
+
+        let payload = Vec::from(agent.poll_transmit().unwrap().contents);
+        let stun_message = StunMessage::parse(&payload).unwrap();
+
+        let valid_reply =
+            make_authenticated_stun_reply(stun_message.trans_id(), ipv4_4(), &remote_creds.pass);
+        let fake_reply =
+            make_authenticated_stun_reply(TransId::new(), ipv4_4(), &remote_creds.pass);
+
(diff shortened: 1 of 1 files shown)


src/ice/agent.rs
thomaseizinger, line 1: The docs of this function will need some adjustment. I am open to suggestions here! Also, I am not entirely sure if this was/is the only missing piece for str0m to better support peer-to-peer full-ICE. I guess more experiments will tell :)

src/ice/agent.rs
```diff
@@ -775,6 +775,18 @@ impl IceAgent {
...
+                .candidate_pairs
+                .iter()
+                .any(|p| p.has_binding_attempt(message.trans_id()))
+        {
+            trace!(
+                "Message rejected, transaction ID does not belong to any of our candidate pairs"
+            );
+            return false;
+        }
+
```
algesten, line 789: (Niggly?) style note. I like to break up iterator/combinator style expressions from the if-statement. That gives us both an opportunity to have a self-documenting variable name as well as a good place for further optional code doc, if needed.

https://github.com/algesten/str0m/blob/main/src/change/sdp.rs#L964-L978

That does give us two nested if-statements, but on balance I think that's OK in this instance.

```rs
if message.is_successful_binding_response() {
    // <optional clarifying doc if variable name isn't enough>
    let good_variable_name = !self.candidate_pairs.iter().any(…);
    
    if !good_variable_name {
        return false;
    }    
}
```

I'm not being consistent with this, it's something I increasingly try to do. The trigger for me is when an if-statement condition starts to break on to multiple lines.

algesten (approved):
Nice find!

This is such a "duh", I don't even understand why this wasn't done already.

Thanks

src/ice/agent.rs
thomaseizinger: I also did this and then clippy complained that they can be collapsed into one! Do you want me to add a `#[allow(clippy::)]` for the lint or should I find a way of refactoring it such that the lint doesn't fire?

thomaseizinger:
> Nice find!
> 
> This is such a "duh", I don't even understand why this wasn't done already.

Thanks! I was a bit worried that I am just missing something. Great to hear that it was simply overlooked. Sometimes things do come easy :)

src/ice/agent.rs
algesten: As you can see in `lib.rs`, I'm quite happy to turn off lints. I don't like clippy getting in the way for code clarity.

algesten:
Thanks!

PR #419, opened 2023-12-22
Allow low-level access to `IceAgent`, `IceAgentEvent` & `StunMessage` via `ice` module

thomaseizinger:
In order to allow low level access to `ice` in a similar way as `rtp`, we introduce a public `ice` module that rexports `IceAgent` and `IceAgentEvent`. In addition, we also export `StunMessage` publicly from the `ice` module. To keep the public API small, all functions apart from `StunMessage::parse` are marked as `pub(crate)`. For standalone usage of `IceAgent`, we only need to be able to parse `StunMessage`s but not expose any other details.

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -141,12 +141,12 @@ impl<'a> StunMessage<'a> {
     }
 
     /// Whether this STUN message is a BINDING request.
-    pub fn is_binding_request(&self) -> bool {
+    pub(crate) fn is_binding_request(&self) -> bool {
         self.method == Method::Binding && self.class == Class::Request
     }
 
     /// Whether this STUN message is a response.
-    pub fn is_response(&self) -> bool {
+    pub(crate) fn is_response(&self) -> bool {
         matches!(self.class, Class::Success | Class::Failure)
     }
 
@@ -154,17 +154,17 @@ impl<'a> StunMessage<'a> {
     ///
     /// STUN binding requests are very simple, they just return the observed address.
     /// As such, they cannot actually fail which is why we don't have `is_failed_binding_response`.
-    pub fn is_successful_binding_response(&self) -> bool {
+    pub(crate) fn is_successful_binding_response(&self) -> bool {
         self.method == Method::Binding && self.class == Class::Success
     }
 
     /// The transaction ID of this STUN message.
-    pub fn trans_id(&self) -> TransId {
+    pub(crate) fn trans_id(&self) -> TransId {
         self.trans_id
     }
 
     /// Constructs a new BINDING request from the provided data.
-    pub fn binding_request(
+    pub(crate) fn binding_request(
         username: &'a str,
         trans_id: TransId,
         controlling: bool,
@@ -200,7 +200,7 @@ impl<'a> StunMessage<'a> {
     }
 
     /// Constructs a new STUN BINDING reply.
-    pub fn reply(trans_id: TransId, mapped_address: SocketAddr) -> StunMessage<'a> {
+    pub(crate) fn reply(trans_id: TransId, mapped_address: SocketAddr) -> StunMessage<'a> {
         StunMessage {
             class: Class::Success,
             method: Method::Binding,
@@ -216,28 +216,28 @@ impl<'a> StunMessage<'a> {
     }
 
     /// If present, splits the value of the USERNAME attribute into local and remote (separated by `:`).
-    pub fn split_username(&self) -> Option<(&str, &str)> {
+    pub(crate) fn split_username(&self) -> Option<(&str, &str)> {
         self.attrs.split_username()
     }
 
     /// If present, returns the value of XOR-MAPPED-ADDRESS attribute.
-    pub fn mapped_address(&self) -> Option<SocketAddr> {
+    pub(crate) fn mapped_address(&self) -> Option<SocketAddr> {
         self.attrs.mapped_address()
     }
 
     /// If present, returns the value of the PRIORITY attribute.
-    pub fn prio(&self) -> Option<u32> {
+    pub(crate) fn prio(&self) -> Option<u32> {
         self.attrs.prio()
     }
 
     /// Whether this message has the USE-CANDIDATE attribute.
-    pub fn use_candidate(&self) -> bool {
+    pub(crate) fn use_candidate(&self) -> bool {
         self.attrs.use_candidate()
     }
 
     /// Verify the integrity of this message against the provided password.
     #[must_use]
-    pub fn check_integrity(&self, password: &str) -> bool {
+    pub(crate) fn check_integrity(&self, password: &str) -> bool {
         if let Some(integ) = self.attrs.message_integrity() {
             let sha1: Sha1 = password.as_bytes().into();
             let comp = sha1.hmac(&[
@@ -254,7 +254,7 @@ impl<'a> StunMessage<'a> {
     /// Serialize this message into the provided buffer, returning the final length of the message.
     ///
     /// The provided password is used to authenticate the message.
-    pub fn to_bytes(&self, password: &str, buf: &mut [u8]) -> Result<usize, StunError> {
+    pub(crate) fn to_bytes(&self, password: &str, buf: &mut [u8]) -> Result<usize, StunError> {
         let attr_len = self.attrs.iter().fold(0, |p, a| p + a.padded_len());
         let msg_len = 20 + attr_len;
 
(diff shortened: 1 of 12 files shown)


thomaseizinger:
An alternative to exposing the types could be to make `ice` public and drop the `Ice` prefix from all the types. Users would refer to them as `ice::Agent`, `ice::Event` etc.

src/io/stun.rs
```diff
@@ -231,12 +255,10 @@ impl<'a> StunMessage<'a> {
         }
     }
 
+    /// Serialize this message into the provided buffer, returning the final length of the message.
+    ///
+    /// The provided password is used to authenticate the message.
```
algesten, line 260: I love the documentation in this file, thanks!

However I don't want to expose more API than strictly necessary. For your use case, do you need all these functions to be `pub`, or could they be considered an internal detail to str0m's ice module and thus be `pub(crate)`?

src/lib.rs
```diff
@@ -597,14 +597,15 @@ use dtls::Fingerprint;
...
-pub use ice::{Candidate, CandidateKind};
+pub use ice::{
+    Candidate, CandidateKind, IceAgent, IceAgentEvent, IceConnectionState, IceCreds, IceError,
+};
 
 mod io;
 use io::DatagramRecv;
 
+pub use io::{StunError, StunMessage};
+
```
algesten, line 608: On the root level of str0m, I only want the types related to the "standard" WebRTC use case. This includes `Candidate` and `CandidateKind`, since a user must interact with those to create an `Rtc` instance. They should remain where they are.

However the rest of the ice types are a "low level use case", which I think should live in an `ice` module. A module we don't expose by default, but with certain feature flags (`ice`?).

Having them as a top level module API wise, doesn't mean we need to pub our internal `ice` mod. See below trick with `rtp_` and `#[path = …]`. I suggest we do this the same way.

I think it would also be fine to have `StunMessage` and `StunError` sitting under this new "synthehtic" `ice` module.

As a starting point we could keep `IceAgent` internally, but rename the export (`pub use ice_::IceAgent as Agent`) in this synthetic module. The drawback is that type name changes/aliases can be confusing in Rust. However the alternative of dropping the `Ice`-prefix would make for inconsistent internal naming in str0m (compare `IceError`, `RtpError`, `DtlsError` etc). We can revisit this later if it sucks too much.

algesten (changes requested):
Looking good!

A couple of notes.

src/io/stun.rs
thomaseizinger: I actually only need `StunMessage::parse`. I could maybe work around that too if we change the interaction of `accepts_message` and `handle_receive`. I'll propose a patch to get your input.

src/lib.rs
thomaseizinger: Sounds good to me!

src/io/stun.rs
thomaseizinger: See https://github.com/algesten/str0m/pull/420.

thomaseizinger:
Setting this to draft whilst I am extracting individual bits from it.

src/io/stun.rs
```diff
@@ -32,16 +32,16 @@ pub fn stun_resend_delay(send_count: usize) -> Duration {
...
+    /// A STUN message could not be parsed.
     #[error("STUN parse error: {0}")]
     Parse(String),
 
+    /// An IO error occurred while handling a STUN message.
     #[error("STUN io: {0}")]
     Io(#[from] io::Error),
-
-    #[error("STUN error: {0}")]
-    Other(String),
```
thomaseizinger, line 44: Extracted removal of this into a dedicated PR: https://github.com/algesten/str0m/pull/421.

src/io/stun.rs
algesten: Seeing #420, I think I want to backtrack on that idea. `Receive` is an abstraction to help demultiplex STUN, DTLS, RTP, etc by packet inspection. Thinking about the `IceAgent` in isolation, it will never need this demultiplexing – i.e. `Receive` is a "too general" type for `ice`.

I think we should pub expose `StunMessage::parse`, and add the `StunMessage` to the synthetic `ice` module since that it's tightly coupled with the `IceAgent`. The rest of the `StunMessage` fns can be `pub(crate)` though.

thomaseizinger: Okay, so do we want to change `handle_receive` to `StunMessage` too then? We will need an additonal parameter for the address. That one is currently taken out of `net::Receive`.

thomaseizinger: > Okay, so do we want to change `handle_receive` to `StunMessage` too then? We will need an additonal parameter for the address. That one is currently taken out of `net::Receive`.

I am not sure how feasible that is. `stun_server_handle_message` needs everything that is within `net::Receive`:

https://github.com/algesten/str0m/blob/bec55eda9c121b86adc0ee19364f061f533a1294/src/ice/agent.rs#L822

I can make the function not depend on `net::Receive` but then we will just pass `proto`, `source` and `destination` in as separate parameters. That is a fair choice and reduces the dependencies of the the `stun` module. However, the agent already depends on many parts from the `io` module:

https://github.com/algesten/str0m/blob/bec55eda9c121b86adc0ee19364f061f533a1294/src/ice/agent.rs#L7-L10

Thoughts?

algesten: Ok. I'll look at this tomorrow morning CET when my brain is fresh.

thomaseizinger: > I can make the function not depend on `net::Receive` but then we will just pass `proto`, `source` and `destination` in as separate parameters.

Let me know what you think of https://github.com/algesten/str0m/pull/422.

algesten (approved):
This looks ready to me.

Thank you for taking it all the way!

src/lib.rs
```diff
@@ -604,6 +604,7 @@ use ice_::IceAgentEvent;
 pub use ice_::{Candidate, CandidateKind, IceConnectionState};
 
 /// Low level ICE access.
+#[cfg_attr(not(feature = "ice-agent"), doc(hidden))]
```
thomaseizinger, line 607: So this just hides the module from the docs unless the feature is activated? Fine with me but I'd be curious to understand the thinking!

algesten: Yeah, it does.

For new users coming to  str0m, I want the Main API to be clear and obvious. We're already breaking with convention by not using the RTCPeerConnection API, so I want a new user's first impressions to lead them straight to where they need to be.

This is why I'm very particular which structs/enums you see in the root – they should all be relevant and things you probably have to interact with as a normal user. `IceCreds` is right there on the boundary – it's possible to get the value from `RtcConfig`, but I don't see a regular user needing it, hence I moved it.

The ice agent is an internal concern to such a user (you never need it and there's no way to insert or interact with it through config or the `Rtc` instance).

thomaseizinger: Thank you for explaining, keeping the root module tidy is definitely something that I consider important. What I don't really understand is why we need a feature-flag for that.

Would it be better to just use `doc(hidden)` unconditionally? I can use the `ice `module without activating the feature, so why would I activate it? To look at the docs locally? If you don't build with all features in docs.rs, then advanced users won't be able to discover the API. If you do build with all features, then the API is documented again, potentially confusing new users.

FWIW: I consider building in docs.rs with all features to be good practice. Together with `doc_auto_cfg`, that explains to users, which features they must activate to get access to certain modules. See `rust-libp2p` for some examples:

- Building with all features in `docs.rs`: https://github.com/libp2p/rust-libp2p/blob/4260702f749d45fb53ea7edd2fb0a728afb83488/libp2p/Cargo.toml#L150-L155
- Activation of `doc_auto_cfg`: See https://github.com/libp2p/rust-libp2p/blob/4260702f749d45fb53ea7edd2fb0a728afb83488/libp2p/src/lib.rs#L33
- Rendered docs.rs: https://docs.rs/libp2p/latest/libp2p/autonat/index.html

I don't think `str0m` needs it at this point because there aren't any features yet (worth documenting). But as we introduce the `crypto` module, this might be something to consider.

algesten: For a while there I was trying to see if we can avoid compiling the rest of str0m when you only want the ice stuff. What I did was.

1. Refactor `Rtc` into `rtc.rs` and `RtcConfig` into `config.rs`
2. Add another feature `webrtc` that is on default and added to most things in `lib.rs`

Sadly it became more of a mess than had hoped for – `#[cfg(feature = "webrtc")]` in a bit too many places. When I also realised the dependency between the modules ice -> parser had to be sorted somehow, I lost interest. 

At this point maybe `#[doc(hidden)]` would be cleaner, true.

thomaseizinger: > At this point maybe `#[doc(hidden)]` would be cleaner, true.

I'll send a follow-up!

PR #420, opened 2023-12-22
Re-work `IceAgent::accepts_message` and `IceAgent::handle_receive`

thomaseizinger:
In preparation for making `IceAgent` public, we re-work parts of its public API to make the process easier. In particular, we remove `StunMessage` from the public API by changing the signature of `accepts_message` to take a `net::Receive` instead. `net::Receive` is already part of str0m's public API and can be easily constructed by end-users.

To align the two APIs further, we also return a `bool` from `handle_receive`, indicating whether or not the message was actually handled.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1537,6 +1537,8 @@ impl IceAgent {
 
 #[cfg(test)]
 mod test {
+    use crate::net;
+
     use super::*;
     use std::net::SocketAddr;
     use std::sync::Once;
@@ -1818,8 +1820,12 @@ mod test {
         let fake_reply =
             make_authenticated_stun_reply(TransId::new(), ipv4_4(), &remote_creds.pass);
 
-        assert!(!agent.accepts_message(&StunMessage::parse(&fake_reply).unwrap()));
-        assert!(agent.accepts_message(&StunMessage::parse(&valid_reply).unwrap()));
+        assert!(!agent.accepts_message(
+            &net::Receive::new(Protocol::Udp, ipv4_1(), ipv4_1(), &fake_reply).unwrap()
+        ));
+        assert!(agent.accepts_message(
+            &net::Receive::new(Protocol::Udp, ipv4_1(), ipv4_1(), &valid_reply).unwrap()
+        ));
     }
 
     fn make_authenticated_stun_reply(tx_id: TransId, addr: SocketAddr, password: &str) -> Vec<u8> {
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -743,7 +743,11 @@ impl IceAgent {
     ///
     /// If no remote credentials have been set using `set_remote_credentials`, the remote
     /// ufrag is not checked.
-    pub fn accepts_message(&self, message: &StunMessage<'_>) -> bool {
+    pub fn accepts_message(&self, receive: &Receive) -> bool {
+        let DatagramRecv::Stun(message) = &receive.contents else {
+            return false;
+        };
+
         trace!("Check if accepts message: {:?}", message);
 
         // The username for the credential is formed by concatenating the
@@ -803,6 +807,11 @@ impl IceAgent {
     pub fn handle_receive(&mut self, now: Instant, r: Receive) {
         trace!("Handle receive: {:?}", r);
 
+        if !self.accepts_message(&r) {
+            debug!("Message not accepted");
+            return;
+        }
+
         let message = match r.contents {
             DatagramRecv::Stun(v) => v,
             _ => {
@@ -810,13 +819,6 @@ impl IceAgent {
                 return;
             }
         };
-
-        // Regardless of whether we have remote_creds at this point, we can
-        // at least check the message integrity.
-        if !self.accepts_message(&message) {
-            debug!("Message not accepted");
-            return;
-        }
 
         if message.is_binding_request() {
             self.stun_server_handle_message(now, r.proto, r.source, r.destination, message);
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -804,19 +804,19 @@ impl IceAgent {
     /// Handles an incoming STUN message.
     ///
     /// Will not be used if [`IceAgent::accepts_message`] returns false.
-    pub fn handle_receive(&mut self, now: Instant, r: Receive) {
+    pub fn handle_receive(&mut self, now: Instant, r: Receive) -> bool {
         trace!("Handle receive: {:?}", r);
 
         if !self.accepts_message(&r) {
             debug!("Message not accepted");
-            return;
+            return false;
         }
 
         let message = match r.contents {
             DatagramRecv::Stun(v) => v,
             _ => {
                 trace!("Receive rejected, not STUN");
-                return;
+                return false;
             }
         };
 
@@ -832,6 +832,8 @@ impl IceAgent {
         });
 
         // TODO handle unsuccessful responses.
+
+        true
     }
 
     pub fn handle_timeout(&mut self, now: Instant) {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1492,7 +1492,7 @@ impl Rtc {
 
         // STUN can use the ufrag/password to identify that a message belongs
         // to this Rtc instance.
-        self.ice.accepts_message(&r)
+        self.ice.accepts_message(r)
     }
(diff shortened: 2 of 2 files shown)


thomaseizinger:
> To align the two APIs further, we also return a `bool` from `handle_receive`, indicating whether or not the message was actually handled.

For my usecase, having `handle_receive` with a bool return value is good enough. I simply iterate through all agents until one of them says that it handled the message. Because we call `accepts_message` internally, I think this is safe. This might extend to APIs further up the stack in str0m as well or is there a usecase where you first call `accepts` but then don't actually want the agent to handle the message?

thomaseizinger:
Closing this in favor of https://github.com/algesten/str0m/pull/422.

thomaseizinger:
> This might extend to APIs further up the stack in str0m as well or is there a usecase where you first call `accepts` but then don't actually want the agent to handle the message?

@algesten I am curious what you think of this. In the `IceAgent`, we use `accepts_message` also internally to guard, which data can actually modify the state of the agent. Instead of having a separate `Rtc::accepts` function, I think we should:

- Extend `Rtc::handle_input` and all `handle_receive` functions with a check whether they actually _want_ the given data.
- Extend those functions with a `bool` return value that indicates whether they handled the data.

I believe the combination of the above would allow us to remove `Rtc::accepts` entirely. Instead, users would iterate over all `Rtc` instances and simply attempt to handle the given message. Perhaps we should rename the function to `Rtc::try_handle_receive`? Assuming that we have sufficiently efficient enough guards at the top of all `handle_receive` functions, this should be equivalent from a performance PoV but provides a simpler API to users.

If you agree, then I can build the above on top of https://github.com/algesten/str0m/pull/422.

algesten:
> @algesten I am curious what you think of this. In the `IceAgent`, we use `accepts_message` also internally to guard, which data can actually modify the state of the agent. Instead of having a separate `Rtc::accepts` function, I think we should:
> 
> * Extend `Rtc::handle_input` and all `handle_receive` functions with a check whether they actually _want_ the given data.
> * Extend those functions with a `bool` return value that indicates whether they handled the data.
> 
> I believe the combination of the above would allow us to remove `Rtc::accepts` entirely. Instead, users would iterate over all `Rtc` instances and simply attempt to handle the given message. Perhaps we should rename the function to `Rtc::try_handle_receive`? Assuming that we have sufficiently efficient enough guards at the top of all `handle_receive` functions, this should be equivalent from a performance PoV but provides a simpler API to users.

We're building a larger server (we do an SFU), and we multiplex all client communication from a single UDP port (mainly for instructing our clients about firewalls "open udp traffic for port x"). We also got each `Rtc` running in it's own async task. Incoming UDP data is read centrally, and dispatched to the correct task. Figuring out which task will accept newly discovered remote addresses is a distinctly different activity to dispatching. This is also why `.accept(&self, input: &Input)` takes a reference vs `.handle_input(&self, input: Input)` takes ownership.

On balance I think this distinction makes more sense than merging the two operations.

thomaseizinger:
Thank you for your answer! I have a couple of questions if you don't mind :)

> We also got each `Rtc` running in it's own async task. Incoming UDP data is read centrally, and dispatched to the correct task.

How do you share the `Rtc` instance between the UDP socket and the individual task? I am assuming you are calling `Rtc::accepts` next to the socket and then dispatch it to the correct task? If yes, doesn't that require `Arc<Mutex>`? If not, then I am not sure I understand the point of `accepts` yet.

In either case, doesn't dispatching to a separate task require copying the memory? I am curious if you've done any performance measurement on this, i.e. comparing the (potential) contention of the lock + copying memory with the ability to dispatch incoming packets quicker.

> This is also why `.accept(&self, input: &Input)` takes a reference vs `.handle_input(&self, input: Input)` takes ownership.

But `Input` always has a reference to the data so having ownership over `Input` doesn't really mean anything, does it? The current modelling of `StunMessage` uses a `Vec` for attributes so it is not `Copy`. But because we only support a limited set of attributes anyway, we could have them all as `Option<Username>` etc, thereby avoid the `Vec` and make `StunMessage` and consequently `Input` implement `Copy`! Alternatively, we could also use a fixed-sized array.

> On balance I think this distinction makes more sense than merging the two operations.

Would you be open to _extend_ `handle_input` to

a) also perform this check internally like we do for the `IceAgent`
b) return a boolean to check whether it has been handled

This would allow the use without `accepts` but if certain usecases require it, they can still use `accepts`.

algesten:
> In either case, doesn't dispatching to a separate task require copying the memory? I am curious if you've done any performance measurement on this, i.e. comparing the (potential) contention of the lock + copying memory with the ability to dispatch incoming packets quicker.

We do this via bidirectional tokio mpsc channels (think RPC). When unknown data arrives at the UdpSocket, we send messages with `Arc<Vec<u8>>` to all tasks. The data is parsed in each task (at least that lifetime on `Receive<'a>` comes in handy 😄). Which leads us on to:

> But `Input` always has a reference to the data so having ownership over `Input` doesn't really mean anything, does it?

Very true. And thinking about it, the `Vec` over the STUN options this means quite a few unnecessary allocations, mitigated somewhat by that having unidentified traffic is quite rare.

> Would you be open to _extend_ `handle_input` to

Let me try another motivation why it feels awkward. Str0m is Sans-IO. That is by no means a rigid design pattern (though I sincerely wish more Rustaceans obsessed over figuring out Sans-IO patterns instead of async!), though there is some loose precedence.

The original inspiration for Sans-IO was [`quinn-proto`](https://docs.rs/quinn-proto/0.10.6/quinn_proto/struct.Connection.html) but also see [this blog post](https://fractalideas.com/blog/sans-io-when-rubber-meets-road/). quinn-proto has a split `handle_event/handle_timout` – I combined them to get the symmetry `handle_input(Input)`/`poll_output -> Output`. However in both referenced examples, the `handle_<whatever>` are void returns.

Sans-IO is obviously no standard (yet?), and even if it was, no design pattern is a rule. I'm sure we can also find examples doing this differently in other places. Overall though, I'm not unhappy with splitting out the `accepts()` – it seems clean to me. Not overload the semantics of `handle_input`.

thomaseizinger:
> > In either case, doesn't dispatching to a separate task require copying the memory? I am curious if you've done any performance measurement on this, i.e. comparing the (potential) contention of the lock + copying memory with the ability to dispatch incoming packets quicker.
> 
> We do this via bidirectional tokio mpsc channels (think RPC). When unknown data arrives at the UdpSocket, we send messages with `Arc<Vec<u8>>` to all tasks.

Interesting! Have you done any performance measurement on the impact of allocating each packet in order to parallelize the processing? What I am currently going for is to have a single loop that reads a packet into a re-used buffer, iterates over all agents and optionally sends out any generated packets at the end before reading the next one (to drop the mutable reference to the read buffer).

> Sans-IO is obviously no standard (yet?), and even if it was, no design pattern is a rule. I'm sure we can find examples doing this differently in other places. Overall though, I'm not unhappy with splitting out the `accepts()` – it seems clean to me. Not overload the semantics of `handle_input`.

What bugs me is that the API isn't as misuse-resistant as it could be: In my opinion, the `handle_input` function should not depend on the user calling `accept` first because we cannot guarantee that. Hence, we should really call it internally at the top and exit early to be safe. At that point, it feels silly to even provide `accept` because why tell the user to call it if we are gonna call it anyway? Returning a boolean to tell the user whether we handled the data is the final conclusion to allow for a small optimisation to stop iterating over the list of agents.

The situation would be different if the pass-by-value of `handle_input` would matter. It does at the moment because `StunMessage` is not `Copy` but that can be fixed as pointed out above. With that being out of the way, I don't really see the value that `accepts` provides.

All this being said, the API as is works fine for me / us. The design of the library I am building on top of `str0m` is heavily inspired by its API and patterns and initially, I also made an `accepts` function. However, I quickly noticed that I don't _need_ it so I removed it in favor of a unified API and a loop that stops iterating once the packet is handled.

> I combined them to get the symmetry `handle_input(Input)`/`poll_output -> Output`.

Side note on this: I think this is actually not ideal. Being able to send data depends on our socket being ready. if you are running over a protocol with backpressure, you'd want to delay reading more packets until you've sent all buffered ones. From that PoV, it would be better to be able to poll for `Transmit` separately. Informational events typically don't need to be backpressured so those could still be handled whereas the call to `poll_transmit` can be delayed until the socket is ready to accept more data.

PR #422, opened 2023-12-22
Decouple low-level components from `net::Receice`

thomaseizinger:
Currently, `net::Receive` is used to parse and de-multiplex incoming messages between STUN, DTLS, RTP and RTCP. `Rtc` however does not use the parsed message and instead passes on the entire `net::Receive` struct to the sub-components. This requires each low-level component like `Dtls`, `Session` and `IceAgent` to revalidate that they are in fact only being passed the data that they can handle.

To fix this, we change the signatures of the various `handle_receive` functions to only take in the parsed message. For DTLS, RTP and RTCP, this is only a byte-buffer which isn't the most type-safe way to go about it. Those APIs are crate-private though so for now, this isn't be a big issue.

### DIFF
--- a/src/dtls/mod.rs
+++ b/src/dtls/mod.rs
@@ -6,7 +6,7 @@ use std::io::{self, ErrorKind, Read, Write};
 use std::net::SocketAddr;
 use thiserror::Error;
 
-use crate::io::{DatagramRecv, DatagramSend, Receive, DATAGRAM_MTU_WARN};
+use crate::io::{DatagramSend, DATAGRAM_MTU_WARN};
 
 mod ossl;
 use ossl::{dtls_create_ctx, dtls_ssl_create, TlsStream};
@@ -251,15 +251,7 @@ impl Dtls {
     }
 
     /// Handles an incoming DTLS datagrams.
-    pub fn handle_receive(&mut self, r: Receive) -> Result<(), DtlsError> {
-        let message = match r.contents {
-            DatagramRecv::Dtls(v) => v,
-            _ => {
-                trace!("Receive rejected, not DTLS");
-                return Ok(());
-            }
-        };
-
+    pub fn handle_receive(&mut self, message: &[u8]) -> Result<(), DtlsError> {
         self.tls.inner_mut().set_incoming(message);
 
         if self.handle_handshake()? {
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -5,9 +5,9 @@ use std::time::{Duration, Instant};
 use rand::random;
 
 use crate::io::Protocol;
-use crate::io::{DatagramRecv, Receive, Transmit, DATAGRAM_MTU};
 use crate::io::{Id, DATAGRAM_MTU_WARN};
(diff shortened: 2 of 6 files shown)


algesten:
@thomaseizinger this looks like a good refactor!

I tried adding some encapsulation of the parameters to IceAgent::handle_receive. What do you think?

thomaseizinger:
> @thomaseizinger this looks like a good refactor!
> 
> I tried adding some encapsulation of the parameters to IceAgent::handle_receive. What do you think?

I agree that the number of parameters is mildly annoying but I don't see it as a big issue actually. I think I'd rather have that than two very similarly named structs (`StunMessage` and `StunPacket`). Plus, we now seem to have some more error cases so to be frank, I am not convinced it is a better design.

I toyed with the idea of making `net::Receive` generic so we can reuse it for concrete messages and not just `DatagramRecv` but I am not sure if that wouldn't add too much complexity.

algesten:
> I agree that the number of parameters is mildly annoying but I don't see it as a big issue actually. I think I'd rather have that than two very similarly named structs (`StunMessage` and `StunPacket`). Plus, we now seem to have some more error cases so to be frank, I am not convinced it is a better design.

Hm. I was quite pleased with it :)

What if we remove the error? Converting Receive -> StunPacket is a strictly internal concern for str0m, so the TryFrom is a bit much really.

thomaseizinger:
@algesten I pushed two more commits. The first one removes the error case of converting an non-STUN `Receive` into a `StunPacket`. The second one might be a bit more controversial (and is a breaking change). `StunPacket` is essentially the same as `Receive` just with a different payload so we can reuse the struct by making `contents` generic. Let me know what you think :)

algesten:
@thomaseizinger not quite feeling it. The code duplication that were in my version between `StunPacket` and `Receive` didn't bother me that much. I'm thinking of this in terms of _two APIs_. Someone using the main WebRTC API, is unlikely to want to use the ICE API, and vice versa. In an extension, the ICE API could become its own crate (though there are strong motivations to keep it in-tree), at which point code duplication would be unavoidable.

algesten:
Random thought: In the spirit of two APIs, would it make sense to refactor out the stun stuff to not be this comingled with the other net code?

thomaseizinger:
> Random thought: In the spirit of two APIs, would it make sense to refactor out the stun stuff to not be this comingled with the other net code?

I think that is too intertwined at the moment and it would create a lot of duplication and mapping that is not necessary whilst we are within the same crate.

thomaseizinger:
> @thomaseizinger not quite feeling it. The code duplication that were in my version between `StunPacket` and `Receive` didn't bother me that much. I'm thinking of this in terms of _two APIs_. Someone using the main WebRTC API, is unlikely to want to use the ICE API, and vice versa. In an extension, the ICE API could become its own crate (though there are strong motivations to keep it in-tree), at which point code duplication would be unavoidable.

Fair! I'll drop the last commit then :)

thomaseizinger:
@algesten I dropped the controversial commit. This is ready for a re-review from my end.

algesten (approved):
Let's ship it!

PR #424, opened 2023-12-24
Move IceCreds/DtlsCert/Fingerprint to root

algesten:
We want to keep the root clean, but the the reason for these sitting
under str0m::change is becoming increasingly untenable.

This is in anticipation of #419


algesten:
Meh. Not happy with this either. Polluting the root too much.

PR #425, opened 2023-12-27
Don't allocate while parsing `StunMessage`

thomaseizinger:
Currently, attributes are parsed into a `Vec` which results in an allocation for each STUN message that is parsed. We only support a limited set of attributes in `str0m` so this can also be expressed using `Option`s which allows us to parse STUN messages allocation free.

As a result, `StunMessage` now implements `Copy`!

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -243,7 +243,7 @@ impl<'a> StunMessage<'a> {
     /// Serialize this message into the provided buffer, returning the final length of the message.
     ///
     /// The provided password is used to authenticate the message.
-    pub(crate) fn to_bytes(&self, password: &str, buf: &mut [u8]) -> Result<usize, StunError> {
+    pub(crate) fn to_bytes(self, password: &str, buf: &mut [u8]) -> Result<usize, StunError> {
         const MSG_HEADER_LEN: usize = 20;
         const MSG_INTEGRITY_LEN: usize = 20;
         const FPRINT_LEN: usize = 4;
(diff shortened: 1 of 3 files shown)


thomaseizinger:
@algesten This is best reviewed commit-by-commit.

thomaseizinger:
@algesten I also extracted a few things into named constants to have less magic constants flying around. Don't feel strongly about that though so happy to drop those commits if you don't like it :)

src/io/stun.rs
```diff
@@ -360,177 +364,137 @@ impl Method {
...
-    FingerprintMark,  // crc32
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct Attributes<'a> {
+    username: Option<&'a str>,              // < 128 utf8 chars
+    message_integrity: Option<&'a [u8]>,    // 20 bytes sha-1
+    error_code: Option<(u16, &'a str)>,     // 300-699 and reason phrase < 128 utf8 chars
+    realm: Option<&'a str>,                 // < 128 utf8 chars
+    nonce: Option<&'a str>,                 // < 128 utf8 chars
+    xor_mapped_address: Option<SocketAddr>, // 0x0020
+    software: Option<&'a str>,
```
algesten, line 375: If we add a doc comment to `software`, I think cargo fmt will align all these comments by itself.

algesten (approved):
👏 This all looks great. A very nice cleanup! Thanks!

src/io/stun.rs
thomaseizinger: I'll see what I can do!

PR #426, opened 2023-12-27
Align comments in on fields in `Attributes`

thomaseizinger:
I had to remove the two links in order to make rustfmt actually align those comments. For `software`, I added the attribute ID as it is done for some of the others.

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -378,7 +378,7 @@ pub struct Attributes<'a> {
     use_candidate: bool,                    // 0x0025
     ice_controlled: Option<u64>,            // 0x8029
     ice_controlling: Option<u64>,           // 0x802a
-    network_cost: Option<(u16, u16)>,       // 0xc057
+    network_cost: Option<(u16, u16)>,       // 0xc057 https://tools.ietf.org/html/draft-thatcher-ice-network-cost-00
 }
 
 impl<'a> Attributes<'a> {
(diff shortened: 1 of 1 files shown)


src/io/stun.rs
```diff
@@ -372,15 +372,13 @@ pub struct Attributes<'a> {
...
-    // https://tools.ietf.org/html/rfc8445
-    priority: Option<u32>,        // 0x0024
-    use_candidate: bool,          // 0x0025
-    ice_controlled: Option<u64>,  // 0x8029
-    ice_controlling: Option<u64>, // 0x802a
-    // https://tools.ietf.org/html/draft-thatcher-ice-network-cost-00
-    network_cost: Option<(u16, u16)>, // 0xc057
+    software: Option<&'a str>,              // 0x0022
+    fingerprint: Option<u32>,               // crc32
+    priority: Option<u32>,                  // 0x0024
```
algesten, line 377:
```suggestion
    priority: Option<u32>,                  // 0x0024 https://tools.ietf.org/html/rfc8445
```

src/io/stun.rs
```diff
@@ -372,15 +372,13 @@ pub struct Attributes<'a> {
...
+    use_candidate: bool,                    // 0x0025
+    ice_controlled: Option<u64>,            // 0x8029
+    ice_controlling: Option<u64>,           // 0x802a
+    network_cost: Option<(u16, u16)>,       // 0xc057
```
algesten, line 381:
```suggestion
    network_cost: Option<(u16, u16)>,       // 0xc057 https://tools.ietf.org/html/draft-thatcher-ice-network-cost-00
```

PR #428, opened 2023-12-28
Reorganize fuzz and integration test code

lolgesten:
This moves the TestRtc and associated functions used
in the integration tests to the module str0m::_internal_text_exports.
By having it in the main source tree we can share the code between
fuzz tests and integration tests.


davibe (approved):

algesten:
I'm going to hold off merging this a bit, since maybe using `TestRtc` in fuzz-tests is just too damn slow.

algesten:
Let's not do all of this. `TestRtc` does not need to be shared. I'll do another PR.

PR #429, opened 2023-12-28
nack register: fix handling big seq jumps

davibe:
A very large seq jump would result in a very large amount of time spent evicting the same buffer entries multiple times.

### DIFF
--- a/src/streams/register_nack.rs
+++ b/src/streams/register_nack.rs
@@ -118,12 +118,17 @@ impl NackRegister {
         };
 
         // reset packets that are rolling our of the nack window
-        for s in *active.start..*start {
+        for (i, s) in (*active.start..*start).enumerate() {
             let p = self.packet(s.into());
             if !p.received && s != *seq {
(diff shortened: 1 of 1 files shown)


src/streams/register_nack.rs
```diff
@@ -275,6 +280,18 @@ mod test {
...
+    #[test]
+    fn nack_test_huge_seq_gap_no_hang() {
+        let mut reg = NackRegister::new();
+
+        let before = Instant::now();
+
+        reg.update(0.into());
+        reg.update(18446744073709551515.into());
+
+        assert!(before.elapsed().as_millis() < 1000);
```
algesten, line 292: I really dislike these kinds of tests, because if the CI is slow, you get random errors. And no, it doesn't matter how huge we make 1000, it's still a flakey test because it can fail despite being correct. This is why str0m always drives time from outside and have no internal timers.

algesten (changes requested):

algesten (approved):

PR #433, opened 2024-01-04
Add contiguity check for VP9 codec

h1t:
Check frame contiguity for VP9 codec in non-flexible mode.
Just copy/paste from VP8 contiguity implementation.
Refactored code duplication.


src/packet/vp8_contiguity.rs
```diff
@@ -2,31 +2,29 @@ use super::Vp8CodecExtra;
 use std::collections::VecDeque;
 
 #[derive(Debug)]
-pub struct Vp8Contiguity {
+pub(crate) struct FrameContiguityState {
```
algesten, line 5: Should we maybe move `FrameContiguityState` to packet/contiguity.rs since it's shared for both vp8 and vp9?

algesten (changes requested):
Hi @h1t, welcome to str0m!

This looks great! I had one little question about the code structure.

About the `contiguous_vp9.pcap`, is it not possible to reuse the `vp9.pcap` for this test? If not, I think we should at least make a `contiguous_vp9.txt` similar to the other files. It's an export of the text form using wireshark.

algesten:
Also, @davibe I know you're on holiday, but your input here would be very appreciated since the previous code was your baby.

davibe:
@algesten I got no objections :)

For context the reason we needed the contiguity check for VP8 is because with network loss Safari (stable) deliberately avoids retransmitting frames that "can be skipped" (no other frame depends on them). In reality sometimes it also avoids retransmitting frames that are needed. The contiguity checks reduces the amount of discontinuities str0m detects and reports in MediaData (which normally results in a keyframe request) and the amount of times our buffer 'holds back' waiting for them (jitter). Ever since we fixed the nack-logic, I no longer think other browser share this behaviour.

Also, Testing Safari Technology preview I have seen it behave "correctly" as other browsers. This is why I thought of this contiguity thing as a mitigation which I hoped to remove when Safari "got fixed". 

I have no data on the behaviour of the different browsers when using VP9.

PR #434, opened 2024-01-05
fix: is_video func missing h265

giangndm:
Current Codec.is_video function is missing H265 type. This PR just add H265 Codec type to is_video function.


lolgesten (approved):

lolgesten:
I've been avoiding h265 until we have proper support for it, but I guess this one doesn't matter.

PR #437, opened 2024-01-06
Switch rand -> fastrand

algesten:


algesten:
@k0nserv you mind look at this PR?

My investigation of the call sites is that we _don't_ require cryptographic randomness at any of these places. To make it explicit I made a rather verbose proxy-struct of the calls.

k0nserv (approved):
Neat, I was concerned that this is introducing a new dependency, but `fastrand` was already in tree

PR #442, opened 2024-01-14
Robust(er) RTX PT allocation for StreamTx

algesten:
[One of our users reports](https://str0m.zulipchat.com/#narrow/stream/377845-general/topic/Getting.20panic.20.22pt_rtx.20resend.2Fblank.22/near/412721167) that we got an `expect()` that trips. Any expect/unwrap that trips means we have a bug.

`/src/streams/send.rs:382:45: pt_rtx resend/blank`

The code in question:

```rs
  NextPacketKind::Resend(_) | NextPacketKind::Blank(_) => {
      let pt_rtx = param.resend().expect("pt_rtx resend/blank");
```

Whenever we have a `Resend` or `Blank` packet, we assume the RTX PT is known. This is sometimes (apparently) not true. Rather than dwelling on exact root cause, I've rewritten the code in question to have a more robust way of detecting which RTX PT to use. It should also be impossible to produce a `Resend` or `Blank` if the PT isn't known.

### DIFF
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -116,6 +116,9 @@ pub struct StreamTx {
 
     // downsampled rtx ratio (value, last calculation)
     rtx_ratio: (f32, Instant),
+
+    // The PT to use for padding RTX.
+    pt_for_padding: Option<Pt>,
 }
 
 /// Holder of stats.
@@ -179,6 +182,7 @@ impl StreamTx {
             pending_request_remb: None,
             stats: StreamTxStats::default(),
             rtx_ratio: (0.0, already_happened()),
+            pt_for_padding: None,
         }
     }
 
@@ -305,8 +309,8 @@ impl StreamTx {
         Ok(())
     }
 
-    fn rtx_enabled(&self) -> bool {
-        self.rtx.is_some()
+    fn padding_enabled(&self) -> bool {
+        self.rtx.is_some() && self.pt_for_padding.is_some()
     }
 
     pub(crate) fn poll_packet(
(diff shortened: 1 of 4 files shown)


src/streams/send.rs
```diff
@@ -470,6 +475,14 @@ impl StreamTx {
         let seq_no = next.seq_no;
         self.last_used = now;
 
+        // This is set here due to borrow checker.
+        if self.pt_for_rtx != set_pt_for_rtx {
```
k0nserv, line 479: Can we have a comment here that explains why we do this?

k0nserv (approved):

src/streams/send.rs
algesten: Grrr. Writing the code doc, I realise this solution is not good enough either.

If the user switches PT mid session, the resends already queued must be sent with the resend-PT they belong to.

algesten: Well. Sort of. I'll make it so we don't enqueue any resend requests for a PT that doesn't have an RTX PT.

algesten:
@k0nserv I tweaked this PR so that even in the face of a changing main PT mid-stream, we are nacking on the correct PT RTX.

src/streams/send.rs
```diff
@@ -470,25 +489,32 @@ impl StreamTx {
         let seq_no = next.seq_no;
         self.last_used = now;
 
+        // Padding comes in two forms, "spurious resends" of sent packets where
```
k0nserv, line 492: Excellent!

Issue #413, opened 2023-12-07
Standalone ICE agent?


thomaseizinger:
I am back with another inquiry! :)

Currently, I am trialing usage of str0m's `IceAgent` in a project that just needs ICE but nothing else around WebRTC. In particular, we don't want to pull in openssl as a dependency. I have a few small patches locally that publicly export a handful APIs around the `ice` module. However, I also have one that pretty much comments out everything that deals with `openssl` which is just a hack for now.

I would like to get some opinions on what a collaborative and useful way forward is. I think a SANS-IO & compliant ICE implementation is something really useful so I don't really want to fork `str0m` longterm. The `ice` module also depends on the `io` and `parser` module which makes extracting it not quite straight-forward. Additionally, splitting up `str0m` into multiple crates creates an additional maintenance burden.

I think extracting a dedicated `str0m-ice` crate (name to be bike-shedded) would make the most sense:

- The dependency on `parser` could be avoided by removing `Candidate::from_sdp_string`. This one is actually a circular dependency so could be argued shouldn't exist in the first place.
- The dependency on `io` is a bit trickier. To do message parsing, [`stun_codec`](https://github.com/sile/stun_codec) could be used. I've had good experiences with it. Alternatively, something like `str0m-stun` would have to be introduced that handles message parsing.
- The rest of IO, i.e. all the SANS-IO modelling would either also have to go into a dedicated crate or perhaps inlined into `str0m-ice`? 

Curious to hear your thoughts!

thomaseizinger:
Thank you for the comment!

> Related to that, str0m's ICE isn't strictly RFC compliant – it's compatible. A [simplifying assumption](https://github.com/algesten/str0m/blob/main/docs/ice.md?rgh-link-date=2023-12-07T20%3A31%3A15Z#one-to-one-relationship-ice-agent---session---data-stream) is that there is one agent instance per peer connection, and there is only a single stream/component to negotiate.

This is okay for us. We don't even want any streams, I just want to use ICE to hole-punch a connection!

> We should consider the long term goals of such crate. Full RFC compliance? Line-by-line translation of libWebRTC? Nothing more than we have?

I'd be pragmatic here and suggest that to begin with, it can be whatever is useful to you and whichever other projects come along. Full RFC compliance could be a novel goal although I'd favor compatibility over it. Most notably for our usecase, only str0m <> str0m needs to be compatible!

[...]

PR #446, opened 2024-01-21
feat: always use `sha1` crate for sha1-hmac

thomaseizinger:
Currently, `openssl` is used to perform SHA1-HMAC on Windows. The module's comment states that `sha1-asm` does not work on Windows which is true. It would only work on the `windows-gnu` toolchain but that one breaks `openssl`. If https://github.com/RustCrypto/asm-hashes/issues/45 were resolved, we could also compile `sha1` asm on `windows-msvc`.

Simply disabling the `asm` feature on Windows seems to be the best trade-off. This makes things slower on Windows.

As part of this refactoring, I designed the new `sha1-hmac` functionality to be stateless which ended up deleting the `RtpHmac` trait. This trait was only implemented on one type which suggests that it wasn't a very good use of a `trait` anyway.

This is the first step in making the `openssl` dependency optional as discussed in https://github.com/algesten/str0m/issues/413.


thomaseizinger:
> Simply disabling the `asm` feature on Windows seems to be the best trade-off. This makes things slower on Windows.

@algesten Is that acceptable? I think the proper fix for this performance regression would be to do https://github.com/RustCrypto/asm-hashes/issues/45 for the `sha1` crate. I have not written assembly code before, so perhaps somebody with more experience on that front could tackle this and restore the performance after merging this?

I am also open to other suggestions but I think we will have to stop using `openssl` for `sha1-hmac` in order to make it entirely optional.

src/rtp/srtp.rs
```diff
@@ -147,13 +146,16 @@ impl SrtpContext {
             #[cfg(feature = "_internal_test_exports")]
             Derived::PassThrough => input.to_vec(),
             Derived::Aes128CmSha1_80 {
-                hmac, salt, enc, ..
+                hmac_key,
```
algesten, line 149: How about calling it just `key`, so we can avoid all these destructures to format vertically?

algesten (approved):
Thanks!

Overall LGTM. My only issue is a code style one with all these vertical formattings.

algesten:
@xnorpx since performance on Windows is very important to you guys, can you weigh in here?

src/rtp/srtp.rs
thomaseizinger: Okay!

xnorpx:
@algesten we are currently only running str0m on Linux servers, while we would like to prototype clients on Windows is more long term. 

Even so, not having the asm optimizations would not be a blocker to get running and compiling and the cargo.toml comments point that there is optimization improvement if the issue is resolved.

So I think it's fine to remove asm for now on Windows and unblock for the feature work of making openssl a feature.

algesten:
Great! Thanks!

src/crypto.rs
```diff
@@ -0,0 +1,13 @@
+pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
+    use hmac::Hmac;
+    use hmac::Mac;
+    use sha1::Sha1;
+
+    let mut hmac = Hmac::<Sha1>::new_from_slice(key).unwrap();
```
xnorpx, line 6: this unwrap will never hit?

thomaseizinger: The same `unwrap` was present in the previous `From` impl for `Sha1`. It is an internal API, let me see what we can do about it!

algesten: It's ok to unwrap internally, but there should be a comment of why it is fine.

thomaseizinger: Interesting, I think this can be hit if the provided STUN password is not 20 bytes long. @algesten Any input on this?

thomaseizinger: No, `Hmac` normalizes the key to 20 bytes: https://github.com/RustCrypto/MACs/blob/1a674e4c162222468b13c8b87265f4eda52b6531/hmac/src/lib.rs#L113.

thomaseizinger:
> @algesten we are currently only running str0m on Linux servers, while we would like to prototype clients on Windows it is more long term.

FWIW, we will be running it on Windows but I don't think SHA1 performance is going to be a bottleneck for us next to Wireguard! We are also only using ICE from str0m so only sending a few STUN messages to establish a connection.

thomaseizinger:
Awesome, thank you both for the quick review turnaround! :)

algesten (approved):
Looks great. Let's merge!

PR #447, opened 2024-01-22
feat: make openssl optional

thomaseizinger:


thomaseizinger:
> some of the crypto but that could also be replaced with pure Rust crates and/or `ring` I think

The RustCrypto org offers [`aes-gcm`](https://github.com/RustCrypto/AEADs/tree/master/aes-gcm) but not `aes-cm` because I guess it is too old / insecure? `ring` also offers only `aes-gcm`: https://docs.rs/ring/latest/ring/?search=aes

xnorpx:
Openssl needs to be a feature for SRTP as well.

Same reasons as stated before:
- Organisations have inflexible dependencies and Ring might not be allowed while Openssl is allowed.
- Performance, encryption and decryption is one of the more expensive parts when running Video/Audio RTP streams. Moving to a less optimized implementation would increase cost unless Ring has same or better optimizations than Openssl.

src/dtls/mod.rs
```diff
@@ -317,6 +321,184 @@ impl Dtls {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "KeyingMaterial")
+    }
+}
+
+#[derive(Debug, Clone)]
+pub(crate) enum Certificate {
+    #[cfg(feature = "openssl")]
+    Openssl(ossl::DtlsCert),
+}
```
algesten, line 352: I'm in two minds about calling this `Certificate` as long as it's hard tied to DTLS. I guess there's nowhere else in WebRTC we have certificates? 🤔 

I would feel more comfortable staying with `DtlsCert` on the top level and rename the other one `OsslCert` or something.

src/dtls/mod.rs
```diff
@@ -317,6 +321,184 @@ impl Dtls {
...
+
+impl Certificate {
+    #[cfg(feature = "openssl")]
+    pub(crate) fn new_openssl() -> Certificate {
+        Certificate::Openssl(ossl::DtlsCert::new())
+    }
+}
+
+enum Context {
```
algesten, line 361: `Context` is too generic here. I think we need `DtlsContext` or maybe `TlsContext`

algesten (changes requested):
Some naming comments.

Otherwise looks ok so far.

src/dtls/mod.rs
thomaseizinger: I've renamed it to `DtlsCert`. I've left the `ossl::DtlsCert` name because we can just fully-qualify it to differentiate. The enum is currently not exposed so users only see `ossl::DtlsCert`.

src/dtls/mod.rs
thomaseizinger: I mean, we are within the `dtls` module here and this is private so it is a bit repetitive but I'll rename it :)

algesten: Ha! :)

thomaseizinger:
Ready for review! The last commit is kind of large. I could probably split that up into more commits if you really wanted.

algesten:
Looking at this code, it strikes me we might be needing some more/other abstraction here.

`TlsStream` exists very out of necessity. I couldn't find a way to use OpenSSL API (in Rust) to make a Sans-IO (poll/push style) API, instead it relies on streams (`std::io::Read`, `std::io::Write`). `TlsStream` is thus kind of a "hack" to turn the OpenSSL API into a push/pull.

My preferred abstraction for the concrete DTLS impls would look more like how `Dtls` presents itself towards `Rtc` (pretty much all `pub fn` on `Dtls`). That mean `TlsStream` would not need to be part of the abstraction, but rather an implementation detail of OpenSSL.

I guess I have a very specific idea in mind here, and you've made a lot of work already. If you rather not continue, I totally understand. I can continue in that case.

thomaseizinger:
> My preferred abstraction for the concrete DTLS impls would look more like how `Dtls` presents itself towards `Rtc` (pretty much all `pub fn` on `Dtls`). That mean `TlsStream` would not need to be part of the abstraction, but rather an implementation detail of OpenSSL.
> 
> I guess I have a very specific idea in mind here, and you've made a lot of work already. If you rather not continue, I totally understand. I can continue in that case.

This is the last item that blocks us from depending on upstream `str0m` and I am very much looking forward to discontinuing our fork!

If you consider the current state okay to be merged then I'd prefer we do that and improve on the code later on! :)

thomaseizinger:
I think what the right abstraction is will likely only show once you add another crypto implementation. Until then, it is a lot of guess-work!

algesten:
@thomaseizinger I've redone some of the work in this PR in #449 going further in moving/teasing apart the OpenSSL from the main code.

SRTP was surprisingly the most work.

thomaseizinger:
Okay, going to close this then!

PR #449, opened 2024-01-23
Refactor openssl to crypto

algesten:
- Reword expect
- Remove `ToRtpIv` trait for `aes_128_cm_sha1_80`
- Remove `ToRtpIv` trait for `aead_aes_128_gcm`
- Move `KeyingMaterial` out of ossl module
- Add CI check for `--no-default-features`
- crypto.rs -> crypto/mod.rs
- Refactor OpenSSL DTLS into crypto mod
- Feature gate openssl
- Move things around in crypto
- Refactor our SRTP crypto
- Feature gate all openssl


src/crypto/ossl/dtls.rs
```diff
@@ -0,0 +1,165 @@
...
+use super::io_buf::IoBuffer;
+use super::stream::TlsStream;
+use super::CryptoError;
+
+const DTLS_CIPHERS: &str = "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
+const DTLS_EC_CURVE: Nid = Nid::X9_62_PRIME256V1;
+
+pub struct OsslDtlsImpl {
+    /// Certificate for the DTLS session.
+    _cert: OsslDtlsCert,
```
k0nserv, line 22: Does this need to be kept alive?

src/crypto/ossl/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+
+mod dtls;
+pub use dtls::OsslDtlsImpl;
+
+mod srtp;
+pub use srtp::OsslSrtpCryptoImpl;
+
+impl SrtpProfile {
+    // All the profiles we support, ordered from most preferred to least.
+    pub(crate) const ALL: &'static [SrtpProfile] =
```
k0nserv, line 19: This and `keying_material_len` isn't OSSL specific and can be lifted to `super`

src/_internal_test_exports/fuzz.rs
```diff
@@ -4,7 +4,8 @@ use std::time::Duration;
 use std::time::Instant;
 
 use crate::change::{SdpAnswer, SdpOffer};
-use crate::dtls::{KeyingMaterial, SrtpProfile};
+use crate::crypto::KeyingMaterial;
```
xnorpx, line 7: merge imports like the others?

src/crypto/dtls.rs
```diff
@@ -0,0 +1,176 @@
...
+    Data(Vec<u8>),
+}
+
+/// Certificate used for DTLS.
+#[derive(Clone)]
+pub struct DtlsCert(DtlsCertInner);
+
+#[derive(Debug, Clone)]
+pub enum DtlsCertInner {
+    #[cfg(feature = "openssl")]
```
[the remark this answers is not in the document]
lolgesten: No. Features are additive, so the crypto backends will be additive. We will eventually have `Rtc::new_openssl` or similar to select backend. I.e. config.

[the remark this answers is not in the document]
k0nserv: Mutually exclusive is achievable, but I guess that's needlessly restrictive. I don't see why anyone would want two different backends enabled but what do I know :D

src/crypto/ossl/mod.rs
lolgesten: Alright

src/crypto/dtls.rs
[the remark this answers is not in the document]
lolgesten: It's a matter of principle. I don't like libs that change behavior depending on feature flags.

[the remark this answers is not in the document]
k0nserv: How do you mean change behaviour?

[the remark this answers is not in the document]
lolgesten: Which crypto backend is in use, can be seen as behavior. Some people _must_ use openssl, other _must_ use nativetls, etc (this is from experience in ureq).

I think the API for this is best left as configuration options in the library, i.e. `new_openssl`, which would hard fail if the dependency doesn't have openssl. I prefer that over `new` which magically changes underlying impl depending on flags.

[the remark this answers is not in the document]
k0nserv: Okay

src/_internal_test_exports/fuzz.rs
lolgesten: Haven't you seen my other PR where I threaten to make all imports like this? :)

Tbh, we are not having consistent imports at this point. I guess I'm ruined by Java, but I hate with a passion imports that break the use over several lines.

```
use foo:{
  blah blah blah
  blah blah blah
}
```

I can live with some being individual and some not though...

src/crypto/dtls.rs
```diff
@@ -0,0 +1,176 @@
...
+    /// Decrypted data from incoming DTLS traffic.
+    Data(Vec<u8>),
+}
+
+/// Certificate used for DTLS.
+#[derive(Clone)]
+pub struct DtlsCert(DtlsCertInner);
+
+#[derive(Debug, Clone)]
+pub enum DtlsCertInner {
```
thomaseizinger, line 40: Make this `pub(crate)` to be safe?

src/crypto/finger.rs
```diff
@@ -0,0 +1,51 @@
...
+    /// Hash function used to produce the `bytes`.
+    ///
+    /// This is normally `sha-256`.
+    pub hash_func: String,
+
+    /// Digest of the certificate by the algorithm in `hash_func`.
+    pub bytes: Vec<u8>,
+}
+
+impl ToString for Fingerprint {
```
thomaseizinger, line 16: I think it is recommended to implement `fmt::Display` instead and rely on the blanket impl to get `ToString`.

thomaseizinger: See https://doc.rust-lang.org/std/string/trait.ToString.html.

src/crypto/srtp.rs
```diff
@@ -0,0 +1,251 @@
...
+pub fn srtp_aes_128_ecb_round(key: &[u8], input: &[u8], output: &mut [u8]) {
+    #[cfg(feature = "openssl")]
+    {
+        super::ossl::OsslSrtpCryptoImpl::srtp_aes_128_ecb_round(key, input, output)
+    }
+    #[cfg(not(feature = "openssl"))]
+    {
+        panic!("No SRTP implementation. Enable openssl feature");
+    }
+}
```
thomaseizinger, line 76: I am surprised to see this here when you've (correctly) advocated for additive features previously. How is this going to evolve in the future?

src/crypto/dtls.rs
```diff
@@ -0,0 +1,176 @@
...
+    fn handle_input(&mut self, data: &[u8]) -> Result<(), CryptoError>;
+
+    /// Whether the DTLS connection is established.
+    fn is_connected(&self) -> bool;
+}
+
+pub enum DtlsImpl {
+    #[cfg(feature = "openssl")]
+    OpenSsl(super::ossl::OsslDtlsImpl),
+}
```
thomaseizinger, line 114: What is the point of the `DtlsInner` trait if you are doing enum-based dispatch here? Isn't this even more LoC compared to having:

```rust
pub struct DtlsImpl(Box<dyn DtlsInner>);
```

YMMV but unless you have a `Box<dyn MyTrait>` or `T: MyTrait` somewhere, you don't actually need a trait so I'd argue you shouldn't have it.

src/crypto/srtp.rs
```diff
@@ -0,0 +1,251 @@
...
+            SrtpProfile::AeadAes128Gcm   => 16 * 2 + 12 * 2,
+        }
+    }
+}
+
+#[allow(unused)]
+
+pub fn new_aes_128_cm_sha1_80(
+    key: AesKey,
+    encrypt: bool,
```
thomaseizinger, line 38: This would be nicer as an enum to avoid the "boolean trap":

```rust
enum Mode {
	Encrypt,
	Decrypt
}
```

TBH, I am not sure it adds much value. Personally, I'd prefer non-branching functions, even if it adds a few LoC.

src/lib.rs
```diff
@@ -1016,14 +1017,23 @@ impl Rtc {
...
+            c
+        } else {
+            #[cfg(feature = "openssl")]
+            {
+                DtlsCert::new_openssl()
+            }
+            #[cfg(not(feature = "openssl"))]
+            {
+                panic!("No DTLS implementation. Enable openssl feature");
+            }
```
thomaseizinger, line 1030: Why not just feature-gate the function? That would be much cleaner.

thomaseizinger (approved):
Left some comments as food for thought! Nothing blocking :)

src/crypto/srtp.rs
algesten: It's a TBD. Passing the openssl config RtcConfig -> SrtpContext felt a bit too far for me to think about today. But you're right, I should leave a comment here.

src/crypto/dtls.rs
algesten: It's about establishing the contract. I find it easier to reason about the internal API when I see it like that. But it's not strictly necessary. It's following the pattern we have for Packetizer/Depacketizer.

k0nserv: Dynamic dispatch is probably less performant than static too. We don't need the flexibility of dynamic dispatch here

src/crypto/srtp.rs
algesten: I did it like this because I felt there were already too many `#[cfg(feature = "openssl")]` so from `rtp/srtp.rs` the contract is currently 3 functions (instead of 5) which ideally I'd reduce even further.

I don't quite see a good pattern here though.

src/crypto/finger.rs
algesten: The format here is very specific (to fit in an SDP). The Display trait won't be right either, we should change this to a function.

src/crypto/dtls.rs
algesten: It's private even!

src/crypto/dtls.rs
thomaseizinger: > Dynamic dispatch is probably less performant than static too. We don't need the flexibility of dynamic dispatch here

Yeah absolutely. I am not arguing for dynamic dispatch, I am arguing against using language features (traits) when they are not needed. Same reason as preferring regular functions over macros.

> It's about establishing the contract.

Aren't the public functions of a type / module a good enough contract? Other modules don't see the internal APIs anyway so I am not sure what exactly is achieved with this rigid enforcement other than adding layers of indirection when navigating the code.

thomaseizinger:
Thank you for acting on this so quickly! :tada:

algesten:
No worries. I want you to get real world usage out of the ICE Agent, so we can start improving that part together! 😍

PR #451, opened 2024-01-23
chore: re-export `StunPacket`

thomaseizinger:
Unless I am missing something, `StunPacket` isn't re-exported anywhere so the current `IceAgent` API cannot be used from outside the crate.


algesten (approved):
Ah yea. This is an omission.

Thanks!

Issue #453, opened 2024-01-27
How to correctly handle `Transmit::source` and `Candidate::base`?


thomaseizinger:
In the connectivity library I built on top of `str0m`'s `IceAgent`, I copied its model of `Transmit` and return those messages to the upper layer. I also handle TURN allocations and channels for the user.

Here comes the catch: When `str0m` wants to make a STUN binding request _from_ one of my relay candidates, it sets `Transmit::source` to the allocation address. This is a result of `Candidate::base` being set to the same value as the actual addr in the constructor:

https://github.com/algesten/str0m/blob/30ba27e4fb683d45559e5a821c0c2c084e02463a/src/ice/candidate.rs#L216

This is actually useful! It allows me to cross-reference `Transmit::source` with the list of TURN servers and if I detect that I have a TURN server for this IP, I know that I need to wrap the `Transmit::contents` in a channel data message to `Transmit::destination`.

The issue gets trickier for server-reflexive addresses. Here, `Transmit::source` is set to the server-reflexive address. Without keeping a mapping myself, I therefore don't know which _local_ interface this packet needs to be sent from. My first thought was that we should simply change the `Candidate` constructors to accept an additional `base` parameter that designates the local socket the data was received on that generated this candidate. However, this will break the case for relay candidates because I now have no way of knowing, whether the data needs to be sent through a relay.

So far, I've come up with the following solutions and I don't like any of them:

- Extend the `Candidate` constructors with `base` and set it correctly BUT also change `Transmit` to always use `addr`. Then, whenever processing a `Transmit` from str0m, one needs to iterate the local candidates, find the corresponding one and check the `base` for server-reflexive address candidates to know which socket to use. For host candidates, `addr` works just fine and for relay candidates, one can look up the mapping similar as to today.
- Extend the `Candidate` constructors with `base` and split `Transmit::source` into two fields:
	- `final_source` / `perceived_source`: The `source` that the remote should see. This is only useful for relayed candidates.
	- `local_source`: The local socket to use when sending the data.
- Extend only the server-reflexive `Candidate` constructors with `base` and keep returning the actual allocation socket on the relay as `base` for relay candidates.

The last one is the easiest but also the most confusing, especially because it is hard to understand and the type system doesn't guide you. Perhaps a variation of the last one would be to model this explicitly by introducing a `Source` enum that is:

```rust
pub enum Source {
	/// The data should be sent from the specified, local socket.
	Local(SocketAddr),
[...]

PR #455, opened 2024-01-30
feat: explicitly set `base` of server-reflexive addresses

thomaseizinger:
Without further adjustments, this will set `Transmit.source` to the provided base here. Opening this here because we are (temporarily) depending on it.

Resolves: #453.

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -177,8 +177,10 @@ impl Candidate {
     ///
     /// Server reflexive candidates are local sockets mapped to external ip discovered
     /// via a STUN binding request.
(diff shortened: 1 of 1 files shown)


algesten:
Thanks. I need to circle back to this later this week. Did a mega-deploy at work today, so it's been busy.

thomaseizinger:
> Thanks. I need to circle back to this later this week. Did a mega-deploy at work today, so it's been busy.

No worries at all. If you want to see how it is used, check the linked PR :)
The gist is: Check if `src` is a TURN server IP, if yes, wrap in channel-data message, otherwise forward. This now works because `base` for srflx addresses is the local host candidate.

PR #456, opened 2024-02-01
Document Transmit.source

algesten:


algesten:
@thomaseizinger do you agree with this documentation?

src/io/mod.rs
```diff
@@ -66,6 +66,22 @@ pub struct Transmit {
...
+    ///     [`Candidate::host()`][crate::Candidate::host]
+    /// * `srflx` - The `base` parameter in
+    ///     [`Candidate::server_reflexive()`][crate::Candidate::server_reflexive]
+    /// * `relayed` - The address provided in
+    ///     [`Candidate::relayed()`][crate::Candidate::relayed]
+    /// * `prflx` - Two variants:
+    ///     * STUN request sent by other peer – Use the source address of the incoming packet
+    ///     * STUN response to one of str0m's requests – Use the address we used to send the
+    ///         original STUN request from. Thus it can be any of the provided
+    ///         `host`, `srflx` or `relayed`.
```
thomaseizinger, line 84: This is technically correct because it describes how we figure out what to put in `source`. However, if I put my user-hat on, then the first question that comes to my mind after reading this is: "So what should I do with `source`?"

As much as I really like SANS-IO, one of its weaknesses is that the user can make mistakes in what data is fed into the state machine or how data emitted is being interpreted. As such, I think we should place special emphasis on the documentation of these API aspects to make it really clear how to _correctly_ handle this data.

I see `Transmit` as a "command", i.e. `str0m` tells me "do this for me". With the current documentation, it isn't immediately clear to me, what I should do with `source`.

What do you think of the following:

- Rename `source` to `socket`
- Update the documentation to something like:

```rust
/// The socket this packet should be sent from.
///
/// For ICE, it is important to send outgoing packets from the correct socket.
/// Features like hole-punching will only work if the packets are routed through the correct interfaces.
///
/// This address will either be:
/// - The address of a socket you have bound locally, such as with [`UdpSocket::bind`].
/// - The address of a relay socket that you have [allocated](https://www.rfc-editor.org/rfc/rfc8656#name-allocations-2) using TURN.
///
/// To correctly handle an instance of [`Transmit`], you should:
///
/// - Check if [`Transmit::source`] corresponds to one of your local sockets, if yes, send it through that.
/// - Check if [`Transmit::source`] corresponds to one of your relay sockets (i.e. allocations), if yes, send it via one of:
///     - a TURN channel data message: https://www.rfc-editor.org/rfc/rfc8656#name-sending-a-channeldata-messa
///     - a SEND indication: https://www.rfc-editor.org/rfc/rfc8656#name-send-and-data-methods
///
/// `str0m` learns about these sockets via the [`Candidate`]s that are given to the [`IceAgent`]:
///
/// - [`Candidate::host`]: The socket is the candidate itself.
/// - [`Candidate::relayed`]: The socket is the candidate itself.
/// - [`Candidate::server_reflexive`]: The socket is the `base` address you provided to the candidate. This should be the address of the socket that you received the reply on from the STUN server.
/// - [`Candidate::peer_reflexive`]: This one is automatically handled within `str0m`.
///
/// In summary this means you must feed `str0m` the correct local address of the data that a packet was received on. This concerns the following APIs:
///     - [`Candidate::server_reflexive`]
///	- [`net::Receive`]
///	- [`ice::StunPacket`]
pub socket: SocketAddr,
```

algesten:
@thomaseizinger Thanks! You're totally right making this more about from the user's perspective.

I've incorporated your suggestions into mine, but sticking with `source` for now.

thomaseizinger:
Reads great, a big improvement! :)

PR #458, opened 2024-02-07
feat: allow user to trigger ICE restart upon invalidating candidate from nominated pair

thomaseizinger:
Currently, invalidating a candidate only returns `true` when we successfully invalidated the candidate. That isn't particularly helpful though. What I really want to know is whether the candidate was part of the currently nominated pair. In that case, I want to trigger an ICE restart to restore the connection.

It is temping to say: "We should automatically trigger an ICE restart for the user". Being a library though, I think `str0m` should refrain from such a _policy_ and instead stick to _mechanisms_. We expose the information that the candidate was part of the nominated pair. If desired, the user can immediately call `ice_restart` as a result.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -10,6 +10,7 @@ use crate::util::NonCryptographicRng;
 
 use super::candidate::{Candidate, CandidateKind};
 use super::pair::{CandidatePair, CheckState, PairId};
+use super::IceError;
 
 /// Timing advance (Ta) value.
 ///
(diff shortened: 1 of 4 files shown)


src/ice/agent.rs
```diff
@@ -678,9 +679,9 @@ impl IceAgent {
     /// This is done for host candidates disappearing due to changes in the network
     /// interfaces like a WiFi disconnecting or changing IPs.
     ///
-    /// Returns `true` if the candidate was found and invalidated.
+    /// Returns `Ok()` if we invalidated a candidate or `Err` if the candidate was unknown to us.
```
thomaseizinger, line 682: There are several ways of modelling this API. I am not particularly set on this one. It felt idiomatic and I wanted to get some feedback on the overall idea.

thomaseizinger:
Closing because we concluded that this is not necessary.

PR #459, opened 2024-02-13
fix: invalidate nominated pair if we invalidated one of its candidates

thomaseizinger:
I didn't manage to write a test for this but I think it is an artifact of the test harness which "simply" flips `source` and `destination` and thus still allows traffic from `1.1.1.1` despite us invalidating that candidate.


thomaseizinger:
@algesten I think there may still be a bug here that needs addressing at some point. It is not critical for us because the ICE timeout is good enough to detect failed connections.

Issue #460, opened 2024-02-16
Panic in `Rtc:accepts()` caused by ICE code


OxleyS:
Under certain circumstances which I am not totally sure of, the `Rtc::accepts()` function panics:
```
thread '<unnamed>' panicked at /Users/sean/.cargo/registry/src/index.crates.io-6f17d22bba15001f/str0m-0.4.1/src/ice/agent.rs:324:18:
Remote ICE credentials
stack backtrace:
   0: rust_begin_unwind
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/std/src/panicking.rs:645:5
   1: core::panicking::panic_fmt
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:72:14
   2: core::panicking::panic_display
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:196:5
   3: core::panicking::panic_str
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/panicking.rs:171:5
   4: core::option::expect_failed
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/option.rs:1980:5
   5: core::option::Option<T>::expect
             at /rustc/07dca489ac2d933c78d3c5158e3f43beefeb02ce/library/core/src/option.rs:894:21
   6: str0m::ice::agent::IceAgent::stun_credentials
             at /Users/sean/.cargo/registry/src/index.crates.io-6f17d22bba15001f/str0m-0.4.1/src/ice/agent.rs:321:24
   7: str0m::ice::agent::IceAgent::accepts_message
             at /Users/sean/.cargo/registry/src/index.crates.io-6f17d22bba15001f/str0m-0.4.1/src/ice/agent.rs:778:29
   8: str0m::Rtc::accepts
             at /Users/sean/.cargo/registry/src/index.crates.io-6f17d22bba15001f/str0m-0.4.1/src/lib.rs:1495:20
<...my project's stack frames omitted...>
```

The offender seems to be the `stun_credentials()` call, which is documented to panic on missing remote credentials, but this is not checked for before calling.

Unfortunately, I don't think I can provide repro code for you anytime soon, as I was in the process of smashing out a proof-of-concept and the current code is in no position to be reduced.

OxleyS:
[...]

PR #461, opened 2024-02-19
Fix panic in Rtc::accepts() with STUN traffic

OxleyS:
Fixes https://github.com/algesten/str0m/issues/460

As a summary of that discussion, if you have two `Rtc` instances, one negotiated and the other not, and you pass incoming STUN traffic intended for the negotiated instance to the `Rtc::accepts()` of the non-negotiated instance, it will panic in `stun_credentials` because without an offer, no remote ICE credentials will have been sent. This PR simply checks for this case first.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -797,6 +797,12 @@ impl IceAgent {
             }
         }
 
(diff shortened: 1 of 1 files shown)


algesten (approved):
Looks good! Thanks!

PR #462, opened 2024-02-19
fix: use a `const fn`  to compute the overall STUN timeout

thomaseizinger:


algesten (approved):
LGTM!

Thanks!

thomaseizinger:
I pushed another version where I am just testing the constant. Less invasive.

What do you prefer?

algesten:
Neat.

I did like the test in the previous as well :)

thomaseizinger:
I am leaning towards just testing the constant until we get better const arithemtic support :)

algesten:
Thanks

PR #463, opened 2024-02-26
fix: ensure newly discovered pair as a better priority before nominating it

thomaseizinger:


src/ice/mod.rs
```diff
@@ -443,6 +455,53 @@ mod test {
...
+            }) {
+                break;
+            }
+
+            progress(&mut a1, &mut a2);
+        }
+
+        assert!(!a1.has_event(|e| {
+            matches!(e, IceAgentEvent::NominatedSend { source, .. } if source == &sock("1.1.1.1:1001"))
+        }));
```
thomaseizinger, line 499: I didn't manage to have the test hit this assertion before fixing it behaviour but I think that may be related how we process candidates for controlled vs controlling agents. I didn't look into it too deeply.

I think it is good to have here to avoid regressions.

src/ice/mod.rs
```diff
@@ -443,6 +455,53 @@ mod test {
...
+    #[test]
+    pub fn candidate_pair_of_same_kind_does_not_get_nominated() {
+        let mut a1 = TestAgent::new(info_span!("L"));
+        let mut a2 = TestAgent::new(info_span!("R"));
+
+        let c1 = relay("1.1.1.1:1000", "udp");
+        a1.add_local_candidate(c1.clone());
+        a2.add_remote_candidate(c1);
+        let c2 = srflx("4.4.4.4:1000", "3.3.3.3:1000", "udp");
+        let c3 = host("3.3.3.3:1000", "udp");
```
thomaseizinger, line 467: I didn't manage to reproduce this using `host` candidates alone so I re-built the scenario that we were seeing in production: relay candidates talking to srflx candidates and discovering new relay candidates.

lolgesten: Nice to have a real world scenario here!

src/ice/agent.rs
```diff
@@ -1410,48 +1410,39 @@ impl IceAgent {
...
-            } else {
-                trace!("Nominating best candidate");
             }
+            trace!("Nominating best candidate");
 
             let pair = self
                 .candidate_pairs
                 .iter_mut()
-                .find(|p| p.id() == best)
+                .find(|p| p.prio() == best_prio)
```
thomaseizinger, line 1445: Is it guaranteed that this finds the one we also compared above? I think it would be better to still search by ID here.

thomaseizinger (commented):
YMMV but I find the `max_by_key` solution cleaner. Finding the pair again by prio which isn't unique sound like a bug with a ticking timer on it :)

src/ice/agent.rs
lolgesten: It works on the assumption that the exact which of the highest prio pairs doesn't matter.

lolgesten:
You might be right, but I don't like the combinatorial logic here. Doubling up on that rather complicated `.filter` – hence I tried to make something easier to follow along.

thomaseizinger:
> You might be right, but I don't like the combinatorial logic here. Doubling up on that rather complicated `.filter` – hence I tried to make something easier to follow along.

Yeah fair. Instead of `.map` and `max`, can we use `max_by` instead and thus retain the `CandidatePair` as the returned type?

algesten (approved):
This looks good now. Let's ship it!

PR #470, opened 2024-03-03
Fix `ChannelId` allocation

altonen:
Previously `ChannelHandler` used the number of active channels to derive the next `ChannelId`. This resulted in two channels getting assigned the same `ChannelId` because the allocation didn't track which `ChannelId`s were free to be reused. For example, when two channels were created, allocating channel IDs 0 and 1, `ChannelId(0)` was freed and then a new channel was created, this new channel had an ID 1, which was already in use.

Start using a `usize` counter to allocate channel IDs, ensuring they're unique

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -99,6 +99,7 @@ impl fmt::Debug for ChannelData {
 #[derive(Debug, Default)]
 pub(crate) struct ChannelHandler {
     allocations: Vec<ChannelAllocation>,
+    next_channel_id: usize,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


altonen:
@algesten friendly ping

does it look good to you or is there some blocker?

algesten:
Can we make a simpler version that just keeps increasing the number? No Reuse?

altonen:
> Can we make a simpler version that just keeps increasing the number? No Reuse?

Done

algesten (approved):
LGTM!

Thanks!

PR #481, opened 2024-03-13
chore: remove superfluous "ICE restart" log

thomaseizinger:
When calling `IceAgent::ice_restart`, we always see the following logs:

```
2024-03-13T23:54:12.115048Z  INFO str0m::ice_::agent: ICE restart
2024-03-13T23:54:12.115057Z  INFO str0m::ice_::agent: State change (ice restart): Completed -> Checking
```

This PR removes the first one because it is superfluous.


algesten:
LGTM!

PR #482, opened 2024-03-16
Update sctp-proto for panic fix

xnorpx:
Found a bug in sctp-proto: https://github.com/webrtc-rs/sctp-proto/issues/8

It's fixed so let's update.


algesten (approved):
Merge!

PR #484, opened 2024-03-19
Revert #461 and add a regression test for what it broke.

OxleyS:
While the bug mentioned in https://github.com/algesten/str0m/pull/461 was real as of version `0.4.1`, it had since been fixed by https://github.com/algesten/str0m/pull/418, rendering my fix redundant.

My fix also caused a regression - STUN binding requests received before remote credentials were set were being dropped by `IceAgent.accept_message()`, instead of being let through and subsequently buffered via `IceAgent.stun_server_queue`.

Rather than just revert the PR, this also adds a test to catch further regressions of this buffering.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -801,12 +801,6 @@ impl IceAgent {
             }
         }
 
-        // The integrity check below may panic if no remote ICE credentials yet exist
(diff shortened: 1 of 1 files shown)


algesten (approved):
Nice!

Love to have a test for it as well!

PR #487, opened 2024-03-21
Check all remote ICE candidates in `Rtc::accepts()`

OxleyS:
Sometimes, a WebRTC client will send e.g. DTLS traffic using a candidate pair that has been found successful, but is not the currently nominated pair. In my local tests, this only happens as the connection is being established. My guess is a race condition - if multiple pairs are nominated in sequence, it's possible for traffic to come in via an old nomination before the new one reaches the remote client and takes effect.

If this happens, `Rtc::accepts()` will reject the traffic due to not being from the current nominated pair. I noticed this while working with Firefox, but it may happen in other WebRTC clients as well. I also noticed that occasionally Firefox will get stuck using the same not-nominated pair to retry DTLS that failed in this manner (although I imagine such behavior is not intended).

To minimize the performance impact, we continue to eagerly check against the nominated pair, and only fallback to checking all candidates as a last resort.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1473,7 +1473,8 @@ impl Rtc {
     /// [`Input::Timeout`] is always accepted. [`Input::Receive`] is tested against the nominated
     /// ICE candidate. If that doesn't match and the incoming data is a STUN packet, the accept call
     /// is delegated to the ICE agent which recognizes the remote peer from `a=ufrag`/`a=password`
-    /// credentials negotiated in the SDP.
+    /// credentials negotiated in the SDP. If that also doesn't match, all remote ICE candidates are
+    /// checked for a match.
     ///
     /// In a server setup, the server would try to find an `Rtc` instances using [`Rtc::accepts()`].
     /// The first found instance would be given the input via [`Rtc::handle_input()`].
@@ -1502,11 +1503,9 @@ impl Rtc {
             return true;
         };
 
-        // This should cover Dtls, Rtp and Rtcp
+        // Fast path: DTLS, RTP, and RTCP traffic coming in from the same socket address
+        // we've nominated for sending via the ICE agent. This is the typical case
         if let Some(send_addr) = &self.send_addr {
-            // TODO: This assume symmetrical routing, i.e. we are getting
-            // the incoming traffic from a remote peer from the same socket address
-            // we've nominated for sending via the ICE agent.
             if r.source == send_addr.destination {
                 return true;
             }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -1518,6 +1517,14 @@ impl Rtc {
             return self.ice.accepts_message(v);
         }
 
+        // Slow path: Occasionally, traffic comes in on a socket address corresponding
+        // to a successful candidate pair other than the one we've currently nominated.
+        // This typically happens at the beginning of the connection
+        let remote_candidates = self.ice.remote_candidates();
+        if remote_candidates.iter().any(|cand| cand.addr() == r.source) {
+            return true;
+        }
```
algesten, line 1526: Can you make this a function on `IceAgent` to avoid having too much of the logic here in `Rtc`? Something like `has_remote_candidate`.

algesten (changes requested):
A good start.

One small comment.

We also need to think through the security consequences here. Does this mean we accept traffic for remote candidates that have not been verified with a STUN request/response? Does that matter? The SCTP and RTP traffic will be protected by the DTLS fingerprint, so maybe it's not a big deal 🤔 

Wonder how libWebRTC does this? Maybe they only accept unverified remote candidates for a short while?

OxleyS:
That's true, it would be a good idea to sanity check candidates first. I can also try digging into libwebrtc source a bit and see if I find anything.

OxleyS:
I digged into libwebrtc source a bit. It seems they have [similar logic for deciding what to forward](https://chromium.googlesource.com/external/webrtc.git/+/refs/heads/main/p2p/base/p2p_transport_channel.cc#2201), complete with a fast path for the selected one. Here, `Connection` basically [boils down to a candidate pair](https://chromium.googlesource.com/external/webrtc.git/+/refs/heads/main/p2p/base/connection.h#73), with one [created for every non-pruned local-remote combination](https://chromium.googlesource.com/external/webrtc.git/+/refs/heads/main/p2p/base/p2p_transport_channel.cc#1384).

The "non-pruned" bit may be important to us as well, in the spirit of the security consequences you mentioned. libwebrtc seems to prune candidates whose networks have disappeared, or when a re-gather is initiated. str0m's equivalent seems to be `discarded`?

OxleyS:
We could then take one of a few approaches:
1. Check all non-discarded remote candidates for a matching source address.
2. Check all candidate pairs with `CheckState::Succeeded` for a remote candidate with a matching source address (this automatically excludes discarded candidates).
3. The second solution, but also check the local candidate against the destination address.

Which of these do you feel is most appropriate?

algesten:
I think 2 is a good choice. If it succeeded, we have got a successful request-response. That might match libWebRTC logic here.

algesten (approved):
Like it!

Thanks!

PR #488, opened 2024-03-21
Fix panic on STUN binding failure while multiplexing

OxleyS:
An unrelated bug I had in my application was causing STUN to fail, and a `Class::Failure` message to come back. Because like in the chat example, I'm still using a single multiplexed UDP socket, this message first went to the `Rtc::accepts()` of a separate client that had not negotiated, which caused it to panic with "Remote ICE Credentials".

This is the same underlying panic as in https://github.com/algesten/str0m/pull/461, but this time the logic was allowed to fall through to the `stun_credentials()` call because we were only doing the preceding candidate-pair test for `Class::Success` responses, not `Class::Failure`.

I wanted to add a test for this, but with the current field visibility on `StunMessage`, I couldn't create a failure message in a test. I wasn't sure if adding a helper to `impl StunMessage` was appropriate.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -798,7 +798,7 @@ impl IceAgent {
             }
         }
 
-        if message.is_successful_binding_response() {
+        if message.is_response() {
             let belongs_to_a_candidate_pair = self
                 .candidate_pairs
                 .iter()
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -2,7 +2,7 @@ use std::collections::{HashSet, VecDeque};
 use std::net::SocketAddr;
 use std::time::{Duration, Instant};
 
-use crate::io::{Id, DATAGRAM_MTU_WARN};
+use crate::io::{Id, StunClass, StunMethod, DATAGRAM_MTU_WARN};
 use crate::io::{Protocol, StunPacket};
 use crate::io::{StunMessage, TransId, STUN_TIMEOUT};
 use crate::io::{Transmit, DATAGRAM_MTU};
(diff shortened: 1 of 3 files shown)


algesten:
Which panic are you hitting?

algesten:
I think the correct fix would be:

```rust
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -799,6 +799,9 @@ impl IceAgent {
                 trace!("Message rejected, transaction ID does not belong to any of our candidate pairs");
                 return false;
             }
+        } else {
+            trace!("Message reject, it is not a succesful binding response");
+            return false;
         }
 
         let (_, password) = self.stun_credentials(!message.is_response());
```

OxleyS:
That patch would end up rejecting everything that's not a successful binding response, since all other message types, including things like binding requests, would fall through to that `else` case as well.

I believe the effect you intended was to just reject all non-success binding responses? That would also fix the panic, although arguably `IceAgent` should be accepting all traffic intended for it, even if it doesn't react to failed responses in any special way.

(Aside: `Class::Indication` and `Class::Unknown` wouldn't trigger panics with either of our solutions).

algesten:
Taking a step back.

`accept` should, as you say, accept all traffic targeted at this Rtc instance.

OxleyS:
I think we have our bases covered here then - STUN requests get handled as before, and now both STUN success and failure gets accepted if it corresponds to a STUN request from us (or rejected before the credentials check if not).

algesten:
I still don't like this change.

What kind of responses are we talking about that we suddenly accept? What about middleboxes? We should be very specific with what we accept. Enumerate the cases.

OxleyS:
I believe this enumeration is exhaustive:
1. `Method::Binding`, `Class::Request`: Unchanged, rejects on a local/remote ufrag mismatch and then checks integrity. Cannot panic because it does not require remote credentials to integrity-check.
2. `Method::Binding`, `Class::Success`: Unchanged, rejects if the transaction ID matches no binding request from us, then checks integrity. If the transaction ID matches, the integrity-check would panic without remote credentials, but we do not generate binding requests without remote credentials, so panic is effectively impossible.
3. Any method, `Class::Failure`: *Changed*, previous behavior would only check integrity, panicking if remote credentials have not been set. New behavior matches the `Class::Success` case.
4. `Method::Unknown`, `Class::Success`: *Changed*, both previous and new behavior match the `Class::Failure` case.
5. `Method::Unknown`, `Class::Request`: Unchanged, only checks integrity. Cannot panic because it does not require remote credentials to integrity-check.
6. Any method, `Class::Indication` or `Class::Unknown`: Unchanged, only checks integrity. Cannot panic because it does not require remote credentials to integrity-check.

This PR was aimed at fixing the (`Method::Binding`, `Class::Failure`) case specifically, but it seems both (`Method::Unknown`, `Class::Success`) and (`Method::Unknown`, `Class::Failure`) had the same panic bug. This PR ~unintentionally~ incidentally prevents these two from panicking as well, by subjecting them to the same transaction-ID-match test as a binding success would be. It's hard to have a good answer for what to do with `Method::Unknown`, but this seems like a reasonable-enough way to handle them.

Considering the confusion that seems to exist around the exact behavior of `accepts()` (and I made several mistakes in the process of putting the above enumeration together!), perhaps we should consider rewriting this method to use an exhaustive `match`, even if it means we can't use `is_response()` and similar utilities.

algesten:
> Considering the confusion that seems to exist around the exact behavior of accepts() (and I made several mistakes in the process of putting the above enumeration together!), perhaps we should consider rewriting this method to use an exhaustive match, even if it means we can't use is_response() and similar utilities.

Agree. I also think this function should default to `false` at the end and explicitly enumerate the cases where we _do_ accept the message. So it's the opposite of today really.

OxleyS:
Okay, I'll take a stab at doing that tomorrow.

src/ice/agent.rs
```diff
@@ -761,63 +761,82 @@ impl IceAgent {
     /// Tells whether the message is for this agent instance.
     ///
     /// This is used to multiplex multiple ice agents on a server sharing the same UDP socket.
-    /// For this to work, the server should operate in ice-lite mode and not initiate any
-    /// binding requests itself.
```
OxleyS, line 765: This comment seemed incorrect to me, to my knowledge there's nothing about locally-originated binding requests that stop this method from working properly.

src/ice/agent.rs
```diff
@@ -761,63 +761,82 @@ impl IceAgent {
...
             }
+            integrity_passed
+        };
 
-            if let Some(remote_creds) = &self.remote_credentials {
-                if remote != remote_creds.ufrag {
+        let method = message.method();
+        let class = message.class();
+        match (method, class) {
+            (StunMethod::Binding, StunClass::Request | StunClass::Indication) => {
```
OxleyS, line 786: RFC 5389 indicates that requests and indications should be [validated in a similar manner](https://www.rfc-editor.org/rfc/rfc5389#section-10.1.2).

src/ice/agent.rs
```diff
@@ -761,63 +761,82 @@ impl IceAgent {
...
+            (StunMethod::Binding, StunClass::Unknown) => {
+                // Without a known class, it's impossible to know how to validate the message
+                trace!("Message rejected, unknown STUN class");
+                false
+            }
+            (StunMethod::Unknown, _) => {
+                // Without a known method, it's impossible to know how to validate the message
+                trace!("Message rejected, unknown STUN method");
+                false
+            }
```
OxleyS, line 838: These are behavior changes from my previous enumeration. Since processing is so method and class-dependent, not knowing either means we can't really know how to even verify that this message is valid, or for this instance.

src/io/mod.rs
```diff
@@ -12,7 +12,10 @@ use thiserror::Error;
 mod stun;
 pub(crate) use stun::stun_resend_delay;
 pub use stun::StunMessage;
-pub(crate) use stun::{StunError, TransId, STUN_MAX_RETRANS, STUN_MAX_RTO_MILLIS, STUN_TIMEOUT};
+pub(crate) use stun::{
+    Class as StunClass, Method as StunMethod, StunError, TransId, STUN_MAX_RETRANS,
```
OxleyS, line 16: I thought their original names would cause too much confusion when used in outside contexts like `IceAgent`, so I renamed them at re-export time. Would it perhaps be better to just rename the original enums?

src/io/stun.rs
```diff
@@ -143,16 +143,19 @@ impl<'a> StunMessage<'a> {
...
+
     /// Whether this STUN message is a BINDING request.
     pub(crate) fn is_binding_request(&self) -> bool {
         self.method == Method::Binding && self.class == Class::Request
     }
 
-    /// Whether this STUN message is a response.
-    pub(crate) fn is_response(&self) -> bool {
-        matches!(self.class, Class::Success | Class::Failure)
-    }
```
OxleyS, line 154: This is no longer used anywhere.

src/io/stun.rs
```diff
@@ -316,7 +319,7 @@ impl<'a> StunMessage<'a> {
 const MAGIC: &[u8] = &[0x21, 0x12, 0xA4, 0x42];
 
 #[derive(Clone, Copy, Debug, PartialEq)]
-enum Class {
+pub(crate) enum Class {
```
OxleyS, line 322: I believe that with the way things are set up, these enums are now visible to other modules in str0m, but not to consumers of the crate, so public API remains unchanged. Let me know if there's a better way to do this, though.

src/ice/agent.rs
algesten: Yeah. I don't know why I wrote that. Brainwaves…

src/ice/agent.rs
```diff
@@ -761,63 +761,82 @@ impl IceAgent {
...
+                }
 
-            if !belongs_to_a_candidate_pair {
-                trace!("Message rejected, transaction ID does not belong to any of our candidate pairs");
-                return false;
+                do_integrity_check(true)
             }
-        }
+            (StunMethod::Binding, StunClass::Success | StunClass::Failure) => {
+                let belongs_to_a_candidate_pair = self
```
algesten, line 817: Maybe we should check nominated pair here first, to make a less compute intensive happy path?

algesten (changes requested):
Looks great!

Thank you for doing this!

Small question

src/io/mod.rs
algesten: Looks good

src/io/stun.rs
algesten: pub(crate) is good

src/ice/agent.rs
OxleyS: Ah, yes, you're right! We even talked about that in the other PR, I guess it got lost amongst the other discussion happening there. I'll do that.

OxleyS: I've looked deeper into the code, and while adding this fast path is possible, it's surprisingly risky and would be better left to a separate PR, if we deem it necessary at all.

The issue is that our nominated pair `IceAgent.nominated_send` is not an index into the pairs array, but rather a `PairId`. Resolving that to an index would require a linear scan of the pairs array anyway, negating any benefits of checking it first. Storing an index instead would be very difficult because we sort pairs by priority and prune failed pairs often, making the bookkeeping very error-prone.

lolgesten: Alright! Thanks for checking!

lolgesten (approved):

algesten (approved):

Issue #486, opened 2024-03-21
Adding new candidates and invalidating all previous ones flaps instantly to disconnected


thomaseizinger:
When roaming networks, we learn about new interfaces and retire old ones. It appears that, if str0m is told about all of this information at once (i.e. `invalidate` all current candidates and add new ones), it instantly flaps to `Disconnected` instead of trying the new candidate pairs.

I think this is due to `evaluate_state` being called at the very top of `handle_timeout`: https://github.com/algesten/str0m/blob/aeda389f22d534729249dbae90cdabb2858252ce/src/ice/agent.rs#L851

Is this on purpose? I would have expected str0m first start testing all newly added candidates before concluding that we are `Disconnected`.

thomaseizinger:
I see, we currently treat it as drastic so that is probably the bug here.

Let me test with delaying that for a bit.

thomaseizinger:
Some preliminary testing suggests that this works quite well! Invalidating the old candidate and delaying acting on `Disconnected` for a few seconds means we can roam quite quickly! :)

algesten:
Wihoo!

thomaseizinger:
One question though, why doesn't it immediately switch to `Checking` in this case? The "not immediately acting on `Disconnected`" works but harms other cases where e.g. the remote peer goes away and we need to fallback to a new one.

algesten:
I think it's because we nominated, and that nomination goes away. If I recall correctly Checking is maybe only seen before anything is connected.

thomaseizinger:
Looking closer at the state machine, there seems to never be a transition to `Checking` unless we were in `New` or did `ice_restart`. Yet, in our case, we end up in `Completed` (we invalidate all candidates apart from the nominated one).

Once we are in `Completed`, the only transitions are `Connected` or `Disconnected`. However, it appears that it would be a valid transition to go from `Completed` to `Checking` if we have non-nominated candidates that were just added.

algesten:
I'm wondering if this is a spec thing. I don't think I ended up here by accident.

thomaseizinger:
I've grepped the spec for `Running` and couldn't find anything in that regards.

But outside of that, this looks like a bug to me. From str0m's _behaviour_ we are in `Checking` mode. All the added candidates are being checked. It is just the reported state that says `Disconnected`.

algesten:
[...]

PR #489, opened 2024-03-21
fix(ice): do not disconnect whilst we still check new candidates

thomaseizinger:
In case the active candidate pair gets invalidated, str0m currently instantly emits a `Disconnected` event. For example, if a candidate pair with a peer-reflexive candidate gets nominated and later replaced. This results in a disconnected event despite us not actually being disconnected (a peer-reflexive candidate that gets replaced is after all the exact same socket).

To avoid this false-positive disconnection, we transition back to `Checking` in case we are still evaluating other candidates.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1673,6 +1673,8 @@ impl IceAgent {
                     if !any_still_possible {
                         self.set_connection_state(Completed, "no more possible to try");
                     }
+                } else if any_still_possible {
+                    self.set_connection_state(Checking, "got new possible");
                 } else {
                     self.set_connection_state(Disconnected, "none nominated");
                 }
@@ -1682,6 +1684,8 @@ impl IceAgent {
                     if any_still_possible && !self.ice_lite {
                         self.set_connection_state(Connected, "got new possible");
                     }
+                } else if any_still_possible {
+                    self.set_connection_state(Checking, "got new possible");
                 } else {
                     self.set_connection_state(Disconnected, "none nominated");
                 }
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -693,6 +693,55 @@ mod test {
         }));
     }
 
+    #[test]
+    pub fn no_disconnect_when_replacing_pflx_with_real_candidate() {
+        let mut a1 = TestAgent::new(info_span!("L"));
+        let mut a2 = TestAgent::new(info_span!("R"));
+
+        let c1 = host("1.1.1.1:1000", "udp");
+        let c2 = host("1.1.1.1:1001", "udp");
+
+        // We need a 2nd pair of candidates to make sure the agent doesn't go straight into `Completed`.
+        let c3 = relay("2.2.2.2:1000", "udp");
+        let c4 = relay("2.2.2.2:1001", "udp");
+
+        // Agent 1 knows all candidates ahead of time.
+        a1.add_local_candidate(c1.clone());
+        a1.add_remote_candidate(c2.clone());
+        a1.add_local_candidate(c3.clone());
+        a1.add_remote_candidate(c4.clone());
+
+        // Agent 2 only knows his own candidates (imagine signalling layer being a bit slow)
+        a2.add_local_candidate(c2.clone());
+        a2.add_local_candidate(c4.clone());
+
+        a1.set_controlling(true);
+        a2.set_controlling(false);
+
+        // Wait until agent 2 is connected (based on a peer-reflexive candidate)
+        loop {
+            if a2.state().is_connected() {
+                break;
+            }
+            progress(&mut a1, &mut a2);
+        }
+
+        // Candidates arrive via signalling layer
+        a2.add_remote_candidate(c1.clone());
+        a2.add_remote_candidate(c3.clone());
+
+        // Continue normal operation.
+        for _ in 0..50 {
+            progress(&mut a1, &mut a2);
+        }
+
+        // We expect to not disconnect as part of this.
+        assert!(!a2.has_event(|e| matches!(
+            e,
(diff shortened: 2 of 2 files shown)


thomaseizinger:
@algesten What is your appetite for merging this? I am hitting this now again in the context of a peer-reflexive candiate getting replaced with a regular one.

In case the peer-reflexive candidate was already nominated, we will flip to disconnected despite us checking the new candidate.

algesten:
> Whilst doable, it is a bit annoying to implement and in the case of a real disconnection, harms the user experience because we can't give feedback until the additional timer expires.

Does your solution deal with road warriors on mobile data? If so, you could easily be in a situation where you reach `Disconnected` without being disconnected (and it coming back). We see the exact same thing when implementing browser WebRTC clients. Knowing that something is _definitely disconnected_ is not something ICE can help you with. It's better to go "Reconnecting…" after some timeout to avoid flapping.

(from #486)
> Currently, we treat `Disconnected` as a hard error and clean up our connection state. That is simple to understand which is why I'd like to keep it.

My worry here is that your need comes from a design decision where you want to treat `Disconnected` as a hard error when that's not what it is even in the ICE spec. I don't think we should land functionality to try and fix something that goes against how ICE works.

My preference is to stick close to [WebRTC](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/iceConnectionState#disconnected), but if it's super important to you, I don't think this PR would break anything, so we could potentially land it.

xnorpx:
For example in our codebase, we are letting our connection stick around for 30 or 45 seconds when we get a disconnect. In case clients shows up again.

thomaseizinger:
> > Whilst doable, it is a bit annoying to implement and in the case of a real disconnection, harms the user experience because we can't give feedback until the additional timer expires.
> 
> Does your solution deal with road warriors on mobile data? If so, you could easily be in a situation where you reach `Disconnected` without being disconnected (and it coming back). We see the exact same thing when implementing browser WebRTC clients. Knowing that something is _definitely disconnected_ is not something ICE can help you with. It's better to go "Reconnecting…" after some timeout to avoid flapping.

100%. Reliable roaming is something that we **really** care about. Especially because for relayed connections, deploying new relays is basically like roaming so we need this to work reliably to not interrupt our users every time we deploy. The way it works is that the client applications listen for system-specific events whenever network interfaces change. When that happens, we reset all connections and establish new ones, i.e. we discard all `IceAgent`s and rebind our sockets to new ports (and potentially new interfaces).

Because our application is an IP-tunnel, this means the layered connection is essentially "reconnecting" (i.e. TCP packets will start to get retransmitted by the kernel etc). We use those packets as "intents" that we need to set up a new ICE connection.

> > Currently, we treat `Disconnected` as a hard error and clean up our connection state. That is simple to understand which is why I'd like to keep it.
> 
> My worry here is that your need comes from a design decision where you want to treat `Disconnected` as a hard error when that's not what it is even in the ICE spec. I don't think we should land functionality to try and fix something that goes against how ICE works.

I don't think it is against the ICE spec per se. Admittedly, I've not dealt with it as much as you have but in my experience, ICE isn't very good at repairing connection state during roaming or when network conditions _actually_ change, at least not without being told about it. An ICE restart is basically what we are doing except that we just make an entire new agent to begin with. What happens in most cases is that the network link is dead and you can only detect it because X STUN messages have timed out.

For those cases where it is really just intermittent packet loss, the bit that we actually care about is for how long we tolerate that. This is where we get to this:

> For example in our codebase, we are letting our connection stick around for 30 seconds when we get an ICE disconnect. In case clients shows up again.

The bit that I don't understand with this approach is: What is the difference between the configured STUN timeout + adding 30 seconds on top vs re-configuring the `IceAgent` timeouts to give you the duration you want?

str0m's default are 9 STUN messages with a 3s interval and after 50% missing responses we probe more quickly (or something like that). Effectively, it results in a timeout of ~12s I think. For us, that is a bit too long, so we tweak those values in our application. If you can tolerate a longer partition, then you can tweak them upwards and allow up to 15 or 20 missed STUN requests.

Eventually though, there needs to be a timeout. It seems redundant to me to re-implement this on top of str0m when str0m has it already built in.

---

The main thing that this PR is changing is that:

- if we have a nominated candidate pair
- and we invalidate that for whatever reason (because it e.g. was a based on a peer-reflexive candidate)
- and we still have candidate pairs whose status we don't know yet

Then we transition back to `Checking` instead of `Disconnected`.

thomaseizinger:
If this PR is not desired, would you accept a PR that exposes a more fine-granular state on the `IceAgent` that allows me to check whether or not we are still testing candidates? Then I could implement a check myself to wait for `Completed` before acting on the `disconnected` event.

algesten:
Let's land this PR. I think it's ok.

thomaseizinger:
> Let's land this PR. I think it's ok.

Awesome! Would you like me to add a test for it?

Also, I am just changing over our dependency to the fork to test drive this a bit more. We used to have this in our fork and the tests were pretty stable so I am pretty confident that the flakiness come with depending on str0m mainline a couple of days ago.

Better be sure though. Will ping once ready!

xnorpx:
Having an integration test that validates the disconnect would be nice together with example how to configure the timeouts.

thomaseizinger:
> Having an integration test that validates the disconnect would be nice together with example how to configure the timeouts.

I added a failing test. The setup is a bit peculiar but it because we need an additional set of candidates. It should demonstrate the problem nicely though.

Regarding the documentation, I think we may need an additional PR to expose the ice timeout config added in https://github.com/algesten/str0m/pull/537 on the `RtcConfig` before that can be used there. Unless you are using the `IceAgent` directly too, then you can just use the exposed config options.

algesten (approved):

algesten:
Looks good! Let's merge it!

PR #492, opened 2024-03-25
Opus DTX support

OxleyS:
This adds support for an Opus format param `usedtx`, which signals a receiver's preference for DTX. DTX stands for [Discontinuous Transmission](https://datatracker.ietf.org/doc/html/rfc7587#autoid-7), and it's a feature in Opus that lowers the packet rate, not just the bitrate, during periods of silence. For instance, in my WebRTC environment, Chrome and Firefox will reduce the packet rate from 50 packets per second to 4 during silence. This makes it ideal for SFU-type applications, where any particular person spends most of the time not talking, and so we can save the CPU time spent processing such packets.

As mentioned in the above link, it does have slight effects on the resulting audio, and browsers don't currently negotiate it without SDP munging, so here I've set it off by default.

Known caveat: at a reduced rate of 4 packets per second, if your `StreamRx.pause_threshold` is less than 250 ms, str0m will emit stream pause/unpause events between each packet. That said, this could be considered expected behavior.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -138,6 +138,12 @@ pub struct FormatParams {
     /// Specifies that the decoder can do Opus in-band FEC
     pub use_inband_fec: Option<bool>,
 
+    /// Opus specific parameter.
+    ///
+    /// Specifies that the decoder prefers DTX (Discontinuous Transmission) such that
+    /// the packet rate is greatly lowered during periods of silence.
+    pub use_dtx: Option<bool>,
+
     /// Whether h264 sending media encoded at a different level in the offerer-to-answerer
     /// direction than the level in the answerer-to-offerer direction, is allowed.
     pub level_asymmetry_allowed: Option<bool>,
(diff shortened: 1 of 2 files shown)


src/format.rs
```diff
@@ -313,6 +319,13 @@ impl PayloadParams {
             score = score.saturating_sub(2);
         }
 
+        // If neither value is specified both sides should assume DTX is not used as this is
+        // the default.
+        let either_dtx_specified = c0.format.use_dtx.is_some() || c1.format.use_dtx.is_some();
+        if either_dtx_specified && c0.format.use_dtx != c1.format.use_dtx {
+            score = score.saturating_sub(1);
+        }
```
OxleyS, line 327: I wasn't sure how much to lower the score for a DTX mismatch. I notice `minptime` lowers it by 1 and `useinbandfec` lowers it by 2. Do they need to be distinct penalties?

lolgesten: I think I might have aimed for making it deterministic what the sorting order is.

I.e. `minptime` and `use_dtx` shouldn't result in the same score.

lolgesten (commented):
Apart from the sort order penalty, I think this is ready to go.

src/format.rs
OxleyS: Hmm, if I had to put an ordering, then, I think DTX differences should be the biggest penalty, due to both the large difference in packet volume (and resulting bandwidth) between having it on vs. off, and the claimed small ding in audio quality.

If we want a deterministic sorting order, then, it sounds like we need to decrease the score by 4 in that case, since 3 would overlap with a simultaneous difference in `minptime` and `useinbandfec`. I'll make that change now.

algesten (approved):
Let's ship it!

algesten:
Thanks!

Issue #473, opened 2024-03-05
Remove panic on missing candidates


thomaseizinger:
The following panic can be triggered if the `IceAgent` is being fed "wrong" data: https://github.com/algesten/str0m/blob/eb0fbe0e01d04e38b8b230a9995951d38447750f/src/ice/agent.rs#L1166-L1168

I'd argue that we should remove this panic and instead just drop the message. This can for example be triggered if the `IceAgent` is not configured with a host candidate but only with srflx candidates. I don't really see why that wouldn't be valid?

Generally, I don't think we should ever panic if the user hands us bad socket addresses or data within `Receive` / `StunPacket`. I'd prefer if we log a `warn!` and drop the packet. We are after all exposed to network traffic here and should gracefully handle whatever is thrown at us.

algesten:
Yeah, agree. Panic is a bit much. Dropping the packet seem reasonable.

algesten:
This has been addressed right?

thomaseizinger:
> This has been addressed right?

Nope! Not as far as I know.

PR #493, opened 2024-04-03
fix: don't panic on STUN requests for unknown interfaces

thomaseizinger:
Resolves: #473.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -842,14 +842,14 @@ impl IceAgent {
     /// Handles an incoming STUN message.
     ///
     /// Will not be used if [`IceAgent::accepts_message`] returns false.
-    pub fn handle_packet(&mut self, now: Instant, packet: StunPacket) {
+    pub fn handle_packet(&mut self, now: Instant, packet: StunPacket) -> bool {
         trace!("Handle receive: {:?}", &packet.message);
 
         // Regardless of whether we have remote_creds at this point, we can
         // at least check the message integrity.
         if !self.accepts_message(&packet.message) {
             debug!("Message not accepted");
-            return;
+            return false;
         }
 
         if packet.message.is_binding_request() {
@@ -864,6 +864,8 @@ impl IceAgent {
         });
 
         // TODO handle unsuccessful responses.
+
+        true
     }
 
     /// Provide the current time to the [`IceAgent`].
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1893,6 +1893,28 @@ mod test {
         assert!(stun_message.is_successful_binding_response());
     }
 
+    #[test]
+    pub fn discards_packet_from_unknown_candidate() {
+        let mut agent = IceAgent::new();
+        let remote_creds = IceCreds::new();
+        agent.set_remote_credentials(remote_creds.clone());
+
+        let request =
+            make_serialized_binding_request(&agent.local_credentials, &remote_creds, false, 0);
+
+        agent.handle_packet(
+            Instant::now(),
+            StunPacket {
+                proto: Protocol::Udp,
+                source: ipv4_1(),
+                destination: ipv4_2(),
+                message: StunMessage::parse(&request).unwrap(),
+            },
+        );
+
+        assert!(agent.poll_transmit().is_none());
+    }
+
     fn make_serialized_binding_request(
         local_creds: &IceCreds,
         remote_creds: &IceCreds,
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1169,25 +1169,28 @@ impl IceAgent {
             self.remote_candidates.len() - 1
         };
 
-        let local_idx = self
-            .local_candidates
-            .iter()
-            .enumerate()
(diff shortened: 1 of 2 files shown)


src/ice/agent.rs
```diff
@@ -1169,25 +1169,28 @@ impl IceAgent {
...
-            .expect(
-                "STUN request for socket that is neither a host nor a relay candidate. This is a config error.",
-            )
-            .0;
+        let local_idx = match self.local_candidates.iter().enumerate().find(|(_, v)| {
+            // The local candidate will be
+            // either a host candidate (for cases where the request was not received
+            // through a relay) or a relayed candidate (for cases where it is
+            // received through a relay).  The local candidate can never be a
+            // server-reflexive candidate.
```
thomaseizinger, line 1177: I am wondering if this is still true with our changes in https://github.com/algesten/str0m/pull/455? Shouldn't we instead also check for server-reflexive candidates and compare the base instead?

lolgesten (commented):
Why does this scenario not use `.accepts_message()` to test whether the ICE agent want the packet?

thomaseizinger:
> While I accept we don't have to panic, I still believe the user should configure the ICE agent correctly. Silently discarding random UDP packets means it's easy to introduce bugs and not notice a misconfiguration.

We do use `accepts_message` in our code and I can still trigger this panic. That being said, we don't yet have https://github.com/algesten/str0m/pull/488in our fork so that might change things?

thomaseizinger:
`accepts_message` only checks the message and doesn't have access to the `destination` `SocketAddr` so it cannot check whether it has a candidate for that:

https://github.com/algesten/str0m/blob/3c5edeea9bb12f62d3e5b185bae6be7165700228/src/ice/agent.rs#L767

algesten:
> We do use `accepts_message` in our code and I can still trigger this panic. That being said, we don't yet have https://github.com/algesten/str0m/pull/488in our fork so that might change things?

Interesting! That would be good to know. It makes a big difference to me if accepts_message says "yes" and handle_message says "panic"!

Spontaneously it seems these should be in lockstep.

thomaseizinger:
> > We do use `accepts_message` in our code and I can still trigger this panic. That being said, we don't yet have [algesten/str0m/pull/488in](https://github.com/algesten/str0m/pull/488in) our fork so that might change things?
> 
> Interesting! That would be good to know. It makes a big difference to me if accepts_message says "yes" and handle_message says "panic"!
> 
> Spontaneously it seems these should be in lockstep.

See the comment above. i think we'd have to change `accepts_message` to accept a `StunPacket` to make it say "no" to packets from unknown interfaces.

Note that the way this panic gets triggered for us is:

1. We invalidate all other candidates as soon as we have a working path
2. We want to delay sending of relay candidates by ~1s to first check if we can hole-punch

Combinding (1) and (2) leads to a race condition where we might have already invalidated a candidate but the remote wants to still send use STUN requests.

thomaseizinger:
My take on it is:

- `IceAgent::handle_packet` must not panic, regardless what data you feed it. IMO, one of the aspects of a SANS-IO design is that you can (directly) feed it unstructured input from the network. Thus, it must not panic or otherwise we have a DoS attack vector. I think the correct solution for this fn is to discard the message.
- `IceAgent::accepts_message` is useful when multiple agents are multiplexed on a single socket. IMO there isn't a lot of value in having this function (we've discussed this before but I'll mention it for completeness). Processing a message and checking whether we want to do so requires very similar code. Code that must not diverge. To me, it seems like we could just do whatever we do in `accepts_message` at the very top of `handle_message` and return `true | false` from it, indicating whether we've handled the message. It might not hold true for the entire `Rtc` instance but at least for the `IceAgent`, doing the actual processing should be very cheap.

Perhaps as a compromise, we could make `accepts_message` `pub(crate)` such that the `Rtc` instance can still call it but the public API of `IceAgent` doesn't expose it? Then it wouldn't be a silent discard but just returning `false` to indicate that we didn't handle it.

thomaseizinger:
We can also (additionally) align the two APIs such that the entire packet can be packet can be passed to `accepts_message` and checked early.

algesten (approved):
I'm good merging this now, if you think it's ok?

thomaseizinger (commented):
If we return `false` now, should we also change `stun_server_handle_message` to return `false` in case we discard it based on an unknown candidate?

algesten:
> If we return `false` now, should we also change `stun_server_handle_message` to return `false` in case we discard it based on an unknown candidate?

I tried that but didn't like where it was going. Places like `if candidate.discarded { return` are not clear whether it's true or false.

I decided that rather than muddle the semantics, the bool is exactly that of `accepts_message`.

thomaseizinger:
> I decided that rather than muddle the semantics, the bool is exactly that of `accepts_message`.

Okay, in that case, isn't that commit unrelated to this change? I am not sure how atomic you want to be on your PRs but it could be done separately in that case.

Otherwise, I am happy with this going in :)

lolgesten:
Let's merge it! I'll reference the PR twice in the changelog.

PR #497, opened 2024-04-17
feat: expose `Candidate::discarded`

thomaseizinger:
This is useful / necessary to differentiate between active and invalidated candidates when accessing them via `IceAgent::local_candidates`.

Alternatively, we could also filter them directly there or prune all discarded candidates from the agent?

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -405,7 +405,10 @@ impl Candidate {
         self.discarded = true;
     }
 
-    pub(crate) fn discarded(&self) -> bool {
+    /// Whether the candidate has been discarded.
+    ///
+    /// For example, it might be redundant or invalidated via [`IceAgent::invalidate_candidate`].
+    pub fn discarded(&self) -> bool {
         self.discarded
     }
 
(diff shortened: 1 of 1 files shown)


algesten (changes requested):
Could we filter them instead?

As I see it, `Candidate` should not have a state. The discarded flag is probably there or convenience, but we don't want it. Especially not in public API.

thomaseizinger:
> Could we filter them instead?
> 
> As I see it, `Candidate` should not have a state. The discarded flag is probably there or convenience, but we don't want it. Especially not in public API.

Yeah definitely. So `local_candidates` and `remote_candidates` always filtered by `!discarded`?

Shall we move them to a different collection perhaps so we can remove the state?

algesten:
I think because we use usize indexes to the vec holding them, we can't move them out easily. We could make some wrapper Entry type to hold the state and candidate though.

thomaseizinger:
> I think because we use usize indexes to the vec holding them, we can't move them out easily. We could make some wrapper Entry type to hold the state and candidate though.

Hmm, right. Perhaps a [`slab`](https://docs.rs/slab/latest/slab/) would help here?

PR #498, opened 2024-04-18
feat: allow invalidating remote candidates

thomaseizinger:
The new API discussed in https://github.com/algesten/str0m/pull/496. Opening here for further discussion as a draft. Will report back with results of experiment but the passing unit test is already giving me good vibes :)

### DIFF
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -75,6 +75,7 @@ mod test {
 
     use crate::io::{Protocol, StunMessage, StunPacket, STUN_TIMEOUT};
     use tracing::Span;
+    use tracing_subscriber::util::SubscriberInitExt;
 
     pub fn sock(s: impl Into<String>) -> SocketAddr {
         let s: String = s.into();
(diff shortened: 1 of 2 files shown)


src/ice/mod.rs
```diff
@@ -302,6 +303,64 @@ mod test {
...
+
+            if a1_nominated && a2_nominated {
+                break;
+            }
+
+            progress(&mut a1, &mut a2);
+        }
+
+        assert!(a1.time.duration_since(a1_time) < STUN_TIMEOUT);
+        assert!(a2.time.duration_since(a2_time) < STUN_TIMEOUT);
```
thomaseizinger, line 361: Without the consideration of remote-candidates in the `invalidate_candidate` function, this assertion fails.

algesten (approved):
This looks good!

thomaseizinger:
Prelimiary testing suggests that this works quite well! Watch this space.

thomaseizinger:
This has been working really well for us and it is not even an API addition itself, just expanding the scope of `invalidate_candidate`.

There is an issue with candidate invalidation in general though, see https://github.com/algesten/str0m/pull/504.

algesten:
Thanks!

PR #501, opened 2024-04-22
Fix lazy NACK timeouts enabling

evdokimovs:
In commit "Lazily enable nack timeouts" ([f72b690][1]) was introduced optimization of NACK timeouts enabling, this commit added `Streams::any_nack_enabled` resetting in `Streams::stream_rx_by_mid_rid()`, but `Streams` has exact the same method called `Streams::rx_by_mid_rid` (which is used under the hood in `Direct::stream_rx_by_mid`) which wasn't updated. So if you're getting `StreamRx` from Direct API and call `suppress_nack(false)` on it, then `any_nack_enabled` is not updated, resulting in NACKs not being transmitted.

This PR removes this duplication of methods `rx_by_mid_rid` and `tx_by_mid_rid` in favor of `stream_tx_by_mid_rid` and `stream_rx_by_mid_rid`.

[1]: https://github.com/algesten/str0m/commit/f72b690dc7961dc40d1c30f1bbbc692d83b13125

### DIFF
--- a/src/change/direct.rs
+++ b/src/change/direct.rs
@@ -211,7 +211,7 @@ impl<'a> DirectApi<'a> {
 
     /// Obtain a recv stream by looking it up via mid/rid.
     pub fn stream_rx_by_mid(&mut self, mid: Mid, rid: Option<Rid>) -> Option<&mut StreamRx> {
-        self.rtc.session.streams.rx_by_mid_rid(mid, rid)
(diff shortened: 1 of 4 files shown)


algesten:
Oh wow. This is a big 🤦  :D

algesten:
Thanks!

PR #503, opened 2024-04-24
chore: set logger without mucking around with env vars

thomaseizinger:
Env vars are global to a process. Setting them from within the tests is problematic because those tests are run concurrently in the same process.

`tracing` allows us to set a logger for the current stack scope using `set_default`. Instead of fighting over which test should initialize the logger, we can only set it for the current scope of each test.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1571,7 +1571,7 @@ impl IceAgent {
 mod test {
     use super::*;
     use std::net::SocketAddr;
-    use std::sync::Once;
+    use tracing_subscriber::util::SubscriberInitExt as _;
 
     impl IceAgent {
         fn pair_indexes(&self) -> Vec<(usize, usize)> {
@@ -1746,21 +1746,10 @@ mod test {
 
     #[test]
     fn form_pairs_replace_remote_redundant() {
-        use std::env;
-        use tracing_subscriber::{fmt, prelude::*, EnvFilter};
-
-        if env::var("RUST_LOG").is_err() {
-            env::set_var("RUST_LOG", "debug");
-        }
-
-        static START: Once = Once::new();
-
-        START.call_once(|| {
-            tracing_subscriber::registry()
-                .with(fmt::layer())
-                .with(EnvFilter::from_default_env())
-                .init();
-        });
+        let _guard = tracing_subscriber::fmt()
+            .with_env_filter("debug")
+            .with_test_writer()
+            .set_default();
 
         let mut agent = IceAgent::new();
         agent.set_ice_lite(true);
--- a/examples/chat.rs
+++ b/examples/chat.rs
@@ -19,25 +19,25 @@ use str0m::media::{Direction, KeyframeRequest, MediaData, Mid, Rid};
 use str0m::media::{KeyframeRequestKind, MediaKind};
 use str0m::net::Protocol;
 use str0m::{net::Receive, Candidate, Event, IceConnectionState, Input, Output, Rtc, RtcError};
+use tracing::subscriber::DefaultGuard;
+use tracing_subscriber::util::SubscriberInitExt as _;
+use tracing_subscriber::EnvFilter;
 
 mod util;
 
-fn init_log() {
-    use std::env;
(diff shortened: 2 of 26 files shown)


src/ice/agent.rs
```diff
@@ -1746,21 +1746,10 @@ mod test {
...
-
-        START.call_once(|| {
-            tracing_subscriber::registry()
-                .with(fmt::layer())
-                .with(EnvFilter::from_default_env())
-                .init();
-        });
+        let _guard = tracing_subscriber::fmt()
+            .with_env_filter("debug")
+            .with_test_writer()
```
thomaseizinger, line 1751: See https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.SubscriberBuilder.html#method.with_test_writer for why this is useful in tests.

lolgesten:
What problem are you trying to fix?

thomaseizinger:
> What problem are you trying to fix?

I hit a panic when trying to use `set_default` in one of the tests because another test was already setting a global logger.

I figured that setting the env var and using a `OnceCell` is kind of a hack so I opened this PR as a way to only set a logger for the current test.

algesten:
This is tailored to my workflow. My requirements are:

1. cargo test shows me log for the integration tests. It's enough for me to run `cargo test flappy_ice_lite_state` to get my reasonable defaults.
2. I can override using `RUST_LOG` like so: `RUST_LOG=blaha=trace cargo test flappy_ice_lite_state`

With 1, this is what I expect.

```text
martin@nugget:~/dev/str0m$ cargo test flappy_ice_lite_state
    Finished test [unoptimized + debuginfo] target(s) in 0.06s
…
running 1 test
2024-04-24T04:41:07.790877Z  INFO str0m::ice_::agent: Add local candidate: Candidate(host=1.1.1.1:1000/udp prio=2130706175)
2024-04-24T04:41:07.790927Z  INFO str0m::ice_::agent: Add local candidate: Candidate(host=2.2.2.2:2000/udp prio=2130706175)
2024-04-24T04:41:07.790977Z DEBUG str0m::channel: Allocate channel id: ChannelId(0)
2024-04-24T04:41:07.791111Z DEBUG str0m::change::sdp: Create offer
2024-04-24T04:41:07.791117Z DEBUG str0m::change::sdp: Accept offer
```

This PR seem to not have any reasonable defaults.

thomaseizinger:
> cargo test shows me log for the integration tests. It's enough for me to run `cargo test flappy_ice_lite_state` to get my reasonable defaults.

cargo's default behaviour is to not show logs if the test is passing. To override that, you need to pass `--nocapture`. Taken that into account, running `cargo test flappy_ice_lite_state -- --nocapture` does indeed display the logs.

I did notice that RA didn't show me all usages of `init_log` and the guard was dropped too early (including for the test you mentioned) which is now fixed.

If you don't want the `--nocapture` behaviour then we can remove the `with_test_writer` from the logger setup. However, having passing tests be quiet by default is IMO kinda nice.

algesten:
> cargo's default behaviour is to not show logs if the test is passing. To override that, you need to pass `--nocapture`. Taken that into account, running `cargo test flappy_ice_lite_state -- --nocapture` does indeed display the logs.

Fully aware of all that. And no, I don't want to write `--nocapture` on every test run.

thomaseizinger:
> > cargo's default behaviour is to not show logs if the test is passing. To override that, you need to pass `--nocapture`. Taken that into account, running `cargo test flappy_ice_lite_state -- --nocapture` does indeed display the logs.
> 
> Fully aware of all that. And no, I don't want to write `--nocapture` on every test run.

We can remove `with_test_writer` then!

I would like to understand the reasoning though. In what scenario do you need to see the output of a passing test?

algesten:
> I would like to understand the reasoning though. In what scenario do you need to see the output of a passing test?

Many times when I use the integration tests, it's to follow a flow of events happening. My use case is not necessarily a pass/fail check, it's figuring out what's going on using an integration test as reproduction.

This is different to unit tests, where there pretty much always is a pass/fail and I have no need to see logs for that.

PR #504, opened 2024-04-24
fix: only invalidate candidates of the same kind

thomaseizinger:
We ran into a bug where a candidate was incorrectly invalidated based on IP only without considering the kind of candidate. The bug was in our software because we didn't invalidate a host candidate and our tests didn't catch it because there, we invalidated it "accidentially" by wanting to invalidate the corresponding srflx candidate. The (unit)-tests don't have a NAT so there host == srflx. But in the real world, we had a NAT but the two nodes were on the same subnet, thus they ended up making a host-host connection and failed to migrate that to another candidate because it was never invalidated.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -700,11 +700,12 @@ impl IceAgent {
     /// Returns `true` if the candidate was found and invalidated.
     #[allow(unused)]
     pub fn invalidate_candidate(&mut self, c: &Candidate) -> bool {
-        if let Some((idx, other)) =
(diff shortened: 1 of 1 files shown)


algesten:
It's our own API, if we want kind, I don't see why not.

thomaseizinger:
@algesten Rebased on latest main now that #498 is merged.

lolgesten (approved):
Looks good!

algesten (approved):
(too early to juggle my two github users)

PR #510, opened 2024-04-30
Tune some logging levels

xnorpx:
To reduce the chattiness on debug move some packet level logging to trace.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1333,9 +1333,11 @@ impl IceAgent {
 
         let reply = StunMessage::reply(req.trans_id, req.source);
 
-        debug!(
+        trace!(
             "Send STUN reply: {} -> {} {:?}",
-            local_addr, remote_addr, reply
+            local_addr,
+            remote_addr,
+            reply
         );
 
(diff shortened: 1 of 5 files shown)


algesten:
@thomaseizinger @davibe @k0nserv in case you want to check whether these proposed log changes would negatively affect you.

k0nserv:
Fine by me!

src/ice/agent.rs
```diff
@@ -1333,9 +1333,11 @@ impl IceAgent {
 
         let reply = StunMessage::reply(req.trans_id, req.source);
 
-        debug!(
+        trace!(
```
thomaseizinger, line 1336: If this is changed to trace, perhaps the sending side should be changed too?

thomaseizinger (approved):
One comment but overall, this follows a convention that we've been following in our codebase: Anything that happens _per packet_ is on `trace`.

src/ice/agent.rs
xnorpx: good catch, updated

xnorpx:
@algesten can this be merged?

algesten (approved):

algesten:
Let's do this

Issue #517, opened 2024-05-21
Firefox rejects DTLS certificate when connected to multiple processes running str0m


OxleyS:
For context, we use str0m as an SFU. Due to our unique setup, a single browser client may be connected to multiple SFUs at the same time. Our SFUs generate one DTLS certificate each on startup and then re-use it for all of that SFU's rooms.

Firefox clients fail to connect to the second SFU in this setup, with an error of:
```
Dtls(Io(Custom { kind: InvalidData, error: Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 167773202, library: "SSL routines", function: "dtls1_read_bytes", reason: "sslv3 alert bad certificate", file: "ssl/record/rec_layer_d1.c", line: 613, data: "SSL alert number 42" }]))) } }))
```

This seems like a resurfacing of https://github.com/versatica/mediasoup/issues/127, which we have a workaround for [here](https://github.com/algesten/str0m/blob/main/src/crypto/ossl/cert.rs#L44). The problem is that both SFUs generate their certificate with the same starting serial number of `1`, since they are different processes. This brings us to the same situation again - same serial + issuer, but different certificate.

The problem seemed to be fixed when I replaced this atomic increment with a randomly-generated 128-bit number. Would that be an okay solution?

algesten:
Sure, I don't care as long as it works with all browsers.

PR #518, opened 2024-05-22
Change DTLS cert serial selection from sequential to random

OxleyS:
Resolves https://github.com/algesten/str0m/issues/517

This simply uses a randomly generated 128-bit value as the serial instead. I don't have a strong opinion on the exact length, but thought this would make collisions with other certs basically impossible.


algesten (approved):
Nice!

Issue #513, opened 2024-05-17
ICE Lite is not considered when determining ICE controlling/controlled


OxleyS:
We currently [set str0m unconditionally as the ICE controller](https://github.com/algesten/str0m/blob/main/src/change/sdp.rs#L565) when it is the initial offerer. However, according to [RFC 5245 Section 5.2](https://www.rfc-editor.org/rfc/rfc5245#section-5.2):

> One agent full, one lite:  The full agent MUST take the controlling role, and the lite agent MUST take the controlled role.

Assuming the other end is a full agent, str0m in lite mode must set itself as controlled, even as the initial offerer. If that's all there was to it, I'd just go straight to opening a PR 🙂  However, the RFC also says:

> Both lite:  The agent that generated the offer which started the ICE processing MUST take the controlling role, and the other MUST take the controlled role.

So in this instance, str0m would be doing the right thing, but it's not clear how we're supposed to tell whether the other side is ICE lite or not if we're the initial offerer (and thus have no remote `ice-lite` attribute to base it on). We would have to switch our role based on the answer, I think, but [that might be too late?](https://github.com/algesten/str0m/blob/main/src/ice/agent.rs#L377)

Not entirely clear what the best approach here may be (other than utilizing direct API, of course). While I decided against using ICE Lite in my project, since I found this I thought it might be good to file anyway, at least.

algesten:
Nice find!

That documentation about panic seems wrong. I can't see any reason this would panic. However changing the controlling after starting to form pairs of local-remote candidates would be wrong (since it affects candidate prio).

I think this is a simple fix, we can change the controlling mode given the response from the remote. So if we are `ice-lite` and remote answers `ice-lite`, we change controlling to true, since we were the offerer.

OxleyS:
Ah, okay, I wasn't sure if it was okay to change the controlling mode at the point of receiving the answer, but I can see how it is. I might take a stab at this next week.

PR #519, opened 2024-05-23
SDP negotiation now considers ICE Lite when determining ICE role

OxleyS:
Resolves https://github.com/algesten/str0m/issues/513. However, it does not make Lite-to-Lite connections fully functional - details below.

[RFC 5245](https://www.rfc-editor.org/rfc/rfc5245#section-5.2) says:
- In a Full-to-Full connection, the initial offerer takes the controlling role in ICE.
- In a Full-to-Lite connection, the Full peer takes the controlling role, regardless of which side offered.
- In a Lite-to-Lite connection, again the initial offerer takes the controlling role.

However, behavior prior to this PR would always consider the initial offerer to be controlling, irrespective of ICE Lite. This discrepancy caused role confusion and thus connection issues when the initial offerer was ICE Lite, regardless of that offerer being str0m itself or another client.

The RFC's logic can be implemented quite directly in our SDP negotiation code. The only bit of complication is in the Lite-to-Lite case, where we initially set ourselves as controlled when offering, but have to switch roles once we learn the remote peer is also ICE Lite via its SDP answer.

With this, Full-to-Full and Full-to-Lite should be fully functional in both offering directions. Lite-to-Lite will at least have the correct role set now, but there are other deficiencies in the ICE implementation that keep this case non-functional, namely the clash between the code both requiring connectivity checks and being unable to perform them. I haven't had time to dig deep into these issues yet, so I can't point out specific code. Since this case is exceedingly rare, though, resolving issues for the much more common Full-to-Lite case is more immediately valuable.

### DIFF
--- a/docs/ice.md
+++ b/docs/ice.md
@@ -76,6 +76,13 @@ Global Ta pacing is out of scope this implementation.
 
 We do not switch ice-lite during a session.
 
+## One or both sides must be ICE-Full
+
+Handling connections where both sides are ICE-Lite requires specialized logic,
+such as assuming connectivity without checks, after-the-fact candidate pair
+reconciliation, and switching roles after initial determination. We consider
+both sides being ICE-Lite exceedingly rare and thus do not support this case.
+
 ## ICE role conflicts won't happen
 
 https://datatracker.ietf.org/doc/html/rfc8445#section-6.1.1
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -85,8 +85,11 @@ impl<'a> SdpApi<'a> {
         }
 
         if !self.rtc.dtls.is_inited() {
-            // The side that makes the first offer is the controlling side.
-            self.rtc.ice.set_controlling(false);
+            // The side that makes the first offer is the controlling side, unless they
+            // are ICE Lite and we are ICE Full, in which case the roles are reversed.
+            self.rtc
+                .ice
+                .set_controlling(offer.session.ice_lite() && !self.rtc.ice.ice_lite());
         }
 
         // Ensure setup=active/passive is corresponding remote and init dtls.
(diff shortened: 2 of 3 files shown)


algesten:
Thanks for looking into this! 

> resolving issues for the much more common Full-to-Lite case is more immediately valuable.

For my understanding: the issues with Full-to-Lite in the current code is that str0m would assume controlling role also as Lite, if it is the first offerer?

algesten:
> Lite-to-Lite will at least have the correct role set now, but there are other deficiencies in the ICE implementation that keep this case non-functional, namely the clash between the code both requiring connectivity checks and being unable to perform them.

This is happening because str0m in ice-lite mode will not make connectivity checks. In the Lite-to-Lite scenario, it sounds like the offerer must not only become controlling, but also take the Full role to start connectivity checks.

OxleyS:
> For my understanding: the issues with Full-to-Lite in the current code is that str0m would assume controlling role also as Lite, if it is the first offerer?

Yes, in this scenario you would end up with both sides claiming to be the controlling role. But also going the other way, if a Lite client makes an initial offer to Full str0m, with the current code they would both claim to be the controlled role.

> That sounds like a lot of work for a fringe edge case. I suggest we state str0m does not support Lite-to-Lite and throw an Error to reject the OFFER/ANSWER when we detect this situation.

I definitely agree. Even if we did implement support for it, I can see it being very easy to accidentally break without anybody noticing. As long as we explicitly check for and reject this case as you suggest, it seems reasonable to not support it. Shall I modify the PR to do so?

algesten:
> I definitely agree. Even if we did implement support for it, I can see it being very easy to accidentally break without anybody noticing. As long as we explicitly check for and reject this case as you suggest, it seems reasonable to not support it. Shall I modify the PR to do so?

Yes please!

OxleyS:
Oops, I forgot to update the doc. I'll do that now.

algesten (approved):
Looks great!

PR #520, opened 2024-05-25
Provide reason for Output::Timeout

algesten:
This is by suggestion from @xnorpx.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -847,6 +847,7 @@ pub struct Rtc {
     peer_bytes_rx: u64,
     peer_bytes_tx: u64,
     change_counter: usize,
+    last_timeout_reason: Reason,
 }
 
 struct SendAddr {
@@ -967,7 +968,6 @@ pub enum Input<'a> {
 }
 
 /// Output produced by [`Rtc::poll_output()`]
-
 #[allow(clippy::large_enum_variant)]
 pub enum Output {
     /// When the [`Rtc`] instance expects an [`Input::Timeout`].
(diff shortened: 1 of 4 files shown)


davibe (approved):

algesten:
My one concern is that it's a breaking change for the API, and maybe details that a new user of str0m might not need.

I'm weighing this API vs adding a `Rtc::last_timeout_reason()`, leaving the API as is, and giving us an API _if we want_ to know the reason.

davibe:
> Rtc::last_timeout_reason()

like

algesten:
Alright. Pushed the alternative API. Also fixed some timeouts that should have returned None.

src/lib.rs
```diff
@@ -1466,9 +1548,34 @@ impl Rtc {
...
     }
 
+    /// The reason for the last [`Output::Timeout`]
+    ///
+    /// This is updated when calling [`Rtc::poll_output()`] and the next output
+    /// is a timeout.
+    ///
+    /// ```
+    /// # use str0m::{Rtc, Input, Output, Reason};
+    /// // A vec holding the managed rtc instances. One instance per remote peer.
```
xnorpx, line 1563: A vec comment can probably be removed here

xnorpx:
Tested it seems to be working:

2024-05-25T15:51:07.036296Z ERROR something: Ice, Instant {
    tv_sec: 12329,
    tv_nsec: 182128520,
}
2024-05-25T15:51:09.489807Z ERROR something: Stats, Instant {
    tv_sec: 12331,
    tv_nsec: 636220215,
}

I like this change, it's easy for us to pull it if we want the information but doesn't brake the current API.

LGTM!

src/lib.rs
algesten: thanks!

xnorpx (approved):
LGTM

PR #522, opened 2024-06-05
Replace eagerly managed source_keys_rx with rx_lookup

algesten:
Keeping source_keys_rx in sync with the changes of the streams_rx is
brittle. This changes so rx_lookup is a cache of confirmed lookups.


src/streams/mod.rs
```diff
@@ -290,21 +303,16 @@ impl Streams {
...
 
     pub fn remove_stream_rx(&mut self, ssrc: Ssrc) -> bool {
         let stream = self.streams_rx.remove(&ssrc);
         let existed = stream.is_some();
-        if let Some(rtx) = stream.and_then(|s| s.rtx()) {
-            self.source_keys_rx.remove(&rtx);
-        }
 
-        self.source_keys_rx.remove(&ssrc);
+        self.rx_lookup.retain(|k, l| *k != ssrc && l.main != ssrc);
```
k0nserv, line 315: Should this `&&` not be `||` to also remove rtx streams?

src/streams/mod.rs
```diff
@@ -334,28 +342,30 @@ impl Streams {
...
         // A direct hit on SSRC is to prefer. The idea is that mid/rid are only sent
         // for the initial x seconds and then we start using SSRC only instead.
-        if let Some(r) = self.source_keys_rx.get(&ssrc).copied() {
-            return Some(r);
+        if let Some(r) = self.rx_lookup.get_mut(&ssrc) {
+            r.last_used = now;
+            return Some((r.mid, r.main));
         }
 
-        // The receiver/source SSRC might already exist.
```
k0nserv, line 344: Is this not still the case, at least for non-simulcast tracks?

src/streams/mod.rs
```diff
@@ -600,24 +615,23 @@ impl Streams {
...
+            // Unwrap is OK, see above.
+            let to_change = self.streams_rx.remove(&ssrc_from).unwrap();
 
-        // Add mapping for new main SSRC
-        self.associate_ssrc_mid(ssrc_to, mid, ssrc_to, None);
+            // Reinsert under new SSRC key.
+            self.streams_rx.insert(ssrc_to, to_change);
 
-        // Map the RTX to the new main SSRC
-        if let Some(rtx) = rtx {
```
k0nserv, line 619: We don't need to remap the RTX SSRC anymore?

k0nserv (approved):

src/streams/mod.rs
lolgesten: `retain` is inverse logic no?

k0nserv: Right... Not enough coffee ☕

src/streams/mod.rs
algesten: Hm. Yes. Why did I drop this doc?

src/streams/mod.rs
algesten: That's the first commit, which means that instead of actively managing a `source_keys_rx`, we are passively using `lookup_rx` as a cache. We can just remove the key from the cache and next time a packet arrives, we will put it back in the cache.

k0nserv: Hm, but is that what we want? If we have

```
1 -> (1, A)
2 –> (1, A) (rtx)
```

then change it to

```
3 -> (3, A)
2 -> (1, A) (rtx)
```

if we get a packet on `2` well recreate `1 -> (1, A)`, but don't we want that packet to repair `3`? I guess it doesn't matter since we are repairing the stream `A` after all

algesten: This would be a change SSRC 1 -> 3, which means the lookup entries for 1 would both be gone since the retain clause is `*k != ssrc_from && l.main != ssrc_from)` – this should clear out both these entires:

```
1 -> (1, A)
2 –> (1, A) (rtx)
```

k0nserv: Ah course

Issue #523, opened 2024-06-06
Panic during TWCC interim building


OxleyS:
A server running str0m recently panicked with `assertion failed: t >= -32_765 && t <= 32_767`. The stack trace was a little busted due to being Release mode, but from the message it must have been this line:
https://github.com/algesten/str0m/blob/08562d74683c66431dcd389b0248fc67bd4396fa/src/rtp/rtcp/twcc.rs#L533
Because it was likely the result of janky network data, I don't have a reliable repro.

Digging into it, I was a little unsure what the intended bounds on `t` were. The context of this line is as follows: https://github.com/algesten/str0m/blob/08562d74683c66431dcd389b0248fc67bd4396fa/src/rtp/rtcp/twcc.rs#L527-L539

Here, `diff_time` is the difference in timestamp between two received packets. The result needs to fit in an `i16`, so our absolute bounds would be `[-32768, 32767]`, but here the panicking assertion instead checks for -32765 on the lower end. In addition, the first `if` branch here checks for `< -8_192_000`, which when divided by 250 yields -32768. Is the intended lower bound -32768 or -32765?

algesten:
This does indeed look broken

lolgesten:
@OxleyS I say it's -32768. Want to make a PR?

OxleyS:
-32768 also seemed like the intended value to me. Sure, I'll make a PR for it.

PR #524, opened 2024-06-06
Fix big TWCC time deltas

OxleyS:
Resolves https://github.com/algesten/str0m/issues/523.

While working on this, I found another issue when you have deltas that are big enough to overflow when cast down to an `i32`. This wasn't panicking, but would nevertheless cause strange results. The second commit contains a fix for this as well.

### DIFF
--- a/src/rtp/rtcp/twcc.rs
+++ b/src/rtp/rtcp/twcc.rs
@@ -530,7 +530,7 @@ fn build_interims(
             break;
         } else if diff_time < 0 || diff_time > 63_750 {
             let t = diff_time / 250;
-            assert!(t >= -32_765 && t <= 32_767);
+            assert!(t >= -32_768 && t <= 32_767);
             (PacketStatus::ReceivedLargeOrNegativeDelta, t as i16)
         } else {
             let t = diff_time / 250;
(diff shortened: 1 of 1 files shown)


src/rtp/rtcp/twcc.rs
```diff
@@ -515,13 +515,19 @@ fn build_interims(
...
         } else {
             let dur = r.time - prev.1;
-            dur.as_micros() as i32
+            dur.as_micros().try_into()
+        };
+
+        let Ok(diff_time) = diff_time else {
+            // This time diff was so large we can't even reliably represent it. Treat the same
+            // as the more sane out-of-range case below.
+            break;
```
algesten, line 530: `Duration::as_micros()` gives us as 128 bit number. I think it is cleaner to cast to an i64 and rely on the rule below rather than introducing this `Result`.

Another idea is to keep the i128, however I have a hunch it's faster to cast to something that has native support on the CPU. In fact, that's probably why we are here to start off with.

OxleyS: It's not clear to me that an i64 would solve the issue - my impression is that it would just push the problematic range further out. It would be great if we could just `as i128` and be done with it, but since `as_micros()` specifically returns `u128` there's a danger zone around `u128::MAX`.

Of course, as we go from `i32` to `i64` to `i128`, the problematic cases get more and more unrealistic 😅 But despite the cleanliness hit, if nothing else I think there's value in the explicitness, signaling to readers that we've thought about this case.

OxleyS: As far as performance concerns of an explicitly checked cast, I'm confident that this is an easy case for a branch predictor, making it basically free.

algesten: > the problematic cases get more and more unrealistic 😅

This is the point, yes. Microseconds represented as 32 bit will handle a duration of ~4300 seconds or ~70 minutes. Having such a delta in this context (twcc) would be madness, but let's assume it can happen if you sleep the laptop or something. For 64 bit we get a duration of ~580 thousand years, at which point the battery of my laptop will be drained.

> As far as performance concerns of an explicitly checked cast

I wasn't thinking about the performance of your solve, but rather about casting to 64 bit and then perform those divisions. Dividing 128 bit on hardware without 128-bit support will be more expensive, hence I propose to cast to 64.

OxleyS: Okay then, I'll change the cast to 64.

OxleyS:
I've rewritten the second commit to just promote the integer width to `i64` instead of doing checked casts.

algesten (approved):

algesten:
Thanks!

PR #528, opened 2024-06-11
Update nack/twcc register after SRTP decrypt

algesten:
Currently we update the TWCC and Nack registers even if we don't manage to decrypt the SRTP and/or remove the padding. This means a corrupted UDP packet might still count as successful wrt to resends and bwe.


pthatcher:
I think this looks good.  We shouldn't do anything with a packet if SRTP decrypt fails.

algesten:
Cool. Thanks. I'll merge!

PR #529, opened 2024-06-12
Move StreamRx.reset_roc() to expect_stream_rx(..)

algesten:
Resetting the ROC can only happen when starting a new stream. There is
no point in having this as a function.


algesten:
This is the receiving side. There is no problem setting the ROC multiple times more than the SRTP might not decode. Let's leave the API as is.

Issue #530, opened 2024-06-18
Panic in DTLS layer


OxleyS:
One of our servers panicked at this line:
https://github.com/algesten/str0m/blob/3ec91e2981cf7de0a373c1033fa1c6d795b0972b/src/crypto/ossl/io_buf.rs#L14

Unfortunately I don't have a repro or much more information than that. What might be related is every now and then, we get an error like this coming back from `Rtc::handle_input()`:

> Dtls(Io(Custom { kind: InvalidData, error: Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 167773415, library: "SSL routines", function: "dtls1_read_bytes", file: "ssl/record/rec_layer_d1.c", line: 613, data: "SSL alert number 255" }]))) } }))

I have not been able to establish whether this is legit data that is getting misprocessed, or just random internet junk.

algesten:
Interesting! On an internet facing server we must of course accept junk. We could probably create a fuzz here which establishes the connection cleanly and then starts sending random UDP gunk maybe mixed with real packets.

algesten:
I suspect this was fixed with #531

PR #531, opened 2024-06-18
Allow IoBuffer to hold multiple packets

algesten:


OxleyS:
I notice the `Read` impl for `IoBuffer` asserts that the buffer is big enough to consume the entire `incoming` buffer. If this assertion did not trip, but nevertheless `self.incoming.is_empty()` was false on the next `set_incoming()` call, wouldn't that mean `read` was never called?

algesten:
@OxleyS yeah that's what I think too.

Would it be possible for you to test with this branch and that log message I just added? I want to see if this happens.

OxleyS:
Sure, I can't promise too much since the panic has so far been extremely rare, but I can certainly throw it up there and see what happens.

Other possibilities to consider are behavior if `handle_receive()` was called again after some sort of error, or if the read inside OpenSSL was spuriously interrupted by a signal (the latter highly unlikely, since `IoBuffer` is just a buffer of bytes).

OxleyS:
A small update on this - we've been running our servers with this fix since Wednesday last week, and so far haven't seen a panic nor the `Data remaining in IoBuffer` log line. We'll keep on running this and keeping an eye on it, though.

OxleyS:
It only happened once, so it must be some rather freak conditions that trigger it.
If we haven't seen a panic nor the log line, that means those conditions have not happened yet, correct? I think we would need to see one or the other to know if this fix is working.

algesten:
Yeah. Let's just give it time.

algesten:
@OxleyS I rebased this PR off latest main so you can get both this fix and the other log changes.

OxleyS:
While I admit that I haven't been super diligently checking (with a vacation putting me out of the office for a while), I still haven't seen any panics nor log lines.

How should we proceed? Personally I feel that whether it fixes #530 or not, this PR is good to have as a "just in case" guard (minus the logging of course). Especially since the happy path is almost certainly taken. I haven't done any benchmarking though.

algesten:
Yeah. Let's land it. I'll remove the logging and get it in later today. Thanks for checking!

PR #532, opened 2024-06-19
Fix panic in `DatagramRecv::try_from` with empty buffer

OxleyS:
`MultiplexKind` (and, by extension, callers such as `DatagramRecv`) didn't check for an empty buffer before attempting to read the first byte, causing a panic.


algesten (approved):

algesten:
Thanks!

PR #533, opened 2024-06-25
Change the `unix_time` function to return `libc::time_t` instead of a hard-coded `i64`

Ma233:
Change the `unix_time` function to return `libc::time_t` instead of a hard-coded `i64`.
This PR attempts to fix the following error when compiling to the armv7 platform.

```
cargo build --target armv7-unknown-linux-gnueabihf
   Compiling str0m v0.5.1 (/home/magine/str0m)
error[E0308]: mismatched types
   --> src/crypto/ossl/cert.rs:50:42
    |
50  |         let before = Asn1Time::from_unix(unix_time() - 3600)?;
    |                      ------------------- ^^^^^^^^^^^^^^^^^^ expected `i32`, found `i64`
    |                      |
    |                      arguments to this function are incorrect
    |
note: associated function defined here
   --> /home/magine/.cargo/registry/src/index.crates.io-6f17d22bba15001f/openssl-0.10.59/src/asn1.rs:330:12
    |
330 |     pub fn from_unix(time: time_t) -> Result<Asn1Time, ErrorStack> {
    |            ^^^^^^^^^
help: you can convert an `i64` to an `i32` and panic if the converted value doesn't fit
    |
50  |         let before = Asn1Time::from_unix((unix_time() - 3600).try_into().unwrap())?;
    |                                          +                  +++++++++++++++++++++

For more information about this error, try `rustc --explain E0308`.
error: could not compile `str0m` (lib) due to 1 previous error
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -928,6 +928,7 @@ dependencies = [
  "crc",
  "fastrand",
  "hmac",
+ "libc",
  "once_cell",
  "openssl",
  "openssl-sys",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ rust-version = "1.65"
 
 [features]
 default = ["openssl"]
-openssl = ["dep:openssl", "dep:openssl-sys"]
+openssl = ["dep:openssl", "dep:openssl-sys", "dep:libc"]
 _internal_dont_use_log_stats = []
 _internal_test_exports = []
 
@@ -32,6 +32,7 @@ combine = "4.6.6"
 # OPENSSL_NO_VENDOR=1 to override the feature flag vendored
 openssl = { version = ">=0.10.66", features = ["vendored"], optional = true }
 openssl-sys = { version = "0.9.80", optional = true }
+libc = { version = "0.2", optional = true }
 # STUN
 hmac = "0.12.1"
 crc = "3.0.0"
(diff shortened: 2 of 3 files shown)


algesten:
Hi @Ma233! Welcome to str0m.

If I understand this correctly. This happens because the target platform is a 32-bit one?

algesten:
@k0nserv @davibe thoughts on this?

Seems using libc::time_t is the only way to ensure compatibility for this call :/

I'm a bit reluctant to introduce a direct dep on libc (juggling more locked versions), but it might not be possible to avoid it.

https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/gnu/b64/mod.rs.html#17-27

k0nserv:
I was about to suggest what you did in #534 so it seems we are aligned. This will open us up to the y2k38 problem though. It's still 14 years away though, maybe by then all 32 bit targets are gone.

algesten (approved):

algesten:
The heuristics for deciding whether a platform uses a 32- or 64-bit `time_t` appears to be more complicated than determining whether it's a 32-bit target. There could be 32 bit libc platforms out therer that still use a 64 bit `time_t` (or vice versa, but that seems more awkward).

I think the only way to ensure we get the right value is to expose `libc` like this. I don't like it, but it is what it is.

algesten:
Alright. I think we should merge this.

PR #534, opened 2024-06-26
Type alias time_t like libc does

algesten:
Here's an alternative take on #533 that does not pull in libc.

Close #533

### DIFF
--- a/src/crypto/ossl/cert.rs
+++ b/src/crypto/ossl/cert.rs
@@ -95,13 +95,21 @@ impl OsslDtlsCert {
     }
 }
 
+// The goal here is to match the behavior in the libc crate without
+// actually needing to declare the libc crate for a type alias.
+// https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/gnu/b64/mod.rs.html#17-27
+#[cfg(all(target_arch = "aarch64", target_pointer_width = "32"))]
+type Time = i32;
+#[cfg(not(all(target_arch = "aarch64", target_pointer_width = "32")))]
+type Time = i64;
+
 // TODO: Refactor away this use of System::now, to instead go via InstantExt
 // and base the time on the first Instant. This would require lazy init of
 // Dtls, or that we pass a first ever Instant into the creation of Rtc.
 //
 // This is not a super high priority since it's only used for setting a before
 // time in the generated certificate, and one hour back from that.
-pub fn unix_time() -> i64 {
+pub fn unix_time() -> Time {
     SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
         .unwrap()


algesten:
@Ma233 would this solve your problem?

src/crypto/ossl/cert.rs
```diff
@@ -95,13 +95,21 @@ impl OsslDtlsCert {
     }
 }
 
+// The goal here is to match the behavior in the libc crate without
+// actually needing to declare the libc crate for a type alias.
+// https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/gnu/b64/mod.rs.html#17-27
+#[cfg(all(target_arch = "aarch64", target_pointer_width = "32"))]
```
Ma233, line 101:
```suggestion
#[cfg(all(target_arch = "arm", target_pointer_width = "32"))]
```

src/crypto/ossl/cert.rs
```diff
@@ -95,13 +95,21 @@ impl OsslDtlsCert {
...
+type Time = i32;
+#[cfg(not(all(target_arch = "aarch64", target_pointer_width = "32")))]
```
Ma233, line 103:
```suggestion
#[cfg(not(all(target_arch = "arm", target_pointer_width = "32")))]
```

src/crypto/ossl/cert.rs
```diff
@@ -95,13 +95,21 @@ impl OsslDtlsCert {
...
 // and base the time on the first Instant. This would require lazy init of
 // Dtls, or that we pass a first ever Instant into the creation of Rtc.
 //
 // This is not a super high priority since it's only used for setting a before
 // time in the generated certificate, and one hour back from that.
-pub fn unix_time() -> i64 {
+pub fn unix_time() -> Time {
     SystemTime::now()
         .duration_since(SystemTime::UNIX_EPOCH)
         .unwrap()
```
Ma233, line 115: Should also change `.as_secs() as i64` to `.as_secs() as Time` here.

Ma233 (commented):
@algesten Hi, the target_arch of `armv7-unknown-linux-gnueabihf` is `arm`. And there is a type conversion at the end of function that needs to be updated.

This modification can solve the compilation problem with target `armv7-unknown-linux-gnueabihf`. However, the situation in `libc` is more complicated, and I am worried about whether it is compatible with the interface of `rust-openssl` for other targets. But for now, there is no problem for me.

algesten:
> Hi, the target_arch of `armv7-unknown-linux-gnueabihf` is `arm`. And there is a type conversion at the end of function that needs to be updated.

Oh. If this works with your change (depending on libc) There must be more logic in the libc crate for arm. I guess me grabbing what I found here: https://docs.rs/libc/latest/src/libc/unix/linux_like/linux/gnu/b64/mod.rs.html#17-27 is not enough to cover all the variants of `time_t`?

Ma233:
@algesten The `aarm64` should be a 64-bit ARM system. Maybe just use `target_pointer_width = "32"`? Not sure...

algesten:
I think this is the wrong route.

PR #535, opened 2024-07-02
Remember max rx SeqNo per SSRC

algesten:
We got code for dynamically discovering SSRC changes. This is detected when a mid/rid or mid/pt header combo starts on one SSRC and then changes to another. 

As a special case, it could happen that the sender reuses an SSRC it used previously. 

A -> B -> A

The SRTP IV is derived from the extended sequence number. If one IV is reused even once, the encryption is broken. Hence the sender is not allowed to reuse sequence numbers. 

That means the above scenario, where we reuse a previously used SSRC, we must remember the old sequence number to be able to restart from the correct ROC.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,4 +1,4 @@
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
 use std::time::{Duration, Instant};
 
 use crate::bwe::BweKind;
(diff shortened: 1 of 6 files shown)


src/session.rs
```diff
@@ -409,6 +413,10 @@ impl Session {
             self.twcc_rx_register.update_seq(extended.into(), now);
         }
 
+        // Store largest seen seq_no for the SSRC. This is used in case we get SSRC changes
+        // like A -> B -> A. When we go back to A, we must keep the ROC.
```
k0nserv, line 417: I don't fully understand this. The ROC is per SSRC so why should it matter if two separate SSRCs have different values, outside of RTX repaired packets I don't think these should be combined

k0nserv (approved):

src/session.rs
algesten: It's not about combining. I'll explain in PR description.

davibe (approved):

Issue #490, opened 2024-03-22
Configurable ICE timeout values


thomaseizinger:
Currently, these are constants: https://github.com/algesten/str0m/blob/2326934668e0a81aafe3e42ff1ecbff2f033c5a5/src/io/stun.rs#L11-L14

Could we make these configurable at runtime via the `IceAgent`? Our fork currently adjusts these which makes it a bit tedious to keep rebasing :)

algesten:
Of course. Make setters!

PR #537, opened 2024-07-05
feat(ice): make timeout values configurable

thomaseizinger:
This makes the 3 components that make up the ICE timeout configurable. I tried to add some documentation to the setters to the best of my understanding of how this works internally.

Resolves: #490.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -4,9 +4,9 @@ use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
 
-use crate::io::{Id, StunClass, StunMethod, DATAGRAM_MTU_WARN};
+use crate::io::{Id, StunClass, StunMethod, StunTiming, DATAGRAM_MTU_WARN};
 use crate::io::{Protocol, StunPacket};
-use crate::io::{StunMessage, TransId, STUN_TIMEOUT};
+use crate::io::{StunMessage, TransId};
 use crate::io::{Transmit, DATAGRAM_MTU};
 use crate::util::NonCryptographicRng;
 
@@ -90,6 +90,9 @@ pub struct IceAgent {
 
     /// Statistics counter for the agent.
     stats: IceAgentStats,
+
+    /// The timing configuration for STUN bindings.
+    timing_config: StunTiming,
 }
 
 #[derive(Debug)]
@@ -269,6 +272,7 @@ impl IceAgent {
             nominated_send: None,
             stats: IceAgentStats::default(),
             timing_advance: Duration::from_millis(50),
+            timing_config: StunTiming::default(),
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/ice/agent.rs
```diff
@@ -316,6 +320,39 @@ impl IceAgent {
         }
     }
 
+    /// Sets the initial STUN **R**etransmission **T**ime**O**ut.
```
algesten, line 323: Nice one :) 

My first reaction was WTF. l33T 5p3ak?!

src/ice/agent.rs
```diff
@@ -316,6 +320,39 @@ impl IceAgent {
...
+    }
+
+    /// Sets the maximum number of retransmits for STUN messages.
+    ///
+    /// Defaults to 9.
+    pub fn set_max_stun_retransmits(&mut self, num: usize) {
+        self.timing_config.max_retransmits = num;
+    }
+
+    /// How long we at most tolerate missing replies for a candidate pair before considering it failed.8
```
algesten, line 351:
```suggestion
    /// How long we at most tolerate missing replies for a candidate pair before considering it failed.
```

src/ice/pair.rs
```diff
@@ -2,16 +2,12 @@ use std::collections::VecDeque;
...
-use crate::io::{Id, TransId, STUN_MAX_RTO_MILLIS};
+use crate::io::{Id, StunTiming, TransId};
 use crate::Candidate;
 
-const MIN_TIMEOUT: Duration = Duration::from_millis(STUN_MAX_RTO_MILLIS);
-
 // When running ice-lite we need a cutoff when we consider the remote definitely gone.
 const RECENT_BINDING_REQUEST: Duration = Duration::from_secs(15);
 
-#[derive(Default)]
```
algesten, line 14: What's the thinking here? Better to be explicit? Considering it adds all those lines below.

src/io/stun.rs
```diff
@@ -8,28 +8,61 @@ use crc::{Crc, CRC_32_ISO_HDLC};
...
+    }
+}
+
+// Consult libwebrtc for default values here.
+impl Default for StunTiming {
+    fn default() -> Self {
+        Self {
+            initial_rto: Duration::from_millis(250),
+            max_retransmits: 9,
+            max_rto: Duration::from_millis(3000),
```
algesten, line 63: We should mention this is 8000 in libwebrtc and we decided to lower it to detect gone peers faster.

algesten (approved):
Looks good. Some minor observations.

src/ice/pair.rs
thomaseizinger: The thinking was that setting `StunTimings` to default is semantically wrong because it needs to be copied over and set and not initialised.

algesten: Right. Ok.

src/ice/agent.rs
thomaseizinger: It looks better in the rendered docs :D

xnorpx:
Should this be available in RtcConfig as well? (I know we want tune some of these values)

algesten:
> Should this be available in RtcConfig as well? (I know we want tune some of these values)

Yeah. Let's make a follow up PR since Thomas doesn't use Rtc.

PR #539, opened 2024-07-10
fix/doc bitwhip

algesten:


algesten:
@Sean-Der does this look alright?

Sean-Der:
LGTM! Thanks for including it @algesten :)

PR #543, opened 2024-07-21
Add `playout_delay` builder function to `Writer`

UE2020:
This adds a function with the signature `pub fn playout_delay(mut self, min: MediaTime, max: MediaTime) -> Self` to the `Writer` that allows the minimum and maximum playout delay values to be set. Feel free to change the doc comment (I wasn't sure exactly how much information to put there).

### DIFF
--- a/src/media/writer.rs
+++ b/src/media/writer.rs
@@ -84,6 +84,14 @@ impl<'a> Writer<'a> {
         self
     }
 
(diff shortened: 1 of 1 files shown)


davibe (approved):

algesten:
Great! Thanks!

PR #548, opened 2024-07-25
ChannelConfig for data channels in SdpApi

uwuPyxl:
Make it possible to configure reliability and ordering when creating data channels through the SdpApi by having `SdpApi::add_channel` take a ChannelConfig.

It might make sense to add convenience methods for common configs e.g. `ChannelConfig::reliable(label)` or `SdpApi::add_reliable_channel(label)`.

With this change it's not clear how to create out-of-band negotiated channels, since on the side that will generate the initial answer applying added channels will create an offer. My guess is adding them after accepting the initial offer would work, but this should be documented.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -308,7 +308,9 @@ impl<'a> SdpApi<'a> {
         }
     }
 
-    /// Add a new data channel and get the `id` that will be used.
+    /// Add a new reliable ordered data channel and get the `id` that will be used.
+    ///
+    /// Use `add_channel_with_config` when unreliable or unordered data channels are preferred.
     ///
     /// The first ever data channel added to a WebRTC session results in a media
     /// of a special "application" type in the SDP. The m-line is for a SCTP association over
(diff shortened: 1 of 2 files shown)


algesten:
Hi @uwuPyxl, welcome to str0m!

This looks good! I strive to keep the SDP API I believe to be used the most as simple as possible, which means I'd like to keep `add_channel` taking just the `label` argument, but also add a second function `add_channel_with_config` which is the way you made it in this PR. This mirrors `RTCPeerConnection.createDataChannel(label)` – which should make it easier for new users to str0m to understand the API and also avoids a breaking change. Internally they should of course proxy to one impl.

If you make that change we can land this straight away.

uwuPyxl:
I just noticed that currently add_channel creates a reliable unordered channel. I am assuming this is unintentional and should be ordered instead?

algesten:
> I just noticed that currently add_channel creates a reliable unordered channel. I am assuming this is unintentional and should be ordered instead?

Oh does it? It should mirror `RTCPeerConnection.add_channel()`

uwuPyxl:
add_channel uses the default ChannelConfig. Since it is derived it uses the default of bool for the ordered field which is false. I added a custom default implementation for ChannelConfig that sets it to true instead.

algesten (approved):

algesten:
Looks great! Thanks!

PR #551, opened 2024-08-02
Add `extmap-allow-mixed` SDP attribute

alexlapa:
So I`ve been experimenting with AV1 SVC, and it seems that Chrome won't send  [Dependency Descriptor RTP Header Extension](https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension) if [a=extmap-allow-mixed](https://www.rfc-editor.org/rfc/rfc8285.html#section-6) is not signaled by the remote, so here it is.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -741,6 +741,7 @@ fn as_sdp(session: &Session, params: AsSdpParams) -> Sdp {
             typ: "BUNDLE".into(),
             mids,
         },
+        SessionAttribute::AllowMixedExts,
         SessionAttribute::MsidSemantic {
             semantic: "WMS".to_string(),
             stream_ids,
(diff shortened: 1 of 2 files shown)


algesten (commented):
Should we parse this too from the incoming descriptor?

alexlapa:
Only if you are going to optionally enable/disable RTP extensions that need this feature based on this, which is not the case at the moment. Janus [just adds this line](https://github.com/meetecho/janus-gateway/blob/master/src/sdp.c#L1519) regardless of anything.

lolgesten:
Alright. Can you make a code comment to that effect, linking to the janus impl? Then we merge.

lolgesten (approved):

algesten (approved):

algesten:
Thanks!

PR #553, opened 2024-08-06
Reduce the range of initial sequence number

xnorpx:
Q8 in: https://srtp.sourceforge.net/historical/faq.html


lolgesten:
@xnorpx Looks good. Can you shorten the commit message so it doesn't "line break"?

lolgesten (approved):

PR #554, opened 2024-08-06
Do not decrypt already received packets

algesten:
This is to protect str0m against SRTP replay attacks where already received packets are being repeated. Before this PR, this would force str0m to spend CPU decrypting it over and over again. With this PR, str0m checks the NACK register whether the packet is one we expect before doing the decryption.

### DIFF
--- a/src/streams/register_nack.rs
+++ b/src/streams/register_nack.rs
@@ -55,20 +55,21 @@ impl<'a> Iterator for NackIterator<'a> {
     type Item = NackEntry;
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.next = (self.next..=self.end).find(|s| self.reg.packet((*s).into()).needs_nack())?;
+        self.next =
+            (self.next..=self.end).find(|s| self.reg.packet_mut((*s).into()).needs_nack())?;
 
         let mut entry = NackEntry {
             pid: (self.next % U16_MAX) as u16,
             blp: 0,
         };
 
-        self.reg.packet(self.next.into()).nack_count += 1;
+        self.reg.packet_mut(self.next.into()).nack_count += 1;
         self.next += 1;
 
         for (i, s) in (self.next..self.end).take(16).enumerate() {
-            let packet = self.reg.packet(s.into());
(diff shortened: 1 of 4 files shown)


algesten:
@xnorpx this is the fix I believe.

xnorpx:
maybe worth a simple end2end tests

lolgesten:
> maybe worth a simple end2end tests

Problem is that there is no observable difference between the packet being decrypted vs the packet being dropped due to being a dupe.

src/streams/register_nack.rs
```diff
@@ -97,6 +98,20 @@ impl NackRegister {
...
+            // if we don't have initialized, we do want the first packet.
+            return true;
+        };
+
+        // behind the window
+        if seq < active.start {
+            return false;
+        }
+
+        !self.packet(seq).received || seq > active.end
```
xnorpx, line 112: how long is this window for active.end? I.e. if we had a 2 second burst loss would this seq number be outside the active window and we would drop it?

lolgesten: It's not a fixed time, but a fixed number of packets. We call it `MAX_MISORDER`, if packets are arriving in sequence, the window shrinks. If they come out of order, we grow up to `MAX_MISORDER`.

xnorpx: ok, so if we have:

SeqNo: 1,2,3,4 <crazy burst loss> 200, 201, 202, 203

the 200 packet will fit in the windows still and be inserted

lolgesten:
```rust
    #[test]
    fn nack_jump_larger_than_misorder() {
        let mut reg = NackRegister::new(None);
        for i in &[1, 2, 3, 4, 200, 201, 202, 203] {
            reg.update((*i).into());
        }

        assert_eq!(Some(103.into()..203.into()), reg.active)
    }
```

`MAX_MISORDER` is 100. So the active window goes from 103-203. I.e. we accept any packet as far back as 103. If you at this state run a nack report, you get 103-199 reported as missing. Sequence numbers 5-102 are silently forgotten.

PR #555, opened 2024-08-06
SRTP replay attack test

xnorpx:
Simple testcase to duplicate incoming rtp packets to simulate a srtp replay attack.


algesten (approved):
Did I miss this one?!

Anyways. Looks great!

Issue #556, opened 2024-08-07
Allow determinstic sorting of candidates in their SDP representation


thomaseizinger:
Upstream issue: https://github.com/firezone/firezone/pull/6200

Essentially, what is happening is that - because `foundation` is a hash - the lexical sort-order of candidates isn't stable in regards to their priority. It would be nice if it were!

algesten:
So you're using `to_sdp_string` and that sort order is not lexically the same as the priority?

thomaseizinger:
> So you're using `to_sdp_string` and that sort order is not lexically the same as the priority?

Correct!

thomaseizinger:
Because it is hashing the IP address and port which is different and I guess completely "randomises" the hash?

algesten:
Well. It's deterministic in that the same ip/port/proto will result in the same foundation.

thomaseizinger:
> Well. It's deterministic in that the same ip/port/proto will result in the same foundation.

Yes but two sets of candidates might have a different ordering, despite being the same candidates (1 host, 1 srflx, 1 relay for example), simply because their addresses are different.

algesten:
Ok so foundation is 1-32 "ice chars". The hash we produce is 64bit, so `u64::MAX.to_string().len()` is 20, which means we got 12 spare chars. The priority is 32 bit. If we encode as hex we get 8 chars.

Ergo. We could prepend the prio:

`<prio as hex><current foundation>`

thomaseizinger:
> Ok so foundation is 1-32 "ice chars". The hash we produce is 64bit, so `u64::MAX.to_string().len()` is 20, which means we got 12 spare chars. The priority is 32 bit. If we encode as hex we get 8 chars.
> 
> Ergo. We could prepend the prio:
> 
> `<prio as hex><current foundation>`

If that works for you, then I'll do that! :)

algesten:
I suppose we can make foundation calculation pluggable if this runs into problem. We will always have a bunch of leading 0. Can we avoid that? 🤔 

`000000ba483948395893849833`

thomaseizinger:
> I suppose we can make foundation calculation pluggable if this runs into problem. We will always have a bunch of leading 0. Can we avoid that? 🤔
> 
> `000000ba483948395893849833`

I was gonna do:

```rust
format!("{:x}{hash:x}", self.prio().to_be())
```

That being said though, prio is "higher is best" so it actually isn't the sort order I want :sweat_smile: 

[...]

PR #557, opened 2024-08-07
fix: ensure lexical ordering of SDP-formatted candidates follows priority

thomaseizinger:
Fixes: #556.

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -303,7 +303,7 @@ impl Candidate {
 
         let hash = hasher.finish();
 
-        format!("{:x}{hash:x}", self.prio().to_be())
+        format!("{:08x}{hash:x}", self.prio().to_be())
     }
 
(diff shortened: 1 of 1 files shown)


src/ice/candidate.rs
```diff
@@ -303,7 +303,7 @@ impl Candidate {
 
         let hash = hasher.finish();
 
-        hash.to_string()
+        format!("{:x}{hash:x}", self.prio().to_be())
```
lolgesten, line 306: I think you need to pad with 0 or you get:

* 10000
* ff

lolgesten (changes requested):

thomaseizinger:
I think the priorities that we generate are always already taking the full range but I added the padding anyway :)

lolgesten (approved):

algesten (approved):

PR #559, opened 2024-09-01
Make start of talkspurt information available for sample api

xnorpx:
When implementing an Audio Jitter buffer, information about new talkspurts can be used to do adjustments to the buffer.

Marker bit is available in the rtp api, but for the sample(frame) api we need it exposed.

### DIFF
--- a/src/media/event.rs
+++ b/src/media/event.rs
@@ -94,7 +94,7 @@ pub struct MediaData {
     /// quantity. The numerator is the timestamp field from the RTP header, the denominator
     /// depends on whether this is an audio or video packet.
     ///
-    /// For audio the timebase is 48kHz for video it is 90kHz.
+    /// For audio the timebase is often 48kHz for video it is 90kHz.
     pub time: MediaTime,
 
     /// The time of the [`Input::Receive`][crate::Input::Receive] of the first packet that caused this MediaData.
@@ -131,6 +131,12 @@ pub struct MediaData {
     ///
     /// If no Sender Report(SR) has been received this is [`None`].
     pub last_sender_info: Option<SenderInfo>,
+
+    /// First packet of a talkspurt, that is the first packet after a silence period during
+    /// which packets have not been transmitted contiguously.
+    ///
+    /// For audio only when dtx or silence suppression is enabled.
+    pub audio_start_of_talk_spurt: bool,
 }
 
 /// Details for an incoming a keyframe request (PLI or FIR).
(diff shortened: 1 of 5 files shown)


xnorpx:
@algesten not sure if this one fits well in the MediaData struct, but it also doesn't feel it should sit in extension.

src/media/event.rs
```diff
@@ -112,6 +112,12 @@ pub struct MediaData {
     /// For audio this flag most likely doesn't matter.
     pub contiguous: bool,
 
+    /// First packet of a talkspurt, that is the first packet after a silence period during
+    /// which packets have not been transmitted contiguously. 
+    /// 
+    /// For audio only when dtx or silence suppression is enabled.
+    pub audio_start_of_talk_spurt: bool,
+
```
algesten, line 120: Move this to after `last_sender_info`

src/media/mod.rs
```diff
@@ -265,6 +265,7 @@ impl Media {
                     network_time: dep.first_network_time(),
                     seq_range: dep.seq_range(),
                     contiguous: dep.contiguous,
+                    audio_start_of_talk_spurt: dep.start_of_talkspurt(),
```
algesten, line 268: Add a check for `codec.spec().codec.is_audio()` here to avoid setting this to true when marker is used as end-of-video-frame.

algesten (commented):
Looks alright. A couple of small things.

algesten:
What is "dtx or silence suppression", is it something negotiated?

xnorpx:
> What is "dtx or silence suppression", is it something negotiated?

DTX or silence suppression is an optimization which indicates that the sender is "quiet" and it will only send update packets every 400ms in DTX mode and then it will indicate with a marker bit when it goes out of DTX mode.

DTX is negotiated in SDP for Opus but it could be received and handled without negotiation. Not sure how silence suppression is done WebRTC or negotiated, would need to look it up. But silence suppression is similar to dtx but Opus handles the silence transition in-band while Silence suppression switches to another special Comfort Noise payload type.

I need to think about how we can add support for sending and maybe SDP support as well.

xnorpx:
Thanks I will fix comments and add one simple integration test and then publish later today.

xnorpx:
@algesten ready for another round.
Added:
- Some cleanup in depacketized
- Comments
- Added a simple test

src/media/mod.rs
```diff
@@ -268,6 +268,11 @@ impl Media {
                     ext_vals: dep.ext_vals().clone(),
                     codec_extra: dep.codec_extra,
                     last_sender_info: dep.first_sender_info(),
+                    audio_start_of_talk_spurt: if codec.spec().codec.is_audio() {
+                        dep.start_of_talkspurt()
+                    } else {
+                        false
+                    },
```
algesten, line 275: No need for if. This can be a `&&`

algesten (approved):
Looks good. Just one thing to fix and then we can merge.

PR #561, opened 2024-09-05
Fix tests for rustc 1.82-beta

algesten:
rustc 1.82-beta in [this commit](https://github.com/rust-lang/rust/commit/0ba9db87e61adcfd9a978188f61c20d9b423a099) changes something about how hashes are calculated (likely [this PR](https://github.com/rust-lang/rust/pull/128946)). Since our tests have pinned hash values, we get a breakage.

Since we uphold MSRV 1.65, we can't pin the hash in the test since we will get broken CI tests when we ensure the MSRV compiles.

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -566,16 +566,23 @@ mod tests {
         let socket_addr = "1.2.3.4:9876".parse().unwrap();
         let mut candidate = Candidate::host(socket_addr, Protocol::Udp).unwrap();
         assert_eq!(
-            serde_json::to_string(&candidate).unwrap(),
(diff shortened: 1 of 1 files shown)


thomaseizinger (commented):
Should we switch to a stable hash-algorithm instead? We already depend on the `sha1` crate so we could use that?

algesten:
Could we chop it down to not be super long and still preserve enough entropy?

thomaseizinger:
We also depend on `crc` which allows for a 16-byte output. That allows us to stay within the 32 char limit of foundation. Could use that?

PR #566, opened 2024-09-26
Fix RTX stops working after packet loss spike

alexlapa:
So i've noticed remote receivers sending PLIs too often even after minor packet loss. Debugging NACKs and RTXs showed that everything actually works fine for some time and at some point receivers start flooding SFU with NACKs  for that same packets again and again and no RTXs are being sent at all which results in receiver ending up with a PLI request.

So whats going on is:
1. At some point packet loss spikes and RTX ratio exceeds 15%.
2. This results in `poll_packet_resend` always returning `None` because of [this](https://github.com/algesten/str0m/blob/35d9bbafa1f4a6146c09be87fdea8a94f5c58ae5/src/streams/send.rs#L545-L548)
3. Since no resends are made it never get to [this point](https://github.com/algesten/str0m/blob/35d9bbafa1f4a6146c09be87fdea8a94f5c58ae5/src/streams/send.rs#L573), so RTX ratio is not being recalculated [here](https://github.com/algesten/str0m/blob/35d9bbafa1f4a6146c09be87fdea8a94f5c58ae5/src/util/value_history.rs#L38).
4. At this point it is basically stuck with RTX ratio locked at `> 0.15` with the only option of it being fixed is if bytes transmitted sum becoming high enough which does not normally happen.

So i believe that ratio recalculation is supposed to happen during the `ValueHistory::sum()` call.

And here is before and after videos of how that works:

<details>
<summary>Before</summary>

[before.webm](https://github.com/user-attachments/assets/0da802e3-16cd-44ad-a86e-819c4a3aafae)

</details>

<details>
<summary>After</summary>

[after.webm](https://github.com/user-attachments/assets/d755d152-e803-49bc-a4af-8794203b228d)

</details>

On the after video tx sum drop is when SVC kicks in so ignore that.

Another thing is it would be great to be able to configure that `0.15` RTX cache drop ratio, mind if i do it in another PR?Probably as an additional argument to `StreamTx::set_rtx_cache`, so it will be `fn set_rtx_cache(&mut self, max_packets: usize, max_age: Duration, rtx_cache_drop_ratio: Option<f32>)` with `None` completely disabling this mechanic.

### DIFF
--- a/src/util/value_history.rs
+++ b/src/util/value_history.rs
@@ -35,12 +35,12 @@ where
     pub fn push(&mut self, t: Instant, v: T) {
         self.value += v;
         self.history.push_back((t, v));
-        self.drain(t);
     }
 
-    /// Returns the sum of all values in the history up to max_time
-    /// This is more efficient than sum_since() as it does not need to iterate over the history
-    pub fn sum(&self) -> T {
+    /// Returns the sum of all values in the history up to now - max_time.
+    pub fn sum(&mut self, now: Instant) -> T {
+        self.drain(now);
+
         self.value
     }
 
@@ -63,7 +63,7 @@ mod test {
     use super::ValueHistory;
 
     #[test]
-    fn test() {
+    fn with_value_test() {
         let now = Instant::now();
 
         let mut h = ValueHistory {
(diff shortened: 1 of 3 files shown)


davibe (approved):

algesten:
@alexlapa great find! Super nice improvement!

For sanity, I think we should change the impl a bit differently. It seems intuitive that the `ValueHistory` would be pruned on `push`, which already is a `&mut` and that's where I'd expect it to happen. However, as you shown, that's not a good idea.

Having it prune on `sum()` seems a little less intuitive. You could maybe argue that you want "a sum up until this point", but to me "calculating a sum" does not sound like a mutation.

I would prefer you drop the pruning from both `push` and `sum`, and expose `drain`, to make it explicit at which point this happens.

I'd also rename and change the signature of `drain` (since that has a specific meaning in rust std api). Something like `purge` or `purge_old`. It should also not return `Option<()>` since that value is of no consequence to the caller, it's simply there to allow one use of `?`.

algesten (changes requested):
See comment

src/util/value_history.rs
```diff
@@ -35,23 +35,26 @@ where
...
 
-    fn drain(&mut self, t: Instant) -> Option<()> {
-        while t.duration_since(self.history.front()?.0) > self.max_time {
+    /// Recalculates sum purging values older than `now - max_time`.
+    pub fn purge_old(&mut self, now: Instant) {
+        while {
+            let Some(front_t) = self.history.front().map(|v| v.0) else {
+                return;
+            };
+            now.duration_since(front_t) > self.max_time
```
algesten, line 52: wow. I never seen you can have expressions like this here. 😎

algesten (approved):

alexlapa:
@algesten ,

And what your opinion on this?

> Another thing is it would be great to be able to configure that 0.15 RTX cache drop ratio, mind if i do it in another PR? Probably as an additional argument to StreamTx::set_rtx_cache, so it will be fn set_rtx_cache(&mut self, max_packets: usize, max_age: Duration, rtx_cache_drop_ratio: Option<f32>) with None completely disabling this mechanic.

algesten:
@alexlapa i like it. There's some overlap with functions that can disable RTX depending on the SDP negotiation, but I think it's worth a PR.

PR #568, opened 2024-09-27
refactor(ice): always use latest timing config

thomaseizinger:
Currently, the `StunTiming` configuration gets copied into every new `CandidatePair` that is created. This means that changes to the timing configuration afterwards are not reflected in existing candidate pairs.

To fix this, we remove the `StunTiming` field from `CandidatePair` and instead pass in a reference everywhere where it is needed.

The usecase that I want to support with this is to allow lowering the frequency of STUN bindings in certain situations like idle connections. When a connection is not in use (i.e. there is no other traffic being sent over the UDP socket), detecting a failed connection isn't as time critical because the user isn't using the application. This allows the CPU to go to sleep and preserves battery on e.g. mobile devices.

### DIFF
--- a/src/ice/pair.rs
+++ b/src/ice/pair.rs
@@ -46,8 +46,6 @@ pub struct CandidatePair {
 
     /// State of nomination for this candidate pair.
     nomination_state: NominationState,
-
-    timing_config: StunTiming,
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
@@ -114,13 +112,12 @@ impl Default for PairId {
 }
 
 impl CandidatePair {
-    pub fn new(local_idx: usize, remote_idx: usize, prio: u64, timing_config: StunTiming) -> Self {
+    pub fn new(local_idx: usize, remote_idx: usize, prio: u64, timing_config: &StunTiming) -> Self {
         CandidatePair {
             local_idx,
             remote_idx,
             prio,
             binding_attempts: VecDeque::with_capacity(timing_config.max_retransmits() + 1),
-            timing_config,
             id: Default::default(),
             valid_idx: Default::default(),
             state: Default::default(),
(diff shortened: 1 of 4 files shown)


src/ice/pair.rs
```diff
@@ -114,13 +112,12 @@ impl Default for PairId {
 }
 
 impl CandidatePair {
-    pub fn new(local_idx: usize, remote_idx: usize, prio: u64, timing_config: StunTiming) -> Self {
+    pub fn new(local_idx: usize, remote_idx: usize, prio: u64, timing_config: &StunTiming) -> Self {
         CandidatePair {
             local_idx,
             remote_idx,
             prio,
             binding_attempts: VecDeque::with_capacity(timing_config.max_retransmits() + 1),
```
algesten, line 120: What if we reintroduce the `STUN_MAX_RETRANS` constant (it's still referenced in doc/comments I realize), and do this capacity as `2 * STUN_MAX_RETRANS`. That way we can avoid needing the timing config for the constructor.

algesten (approved):
Looks good!

Ideally I'd like to avoid passing this in the constructor. I think we can just make a best effort on the capacity.

src/ice/pair.rs
thomaseizinger: You mean with the default value? Because the max retransmissions are configurable now.

lolgesten: Yeah, so the 9 goes back into a constant, and it's used from the configuration.

This call site can use 2 * 9 which means _most_ configurations would never incur an extra allocation here due to the capacity being too low.

lolgesten:
Great. Can you do a search for `STUN_MAX_RETRANS` and change those stale doc comments too?

thomaseizinger:
> Great. Can you do a search for `STUN_MAX_RETRANS` and change those stale doc comments too?

Done!

algesten:
Excellent! Thanks!

thomaseizinger:
Thanks for the prompt turnaround (as always)! :)

algesten:
I aim to please :)

thomaseizinger:
In case you are curious, here is the upstream feature: https://github.com/firezone/firezone/pull/6845.

PR #570, opened 2024-10-07
Configure RTX ratio cap

alexlapa:
https://github.com/algesten/str0m/pull/566#issuecomment-2377171154

> Another thing is it would be great to be able to configure that `0.15` RTX cache drop ratio, mind if i do it in another PR? Probably as an additional argument to `StreamTx::set_rtx_cache`, so it will be `fn set_rtx_cache(&mut self, max_packets: usize, max_age: Duration, rtx_cache_drop_ratio: Option<f32>)` with `None` completely disabling this mechanic.

### DIFF
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -38,7 +38,7 @@ const MIN_SPURIOUS_PADDING_SIZE: usize = 50;
 
 pub const DEFAULT_RTX_CACHE_DURATION: Duration = Duration::from_secs(3);
 
-pub const DEFAULT_RTX_CACHE_DROP_RATIO: Option<f32> = Some(0.15f32);
+pub const DEFAULT_RTX_RATIO_CAP: Option<f32> = Some(0.15f32);
 
 /// Outgoing encoded stream.
 ///
@@ -104,7 +104,8 @@ pub struct StreamTx {
     /// sending spurious resends as padding.
     rtx_cache: RtxCache,
 
-    rtx_cache_drop_ratio: Option<f32>,
+    /// Determines retransmitted bytes ratio value to clear queued resends.
+    rtx_ratio_cap: Option<f32>,
 
     /// Last time we produced a SR.
     last_sender_report: Instant,
@@ -150,7 +151,11 @@ pub(crate) struct StreamTxStats {
     rtt: Option<f32>,
     /// losses collecter from RR (known packets, lost ratio)
     losses: Vec<(u64, f32)>,
+
+    /// `None` if `rtx_ratio_cap` is `None`.
     bytes_transmitted: Option<ValueHistory<u64>>,
+
+    /// `None` if `rtx_ratio_cap` is `None`.
     bytes_retransmitted: Option<ValueHistory<u64>>,
 }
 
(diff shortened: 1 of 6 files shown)


lolgesten:
When would we want this to be `None`?

alexlapa:
You mean why not just use value bigger than 1.0? It's just seems to be a more obvious off switch. I dont mind changing this if you want.

Or do you mean why would anyone want to disable this functionality? Well, i consider disabling it since its up to receiver to decide whether it need some specific packet or not, however I'm still testing this.

> It is not expected that a receiver will send a NACK for every lost RTP packet; rather, it needs to consider the cost of sending NACK feedback and the importance of the lost packet to make an informed decision on whether it is worth telling the sender about a packet-loss event.

algesten:
@davibe can provide more context, but I think this was put into place for _really bad connections_, where we risk crowding the network with resends rather than real packets. I think there's definitely a case where the client keeps nacking so much that it makes no sense to just try and fulfill that.

davibe:
Packets can be lost due to bandwidth constraints (or be or too late, which is the same). Retransmitting "too much" uses additional bandwidth which further exceeds the connection capabilities causing even more loss. This can happen also in scenarios that are not delay-sensitive because bandwidth may become not-enough indefinitely.

In our internal products - without this arbitrary 0.15 limit - we found that varying network conditions would get the peer stuck in a loop where str0m would continuously flood them with retransmissions and big keyframes while they kept nacking and sending PLIs as they were unable to recover.

Maybe what we need is to evolve this 0.15 limit into something that is more elaborate and takes more things into account (what is the bw of the peer, how much are we overusing, are we about to send a keyframe, ...). In absence of a more elaborate logic, configurability may be useful to some.

I thought 15% was quite a high limit, but public wifis may surprise. If a network has more than 15% loss but can still sustain high speeds, then 15% is not the best option. This is a dynamic per-peer consideration. This PR seems to allow to change the limit dynamically via direct api but it does not handle the change smoothly (the rtx cache is dropped on reconfiguration). Maybe it's enough for now.

src/streams/send.rs
```diff
@@ -209,10 +237,27 @@ impl StreamTx {
...
+    pub fn set_rtx_cache(
+        &mut self,
+        max_packets: usize,
+        max_age: Duration,
+        rtx_ratio_cap: Option<f32>,
+    ) {
         // Dump old cache to avoid having to deal with resizing logic inside the cache impl.
         self.rtx_cache = RtxCache::new(max_packets, max_age);
+        if rtx_ratio_cap.is_some() {
+            self.stats.bytes_transmitted = Some(ValueHistory::default());
```
davibe, line 254: I am not sure why we want to reset the stats when the stream is configured. If we allow reconfiguring the rtx-cache of the stream I think the stats should not be reset.

alexlapa: Well `RtxCache` is dropped, and since these stats don't do anything else i thought it would make sense, but you are probably right. Updated.

alexlapa:
@davibe ,

Is there still any issues with this PR?

algesten:
Thanks for the reminder. I'll
do a final review later today.

algesten (approved):

algesten:
Thanks!

PR #572, opened 2024-10-13
feat: add marker bit when packetizing opus.

### LINKED ISSUE #125: Support for Opus DTX
Currently there is no support for dtx.

Add:

Support for dtx in FormatParams
Change Opus packetizer to be able to write markerbit.

ramyak-mehra:
Fixes #125

### DIFF
--- a/src/packet/opus.rs
+++ b/src/packet/opus.rs
@@ -1,10 +1,20 @@
 use super::{CodecExtra, Depacketizer, MediaKind, PacketError, Packetizer};
 
 /// Packetizes Opus RTP packets.
-#[derive(Default, Debug, Copy, Clone)]
+#[derive(Debug, Copy, Clone)]
 pub struct OpusPacketizer {
     // stores if a marker was previously set
     marker: bool,
+    use_dtx: bool,
+}
+
+impl OpusPacketizer {
+    pub fn new(use_dtx: bool) -> Self {
+        Self {
+            marker: false,
+            use_dtx,
+        }
+    }
 }
 
 impl Packetizer for OpusPacketizer {
@@ -26,6 +36,9 @@ impl Packetizer for OpusPacketizer {
     }
 
     fn is_marker(&mut self, data: &[u8], previous: Option<&[u8]>, last: bool) -> bool {
+        if !self.use_dtx {
+            return false;
+        }
         // any non silenced packet would generally have more than 2 byts
         let mut is_marker = data.len() > 2;
 
@@ -102,7 +115,7 @@ mod test {
 
     #[test]
     fn test_opus_payload() -> Result<(), PacketError> {
-        let mut pck = OpusPacketizer::default();
+        let mut pck = OpusPacketizer::new(true);
         let empty = &[];
         let payload = &[0x90, 0x90, 0x90];
 
@@ -117,6 +130,43 @@ mod test {
         // Positive MTU, small payload
         let result = pck.packetize(2, payload)?;
         assert_eq!(result.len(), 2, "Generated payload should be the 2");
+
+        Ok(())
+    }
+
+    #[test]
+    fn test_opus_packetizer_dtx() -> Result<(), PacketError> {
+        // packetizer with dtx on
+        let mut pck = OpusPacketizer::new(true);
+
+        let payload = &[0x90, 0x90, 0x90];
+
+        // Start of talking spurt, marker bit is set.
+        let is_marker = pck.is_marker(payload, None, false);
+        assert!(is_marker);
+        assert!(pck.marker);
+
+        // More talking so is_marker should be false.
+        let is_marker = pck.is_marker(payload, None, false);
+        assert!(!is_marker);
+        assert!(pck.marker);
+
+        // silence packet inserted, internal packetizer state should be reset.
+        let is_marker = pck.is_marker(&[], None, false);
+        assert!(!is_marker);
+        assert!(!pck.marker);
+
+        // talking start again, marker bit is set.
+        let is_marker = pck.is_marker(payload, None, false);
+        assert!(is_marker);
+        assert!(pck.marker);
+
+        let mut pck = OpusPacketizer::new(false);
+
+        // Start of talking spurt, since dtx is false marker should be false
+        let is_marker = pck.is_marker(payload, None, false);
+        assert!(!is_marker);
+        assert!(!pck.marker);
 
         Ok(())
     }
--- a/src/packet/opus.rs
+++ b/src/packet/opus.rs
@@ -2,7 +2,10 @@ use super::{CodecExtra, Depacketizer, MediaKind, PacketError, Packetizer};
 
 /// Packetizes Opus RTP packets.
 #[derive(Default, Debug, Copy, Clone)]
-pub struct OpusPacketizer;
+pub struct OpusPacketizer {
+    // stores if a marker was previously set
+    marker: bool,
+}
 
 impl Packetizer for OpusPacketizer {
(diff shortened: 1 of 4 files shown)


xnorpx:
So markerbit should be set at "start" of talkspurt. Is it true that the first packet in a talkspurt is 1 byte?

xnorpx:
https://datatracker.ietf.org/doc/html/rfc3551#section-4.1

ramyak-mehra:
I might have read some details wrong, i actually added it at the end of a talk spurt, fixed that

xnorpx:
You can rename this file to test_talk_startspurt: https://github.com/algesten/str0m/blob/main/tests/rtp_to_frame.rs

Then validate that marker bit get's added.

Ideally test the different combinations of rtp/frame api.

xnorpx:
@algesten should some api be added for this, so in case dtx is set to false marker bit will never be set and it will not be negotiated over sdp.

xnorpx:
> Can we really auto-detect talk spurts without some audio analysis? I mean just because there's some noise on the line, doesn't mean there is a talk spurt starting. How does libWebRTC do this? (or pion if they support it)

Audio analysis is done in Opus so it will handle this.

I need to double check the packet pattern that Opus emits not sure this logic will work.

xnorpx:
I don't think Pion supports this, best would be to check how libwebrtc set the marker bit on Opus packages.

ramyak-mehra:
I took the inspiration from [rtpopuspayloader](https://gitlab.freedesktop.org/gstreamer/gst-plugins-good/-/blob/discontinued-for-monorepo/gst/rtp/gstrtpopuspay.c#L332)

>so in case dtx is set to false marker bit will never be set and it will not be negotiated over sdp.
I was thinking of passing down the UseDtx flag from sdp to codec but never found any other codec doing it, so skipped it for now. I did see something like `H264CodecExtra` to pass down options to de packetizer should something like that exists for packetizer as well?

Or probably at the start of initialising the struct codec params can be passed

ramyak-mehra:
I didn't find a way to get access to previous packet inside the opus packetizer, since we don't cache it (something about audio packets not being nackable), so i changed the implementation a bit.

I have also added a rtp level test for the same.

davibe:
For doing this only when dtx is enabled: we have `FormatParams::use_dtx` inside `CodecSpec`, maybe a simple solution would be impl this around here https://github.com/algesten/str0m/blob/5c4c9d54bb2331ff8a15a2e62e2575f6ec82286d/src/packet/payload.rs#L58
passing the CodecSpec down from the caller ?

src/packet/opus.rs
```diff
@@ -23,8 +26,23 @@ impl Packetizer for OpusPacketizer {
     }
 
     fn is_marker(&mut self, data: &[u8], previous: Option<&[u8]>, last: bool) -> bool {
-        // TODO: dtx
-        false
+        // any non silenced packet would generally have more than 2 byts
```
davibe, line 29: If there is one single marker packet (`self.marker = true`), then silence for 15 minutes, then again some audio (`is_marker = true`), this new audio packet would not go out as `marker` (bug)

davibe, line 29: `s/byts/bytes/`

ramyak-mehra: in case of no packets for a while there are two scenarios at play
* Packet loss: The speaker might still be speaking so can't determine for sure.
* Stopped speaker: In this case the packets would still be produced but at a lower frequency. This is also taken care of.

lolgesten:
@ramyak-mehra when we instantiate the packetizer, the fmtp parameters are available:

```
            Payloader::new(params.spec)
```

The `spec` here is `CodecSpec`, which has a field for `FormatParams`, which in turn has the `use_dtx` setting. https://github.com/algesten/str0m/blob/5c4c9d54bb2331ff8a15a2e62e2575f6ec82286d/src/format.rs#L145

I think all is in place for that already.

ramyak-mehra:
while we can pass it down i want to understand the rational behing passing extra_codec_params in each [depacketize](https://github.com/algesten/str0m/blob/5c4c9d54bb2331ff8a15a2e62e2575f6ec82286d/src/packet/mod.rs#L108) call.
Are there para meters that change with each packet thats why its there? Do we want to enable both kinds of approach passing down with each call as well as at initilizing?

davibe:
No. I meant it can be done from the caller of is_marker(). Like: marker = if audio && !dtx { false } else { packetizer…

algesten:
> while we can pass it down i want to understand the rational behing passing extra_codec_params in each [depacketize](https://github.com/algesten/str0m/blob/5c4c9d54bb2331ff8a15a2e62e2575f6ec82286d/src/packet/mod.rs#L108) call. Are there para meters that change with each packet thats why its there? Do we want to enable both kinds of approach passing down with each call as well as at initilizing?

I don't understand why we are talking about the depacketizer? This PR is about the packetizer, right?

> No. I meant it can be done from the caller of is_marker(). Like: marker = if audio && !dtx { false } else { packetizer…

I mean that this has nothing to do with the code calling the packetizer. The `use_dtx` function is dependent on which packetizer we are talking about. This should be passed into creating the `OpusPacketizer` (as I illustrated above), and be remembered internally for that packetizer only.

davibe:
If we do it via OpusPacketizer constructor I th we have take care of dynamic updates somehow. DTX can be activated and deactivated via js apis and user settings during an ongoing call. Currently payloaders are stored [based on `(pt, rid)`](https://github.com/algesten/str0m/blob/5c4c9d54bb2331ff8a15a2e62e2575f6ec82286d/src/media/mod.rs#L355-L359) so if an SDP renegotiation changes `a:fmtp:...` adding `usedtx` I am afraid we would not pick up the change, or do we ?.

ramyak-mehra:
We can just ignore the sdp use_dtx, if it's off the decoder on the other side should just ignore the marker bit

algesten:
> I think we have take care of dynamic updates somehow. DTX can be activated and deactivated via js apis and user settings during an ongoing call.

> Thogh I am not sure how do we handle sdp re negotiations in this library will take a look

I'm not convinced this is an attribute that can be changed runtime. AFAIK, the only thing you can change on an already negotiated m-line is the direction.

However even if it can be changed runtime, why would you want that? I think we can just say _str0m does not change this dynamically_ and we could potentially detect dynamic changes and reject them in incoming SDP.

ramyak-mehra:
Updated the PR to pass down use_dtx and added tests on the packetizer itself.
On a side note:
I see we create a new payloader everytime we want to write. Seems a bit redundant though or am i missing something?

ramyak-mehra:
I was reading more on the whole "talk spurt" thing and its not a very well defined thing. Yes you set the marker bit on start of a talk spurt but to detect talk spurt there are various alogorithms which seems out of scope for the project itself. There should ideally be a way to set the marker bit from outside the packetizer which seems like a bigger change

algesten:
> I was reading more on the whole "talk spurt" thing and its not a very well defined thing. Yes you set the marker bit on start of a talk spurt but to detect talk spurt there are various alogorithms which seems out of scope for the project itself. There should ideally be a way to set the marker bit from outside the packetizer which seems like a bigger change

I don't think it's too bad. Check out the `Writer` struct. It's a builder pattern and you can see how we implement stuff like `audio_level`. This is similar, but the data is not transported in an RTP header extension.

https://docs.rs/str0m/0.6.3/src/str0m/media/writer.rs.html#74-78

algesten:
Cleaning up. We can re-open this anew if it's still relevant.

PR #573, opened 2024-10-21
Correctly handle per m-line TWCC

alexlapa:
So I've been noticing that some receivers report `PeerStats.egress_loss_fraction` sitting at 20-40% with no other indications of packet loss,  and `MediaEgressStats.loss` reporting ~0%. Further investigation showed that it only happens with Firefox receivers cause Firefox does not support TWCC for audio m-lines. Example SDP:

```
v=0
o=mozilla...THIS_IS_SDPARTA-99.0 812743268998139128 0 IN IP4 0.0.0.0
...
a=sendrecv
a=extmap:1 urn:ietf:params:rtp-hdrext:ssrc-audio-level
a=extmap:2/recvonly urn:ietf:params:rtp-hdrext:csrc-audio-level
a=extmap:3 urn:ietf:params:rtp-hdrext:sdes:mid
...
a=sendrecv
a=extmap:3 urn:ietf:params:rtp-hdrext:sdes:mid
a=extmap:4 http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time
a=extmap:5 urn:ietf:params:rtp-hdrext:toffset
a=extmap:6/recvonly http://www.webrtc.org/experiments/rtp-hdrext/playout-delay
a=extmap:7 http://www.ietf.org/id/draft-holmer-rmcat-transport-wide-cc-extensions-01
...
```

So Firefox does not report audio in its TWCC reports, so str0m's TWCC-based loss calculation desides that all audio packets are lost.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
   * New add_channel_with_config() for configured data channels #548
   * Fix RTX stops working after packet loss spike #566
   * Configure RTX ratio cap via `StreamTx::set_rtx_cache` #570
+  * Correctly handle per m-line TWCC #573
 
 # 0.6.1
   * Force openssl to be >=0.10.66 #545
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1012,9 +1012,10 @@ fn update_session(session: &mut Session, sdp: &Sdp) {
         .id_of(Extension::TransportSequenceNumber)
         .is_some();
 
-    // Since twcc feedback is session wide and not per m-line or pt, we enable it if
-    // there are _any_ m-line with a a=rtcp-fb transport-cc parameter and the sequence
-    // number header is enabled.
+    // Since twcc feedback is session wide we enable it if there are _any_
+    // m-line with a a=rtcp-fb transport-cc parameter and the sequence number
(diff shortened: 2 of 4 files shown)


k0nserv:
Oh nice, I've been spotting these high loss reports periodically and have been meaning to investigate it. This sounds like what I've been seeing too!

alexlapa:
Also I've been considering caching if TWCC and AbsSendTime are enabled in `Media` to avoid iterating over `ExtensionMap` on each packet, but I'm really not sure that its worth it: iterating over 3-6 elements comparing enums should be fast enough. So let me know if you'd like that changed.

k0nserv:
Can you split out the change to AbsSendTime in a separate commit and mention it in the changelog? 

> but I'm really not sure that its worth it: iterating over 3-6 elements comparing enums should be fast enough. So let me know if you'd like that changed.

I've benchmarked this before(only on macOS aarch64) and for lists of < 10 elements direct iteration is faster than hash lookups. If you use a faster hasher a hashet is faster though

k0nserv:
I'm happy with this, @algesten do you wanna weigh in too?

alexlapa:
> Can you split out the change to AbsSendTime in a separate commit and mention it in the changelog?

Sure, removed it.

PR #575, opened 2024-10-22
Correctly handle per m-line Absolute Send Time

alexlapa:
Related to #573

Remote peer might have [`Absolute Send Time`](http://www.webrtc.org/experiments/rtp-hdrext/abs-send-time) disabled for some m-line (like Firefiox doest not support abs-send-time on audio lines), so str0m should not add it to RTP header in this case.


algesten (approved):

algesten:
Looks good! Just a lil lint problem.

algesten:
Thanks!

PR #576, opened 2024-10-23
Dedupe remote ICE candidates

k0nserv:
The changes in `5c4c9d5` caused problems because we'd end up creating new identical pairs. For example, if an existing pair, `0-0`, exists as local: 192.168.1.33:5445, remote: 192.168.1.33:63461. Then, after processing a new offer another remote candidate would be added for 192.168.1.33:63461, call this pair `0-1`. When determining which pair an incoming binding request belonged too we'd then identify `0-1` instead of `0-0`(because `max_by_key`) returns the last value on ties. In Chrome this would work correctly, but in Firefox connectivity would be lost.

Ultimately adding a new remote candidate that is exactly the same as one that's already part of a viable pair is redundant, so we resolve this issue by simply skipping such candidates.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -390,16 +390,14 @@ impl IceAgent {
         self.candidate_pairs
             .iter()
             .filter(|cand| cand.state() == CheckState::Succeeded)
-            .find_map(|pair| {
+            .find(|pair| {
                 let o = &self.remote_candidates[pair.remote_idx()];
 
-                let same = c.addr() == o.addr()
+                c.addr() == o.addr()
                     && c.base() == o.base()
                     && c.proto() == o.proto()
                     && c.kind() == o.kind()
-                    && c.raddr() == o.raddr();
-
-                same.then(|| pair)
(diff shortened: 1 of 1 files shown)


src/ice/agent.rs
```diff
@@ -382,6 +382,27 @@ impl IceAgent {
...
+        &self,
+        c: &Candidate,
+    ) -> Option<&CandidatePair> {
+        self.candidate_pairs
+            .iter()
+            .filter(|cand| cand.state() == CheckState::Succeeded)
+            .find_map(|pair| {
+                let o = &self.remote_candidates[pair.remote_idx()];
+
+                let same = c.addr() == o.addr()
```
k0nserv, line 396: Maybe this business should be encapsulated in some method `Candidate::equivalent`

lolgesten (approved):

algesten (approved):

src/ice/agent.rs
algesten: Maybe. But let's revise that in another PR.

PR #579, opened 2024-10-24
Add support for loss in BWE

k0nserv:
This PR, which resurrects #147, adds support for BWE based on loss. Prior to this the BWE system relied solely on congestion manifesting as increasing delay, but it can also manifest as loss, with this PR we'll also react to loss.

Like the existing delay based BWE this is based on goog_cc, in particular it's based on the commit `14e2779a6c` from last year. 

## How it works

The fundamental problem with reacting to loss arising from congestion, is separating that loss from any inherent loss in the environment. For example, a peer might be using a poor WiFi with lots of packet loss on a downlink that can sustain, say 200Mbit/s. In such a situation we'd want to estimate their bandwidth at 200Mbit/s in the extreme, even if the inherent loss is 10%.

Simply reducing estimates based on observed packet loss does not work, in the above scenario we'd cap the peer at a very low estimate incorrectly.

This implementation uses [maximum likelihood estimation](https://en.wikipedia.org/wiki/Maximum_likelihood_estimation) to estimate the inherent loss $p$. This can then be used to distinguish inherent loss from loss arising from congestion.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -627,7 +627,7 @@ mod packet;
 
 #[path = "rtp/mod.rs"]
 mod rtp_;
-use rtp_::Bitrate;
+use rtp_::{Bitrate, DataSize};
 use rtp_::{Extension, ExtensionMap};
 
 /// Low level RTP access.
@@ -1830,13 +1830,19 @@ pub struct RtcConfig {
     exts: ExtensionMap,
     stats_interval: Option<Duration>,
     /// Whether to use Bandwidth Estimation to discover the egress bandwidth.
-    bwe_initial_bitrate: Option<Bitrate>,
+    bwe_config: Option<BweConfig>,
     reordering_size_audio: usize,
     reordering_size_video: usize,
     send_buffer_audio: usize,
     send_buffer_video: usize,
     rtp_mode: bool,
     enable_raw_packets: bool,
+}
+
+#[derive(Debug, Clone)]
+struct BweConfig {
+    initial_bitrate: Bitrate,
+    enable_loss_controller: bool,
 }
 
 impl RtcConfig {
@@ -2079,7 +2085,25 @@ impl RtcConfig {
     ///
     /// This includes setting the initial estimate to start with.
     pub fn enable_bwe(mut self, initial_estimate: Option<Bitrate>) -> Self {
-        self.bwe_initial_bitrate = initial_estimate;
+        match initial_estimate {
+            Some(b) => {
+                let conf = self.bwe_config.get_or_insert(BweConfig::new(b));
+                conf.initial_bitrate = b;
+            }
+            None => {
+                self.bwe_config = None;
+            }
+        }
+
+        self
+    }
+
+    /// Enable the experimental loss based BWE subsystem.
+    /// Defaults to disabled for now, will be enabled by default in the future.
+    pub fn enable_experimental_loss_based_bwe(mut self, enabled: bool) -> Self {
+        if let Some(c) = &mut self.bwe_config {
+            c.enable_loss_controller = enabled;
+        }
 
         self
     }
@@ -2094,7 +2118,7 @@ impl RtcConfig {
     /// assert_eq!(config.bwe_initial_bitrate(), None);
     /// ```
     pub fn bwe_initial_bitrate(&self) -> Option<Bitrate> {
-        self.bwe_initial_bitrate
+        self.bwe_config.as_ref().map(|c| c.initial_bitrate)
     }
 
     /// Sets the number of packets held back for reordering audio packets.
@@ -2265,6 +2289,15 @@ impl RtcConfig {
     }
 }
 
+impl BweConfig {
+    fn new(initial_bitrate: Bitrate) -> Self {
+        Self {
+            initial_bitrate,
+            enable_loss_controller: false,
+        }
+    }
+}
+
 impl Default for RtcConfig {
     fn default() -> Self {
         Self {
@@ -2275,7 +2308,7 @@ impl Default for RtcConfig {
             codec_config: CodecConfig::new_with_defaults(),
             exts: ExtensionMap::standard(),
             stats_interval: None,
-            bwe_initial_bitrate: None,
+            bwe_config: None,
             reordering_size_audio: 15,
             reordering_size_video: 30,
             send_buffer_audio: 50,
--- a/src/packet/bwe/delay_controller.rs
+++ b/src/packet/bwe/delay_controller.rs
@@ -1,0 +1,166 @@
+use std::collections::VecDeque;
+use std::time::{Duration, Instant};
+
+use crate::rtp_::Bitrate;
+use crate::util::already_happened;
+
+use super::arrival_group::ArrivalGroupAccumulator;
+use super::rate_control::RateControl;
+use super::trendline_estimator::TrendlineEstimator;
+use super::{AckedPacket, BandwidthUsage};
+
+const MAX_RTT_HISTORY_WINDOW: usize = 32;
+const UPDATE_INTERVAL: Duration = Duration::from_millis(25);
+/// The maximum time we keep updating our estimate without receiving a TWCC report.
+const MAX_TWCC_GAP: Duration = Duration::from_millis(500);
+
+/// Delay controller for googcc inspired BWE.
+///
+/// This controller attempts to estimate the available send bandwidth by looking at the variations
+/// in packet arrival times for groups of packets sent together. Broadly, if the delay variation is
+/// increasing this indicates overuse.
+pub struct DelayController {
+    arrival_group_accumulator: ArrivalGroupAccumulator,
+    trendline_estimator: TrendlineEstimator,
+    rate_control: RateControl,
+    /// Last estimate produced, unlike [`next_estimate`] this will always have a value after the
+    /// first estimate.
+    last_estimate: Option<Bitrate>,
+    /// History of the max RTT derived for each TWCC report.
+    max_rtt_history: VecDeque<Duration>,
+    /// Calculated mean of max_rtt_history.
+    mean_max_rtt: Option<Duration>,
+
+    /// The next time we should poll.
+    next_timeout: Instant,
+    /// The last time we ingested a TWCC report.
+    last_twcc_report: Instant,
+}
+
+impl DelayController {
+    pub fn new(initial_bitrate: Bitrate) -> Self {
+        Self {
+            arrival_group_accumulator: ArrivalGroupAccumulator::default(),
+            trendline_estimator: TrendlineEstimator::new(20),
+            rate_control: RateControl::new(initial_bitrate, Bitrate::kbps(40), Bitrate::gbps(10)),
+            last_estimate: None,
+            max_rtt_history: VecDeque::default(),
+            mean_max_rtt: None,
+            next_timeout: already_happened(),
+            last_twcc_report: already_happened(),
+        }
+    }
+
+    /// Record a packet from a TWCC report.
+    pub(crate) fn update(
+        &mut self,
+        acked: &[AckedPacket],
+        acked_bitrate: Option<Bitrate>,
+        now: Instant,
+    ) -> Option<Bitrate> {
+        let mut max_rtt = None;
+
+        for acked_packet in acked {
+            max_rtt = max_rtt.max(Some(acked_packet.rtt()));
+            if let Some(delay_variation) = self
+                .arrival_group_accumulator
+                .accumulate_packet(acked_packet)
+            {
+                crate::packet::bwe::macros::log_delay_variation!(delay_variation.delay_delta);
+
+                // Got a new delay variation, add it to the trendline
+                self.trendline_estimator
+                    .add_delay_observation(delay_variation, now);
+            }
+        }
+
+        if let Some(rtt) = max_rtt {
+            self.add_max_rtt(rtt);
+        }
+
+        let new_hypothesis = self.trendline_estimator.hypothesis();
+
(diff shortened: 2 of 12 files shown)


src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1341 @@
...
+    fn default() -> Self {
+        Self {
+            observation_window_size: 20, // minimum is 2
+            observation_duration_lower_bound: Duration::from_millis(0),
+            trendline_integration_enabled: false,
+            trendline_observations_window_size: 40,
+            temporal_weight_factor: 0.9,
+            instant_upper_bound_temporal_weight_factor: 0.9,
+            instant_upper_bound_loss_offset: 0.05,
+            instant_upper_bound_bandwidth_balance: 75_000.0, // 75 kbps
```
xnorpx, line 970: can't you use the DataRate type on all bps?

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1341 @@
...
+            use_acked_bitrate_only_when_overusing: false,
+            not_increase_if_inherent_loss_less_than_average_loss: true,
+            delayed_increase_window: Duration::from_millis(1000),
+            bandwidth_rampup_upper_bound_factor: 1000000.0,
+            probe_integration_enabled: false,
+            candidate_factor: [1.02, 1.0, 0.95],
+            append_acknowledged_rate_candidate: true,
+            append_delay_based_estimate_candidate: true,
+            bandwidth_backoff_lower_bound_factor: 1.0,
+            rampup_acceleration_maxout_time: 60.0, // 60s
```
xnorpx, line 989: Duration?

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1341 @@
...
+        // TODO: figure out whether to use usize or u64
+        self.num_packets += summary.num_packets;
+        self.num_lost_packets += summary.num_lost_packets;
+        self.size += summary.total_size;
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+struct ChannelParameters {
+    inherent_loss: f64,
```
xnorpx, line 922: inherent_loss, is it fraction or percent? either in name or newtype?

src/packet/bwe/loss_controller.rs
k0nserv: Indeed

src/packet/bwe/loss_controller.rs
k0nserv: Good point

src/packet/bwe/loss_controller.rs
k0nserv: It's percent expressed as a value in the range `[0, 1.0]`

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1341 @@
...
+
+    // output
+    loss_based_results: Bitrate,
+    _activated: bool,
+    _is_ready: bool,
+    //
+    // NB: Not ported from goog_cc(ALR, probing, link capcity)
+}
+
+struct Config {
```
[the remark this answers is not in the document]
k0nserv: Makes sense to me

src/lib.rs
```diff
@@ -2079,7 +2085,27 @@ impl RtcConfig {
...
+            None => {
+                self.bwe_config = None;
+            }
+        }
+
+        self
+    }
+
+    /// Enable the experimental loss based BWE subsystem.
+    pub fn enable_bwe_loss_experiment(mut self, enabled: bool) -> Self {
```
k0nserv, line 2102: Maybe this should just be called `enable_bwe_loss` and we can not in the comment that it's experimental. Then, when it becomes stable, we can default it to on

k0nserv:
Thanks for the feedback @xnorpx. I really should've created this as draft since I'm still working on it. I don't expect big changes, but the crate level configuration is particularly unfleshed out

src/lib.rs
algesten: Yeah, let's just make it the way it will be.

src/packet/bwe/delay_controller.rs
```diff
@@ -0,0 +1,174 @@
...
+use super::rate_control::RateControl;
+use super::trendline_estimator::TrendlineEstimator;
+use super::{AckedPacket, BandwidthUsage};
+
+const MAX_RTT_HISTORY_WINDOW: usize = 32;
+const UPDATE_INTERVAL: Duration = Duration::from_millis(25);
+/// The maximum time we keep updating our estimate without receiving a TWCC report.
+const MAX_TWCC_GAP: Duration = Duration::from_millis(500);
+
+/// Delay controller for googcc inspired BWe.
```
algesten, line 17: "BWe"?

src/packet/bwe/loss_controller.rs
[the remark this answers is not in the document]
algesten: Are they configurable in google's impl?

src/rtp/rtcp/twcc.rs
```diff
@@ -1010,7 +1010,7 @@ impl<'a> IntoIterator for &'a TwccSendRegister {
 }
 
 /// Record for a send entry in twcc.
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone)]
```
algesten, line 1013: I think this was deliberate. The record is only in one place.

src/packet/bwe/loss_controller.rs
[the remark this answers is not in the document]
k0nserv: > Are they configurable in google's impl?

Yep

pthatcher:
Is this based on libwebrtc's loss_based_bwe_v2 (https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc)?

src/rtp/rtcp/twcc.rs
```diff
@@ -1025,6 +1025,35 @@ pub struct TwccSendRecord {
 }
 
 impl TwccSendRecord {
+    /// Create a new record for which no report has been received.
+    pub fn new_unreported(seq: SeqNo, local_send_time: Instant, size: u16) -> Self {
```
k0nserv, line 1029: These are only needed for tests(we could guard them behind that cfg)

src/rtp/rtcp/twcc.rs
k0nserv: This enables an optimisation where we reuse allocations, see `Scratch` `src/packet/bwe/mod.rs`. They are only 64 bytes in size so it didn't feel like a problem to make them copy. 

Maybe `Scratch` is a bit premature though

src/packet/bwe/loss_controller.rs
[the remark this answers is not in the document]
xnorpx: They probably have some secret protobuf reader hidden there for their use :)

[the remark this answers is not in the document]
k0nserv: It's here https://chromium.googlesource.com/external/webrtc/+/14e2779a6ccdc67038ed2069a5732dd41617c6f0/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc#347

You can configure these overrides directly in Chrome

k0nserv:
> Is this based on libwebrtc's loss_based_bwe_v2 (https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc)?

Yeah, it is based on that one.

> In particular, is it based on "v2"? The description says it's based on a commit from a year ago. Is that too old?

It's not the latest version of "v2", but since it's tightly coupled with the delay estimator I thought it was good to base them both on the same era of code from libWebRTC. They can then be upgraded to the latest libWebRTC in tandem

src/rtp/rtcp/twcc.rs
k0nserv: Drop entirely in aa43264

k0nserv (commented):
@pthatcher, @davibe, and @xnorpx if you wanna have a look I'd appreciate it.

You can ignore the diff in `src/packet/bwe/delay_controller.rs` that was just moved from `src/packet/bwe/mod.rs` but other than that was not really touched.

src/packet/bwe/delay_controller.rs
```diff
@@ -0,0 +1,166 @@
...
+
+use crate::rtp_::Bitrate;
+use crate::util::already_happened;
+
+use super::arrival_group::ArrivalGroupAccumulator;
+use super::rate_control::RateControl;
+use super::trendline_estimator::TrendlineEstimator;
+use super::{AckedPacket, BandwidthUsage};
+
+const MAX_RTT_HISTORY_WINDOW: usize = 32;
```
xnorpx, line 12: what unit is this?

xnorpx: ok it's just slots

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1296 @@
+// ref: https://webrtc.googlesource.com/src/+/14e2779a6ccdc67038ed2069a5732dd41617c6f0/modules/congestion_controller/goog_cc/loss_based_bwe_v2.cc
```
xnorpx, line 1: same link is also below

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1296 @@
...
+
+    fn newtons_method_update(&self, channel_parameters: &mut ChannelParameters) {
+        if self.num_observations == 0 {
+            return;
+        }
+
+        for _ in 0..self.config.newton_iterations {
+            let derivatives = self.get_derivatives(channel_parameters);
+            channel_parameters.inherent_loss -=
+                self.config.newton_step_size * (derivatives.0 / derivatives.1);
```
xnorpx, line 440: it looks like your safe but still pointing out the / 0 risk here

src/packet/bwe/loss_controller.rs
```diff
@@ -0,0 +1,1296 @@
...
+
+        // maybe warn if sending rate or loss limited bandwidth are not finite
+
+        let mut loss_probability = inherent_loss;
+        if sending_rate.is_valid()
+            && loss_limited_bandwidth.is_valid()
+            && sending_rate > loss_limited_bandwidth
+        {
+            loss_probability += (1.0 - inherent_loss)
+                * (sending_rate.as_f64() - loss_limited_bandwidth.as_f64())
```
xnorpx, line 509: (bitrate - bitrate).as_f64 ?

xnorpx:
Integration tests?

xnorpx:
Maybe you already have something that can replay pcap and plot them is quite useful

k0nserv:
> Integration tests?

Hmm, yes. We don't have them for the delay based version either. I think to keep the size of this down I'd like to do that as a separate PR

> Maybe you already have something that can replay pcap and plot them is quite useful. But it's probably enough with debug logging and an parser for that.

Yeah, I use the stuff behind the feature `_internal_dont_use_log_stats` a lot when working on this. I pipe stdout to a file and plot the resulting metrics.

davibe (approved):

src/lib.rs
```diff
@@ -2079,7 +2085,25 @@ impl RtcConfig {
     ///
     /// This includes setting the initial estimate to start with.
     pub fn enable_bwe(mut self, initial_estimate: Option<Bitrate>) -> Self {
-        self.bwe_initial_bitrate = initial_estimate;
+        match initial_estimate {
```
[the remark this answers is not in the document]
k0nserv: Yeah I'm not entirely happy with this, but I think the current API is a poor since the order of calling is important(enable_bwe before `enable_loss_based_bwe`). Like you say either we take the entire BweConfig here or we do something like

```rust
pub fn bwe_config(&mut self) -> &mut BweConfig {
}
```

with `BweConfig` as 

```rust
struct BweConfig {
  enabled: bool,
  loss_based_config: LossBasedBweConfig,
}
```

Then usage becomes

```rust
let rtc = todo!();
let bwe_config = rtc.bwe_config();
bwe_config.loss_based_config.enabled = true;
bwbwe_config.loss_based_config.some_other_tunable_config = 10;
```

I think I prefer this a bit because it let's str0m stay in control of the defaults for these structs and doesn't require the user to construct them

src/lib.rs
```diff
@@ -1839,6 +1839,12 @@ pub struct RtcConfig {
     enable_raw_packets: bool,
 }
 
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
k0nserv: Could do, the I only added this because `RtcConfig` is `Clone`. If we add the entirety of the loss config in there we might not want `Copy` though

src/lib.rs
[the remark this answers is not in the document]
algesten: what about `enable_bwe(initial_estimate, loss_based_enabled)`?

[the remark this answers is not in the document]
algesten: Well. We're in semver 0.x and there will be breaking changes. Thinking about that, I think what you got is probably correct. Just rename `enable_loss_based_bwe` to `experimental_enable_loss_based_bwe` maybe?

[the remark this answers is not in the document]
k0nserv: Aight

src/packet/bwe/loss_controller.rs
k0nserv: If all the observations have no lost or received packets it could happen. I'll add it to the assert below

src/packet/bwe/loss_controller.rs
k0nserv: Removign

PR #580, opened 2024-10-25
Move openssl vendored featureflag to default

xnorpx:
Some security requirements might be that a user must use system provided OpenSSL.

Added a new feature called openssl-vendored which uses openssl feature and also add the vendored flag.

Easy way to test if open ssl is vendored or not is to.

cargo tree | grep openssl-src -> vendored

changed default to be openssl-vendored to ensure no functionality change

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,19 @@ exclude = ["/cargo_deny.sh", "/deny.toml", "/run-fuzz.sh"]
 # MSRV
 rust-version = "1.65"
 
+# Notes on OpenSSL
+# - Vendored (static link source build) ensures version consistency and easier 
+#   to build but increases size.
+# - Non-vendored (dynamic link) uses system libraries, reducing size but may
+#   cause compatibility issues. 
+# Choose based on security and footprint needs.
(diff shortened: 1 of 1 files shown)


xnorpx:
No still needs some work

xnorpx:
This is ready for review

algesten (commented):
As it happens, someone just now submitted a PR to ureq that shows a slightly more elegant way of doing this.

https://github.com/algesten/ureq/pull/866

We can have a `vendored` feature flag that applies to `openssl` if it is enabled. See the `?` syntax. I think that looks nicer.

xnorpx:
> As it happens, someone just now submitted a PR to ureq that shows a slightly more elegant way of doing this.
> 
> [algesten/ureq#866](https://github.com/algesten/ureq/pull/866)
> 
> We can have a `vendored` feature flag that applies to `openssl` if it is enabled. See the `?` syntax. I think that looks nicer.

Indeed, will try it out tonight to make sure it does what I think it does.

xnorpx:
@algesten ok ready for review, tested all combinations

algesten (approved):

PR #585, opened 2024-11-09
Add fuzz to CI

xnorpx:


xnorpx:
Ok here is example of run: https://github.com/algesten/str0m/actions/runs/11761427188/job/32763136796?pr=585

algesten (approved):

xnorpx:
> The purpose is to find regressions?

Regressions and also (assuming the seeds is random) it will hopefully fuzz with different inputs over time.

If it's noisy or flaky we can always remove/disable again.

PR #587, opened 2024-11-13
chore: add warning log when exceeding max number of pairs

thomaseizinger:
We've recently had a bug in Firezone where we generated a lot of unnecessary server-reflexive candidates. This was only caught by chance due to a different bug report from a user.

We do have automated reporting set up whenever parts of our code emit `WARN` and `ERROR` logs. Adding a WARN log here allows this automated reporting to kick in if that should happen again.

I think - in general - adding a WARN log here is justified. Exceeding the max number of candidates can technically lead to connectivity problems so informing the user about this via a WARN log seems good.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -42,6 +42,9 @@ pub struct IceAgent {
     // but the value MUST be configurable.
     max_candidate_pairs: Option<usize>,
 
+    /// Whether we have previously exceeded the `max_candidate_pairs` limit.
+    has_exceeded_max_candidate_pairs: bool,
+
     /// Credentials for this side. Set on init and ice-restart.
     local_credentials: IceCreds,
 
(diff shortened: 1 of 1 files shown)


algesten:
@thomaseizinger looks good.

Won't this give you a false positive for long lived road warriors? Like someone having a laptop over 5G on a train.

algesten (changes requested):
Let's land it, but with a change to ensure it only logs once.

thomaseizinger:
> Let's land it, but with a change to ensure it only logs once.

Once per what? This happens before the loop so only once per removal of overflow candidates.

Do you also want it to not log when we repeatedly add more candidates after that?

algesten:
That's it. Our instances can live for very long. So once per instance of the ice agent.

thomaseizinger:
> That's it. Our instances can live for very long. So once per instance of the ice agent.

Done. I added a flag that prevents this from logging repeatedly. The flag is reset on `ice_restart` and when the limit gets changed.

PR #589, opened 2024-11-25
Wincrypto Implementation for Str0m

efer-ms:
Implement wincrypto feature to use Windows Crypto APIs instead of OpenSSL on Windows platforms


efer-ms:
@algesten 

For being able to have multiple providers at once and reduce the number of touch points for providers, I was thinking of defining a couple of new traits, which will hopefully eliminate a bunch of the #[cfg]s. So you create the correct cert based on the configured provider, from the Cert you can then create a DtlsContext (or multiple with the same certificate), which will, and start the dtls machinery. Basically, the core of this is the same interface DtlsImpl, but added the SRTP cipher/hashes.

Anyways it would look something like:
```

trait DtlsCertificate: Clone {
	fn fingerprint(&self) -> Fingerprint;
	fn create_context(&self) -> Box<dyn DtlsContext>	
}

trait DtlsContext {
	// Returns the local certificate fingerprint.
	fn local_fingerprint(&self) -> Fingerprint;

	// DTLS session management
	fn set_active(&mut self, active: bool) -> ();
	fn is_active() -> Option<bool>;
	fn is_connected() -> bool;
	fn handle_handshake(&mut self, out_events: &mut VecDeque<DtlsEvent>) -> Result<bool, CryptoError>;
	fn handle_receive(&mut self, datagram: &[u8], out_events: &mut VecDeque<DtlsEvent>) -> Result<(), CryptoError>;
	fn poll_datagram(&mut self) -> Option<DatagramSend>;
	fn poll_timeout(&mut self, now: Instant) -> Option<Instant>;
	fn handle_input(&mut self, data: &[u8]) -> Result<(), CryptoError>;

	// RTP/SRTP ciphers and hashes
	fn create_aes_128_cm_sha1_80_cipher(key, encrypt) -> Box<dyn aes_128_cm_sha1_80::CipherCtx>;
	fn create_aead_aes_128_gcm_cipher(key, encrypt) -> Box<dyn aes_128_cm_sha1_80::CipherCtx>;
	fn srtp_aes_128_ecb_round(key, input, output) -> ();
	fn sha1_hmac(key, payloads) -> [u8;20];
}
```

Then in `new_from_config` we create the context appropriate for the configured provider.

RtcConfig would have `crypto_provider: Option<CryptoProvider>`
```
enum CryptoProvider {
  #[cfg(all(feature="openssl", feature="sha1"))]
  OpenSslWithSha1Crate,
  #[cfg(feature="openssl")]
  OpenSsl,
  #[cfg(feature="wincrypto")]
  WinCrypto,
}
```

If not provided, the default for CryptoProvider is OpenSslWithSha1Crate if the `sha1` feature is enabled, otherwise it's OpenSsl , which matches current behavior.

The Sha1 swapping I moved into the openssl provider for simplicity, if wincrypto needs it cause it's sha1 is slow too (haven't measured, it can be added in a similar manner).

...

This is a pretty big refactor, I think we should keep it separate from this wincrypto impl. If you agree this is a good direction, I can work on it separately, I guess the question is whether to get this wincrypto in now, then change the configuration stuff in a follow-up... or hole off on getting wincrypto in until we have this refactor in.

We've already started using the wincrypto provider so we can get some mileage on it.

algesten:
> Anyways it would look something like...

This plan looks great!

> This is a pretty big refactor, I think we should keep it separate from this wincrypto impl.

My spontaneous thought is that it doesn't look that big. I guess we can merge, although having those conditional compilations here and there doesn't spark joy.

algesten (approved):

algesten:
Let's fix the conflicts and then merge.

efer-ms:
> > Anyways it would look something like...
> 
> This plan looks great!
> 
> > This is a pretty big refactor, I think we should keep it separate from this wincrypto impl.
> 
> My spontaneous thought is that it doesn't look that big. I guess we can merge, although having those conditional compilations here and there doesn't spark joy.

Ok.. I'll get started on the refactor right away. I don't want to let the thoughts escape, and take advantage of the short work week in the US.

algesten:
> Ok.. I'll get started on the refactor right away. I don't want to let the thoughts escape, and take advantage of the short work week in the US.

It's ok either way. @xnorpx convinced me :)

efer-ms:
> @efer-ms looks like you need to do a manual rebase and force push

I'm confused.. github shows no conflicts. What am I missing?

xnorpx:
> > @efer-ms looks like you need to do a manual rebase and force push
> 
> I'm confused.. github shows no conflicts. What am I missing?

Never mind need to squash and merge it. Let me do that.

src/crypto/mod.rs
```diff
@@ -19,6 +22,11 @@ mod srtp;
 pub use srtp::{aead_aes_128_gcm, aes_128_cm_sha1_80, new_aead_aes_128_gcm};
 pub use srtp::{new_aes_128_cm_sha1_80, srtp_aes_128_ecb_round, SrtpProfile};
 
+#[cfg(all(feature = "openssl", feature = "wincrypto"))]
+compile_error!("features `openssl` and `wincrypto` are mutually exclusive");
+#[cfg(not(any(feature = "openssl", feature = "wincrypto")))]
+compile_error!("either `openssl` or `wincrypto` must be enabled");
```
thomaseizinger, line 28: I am not quite sure why it works if we depend on `str0m` with `default-features = false, features = ["sha1"]` but I expected this line to cause problems.

In the str0m repo, `cargo check --no-default-features --features sha1` doesn't work.

Issue #586, opened 2024-11-11
h264.rs:263:30: index out of bounds: the len is 42 but the index is 42


xnorpx:
I have not checked the code yet, but save it here to not lose it

#101072	REDUCE cov: 940 ft: 2951 corp: 279/91Kb lim: 690 exec/s: 12634 rss: 318Mb L: 166/663 MS: 5 InsertRepeatedBytes-EraseBytes-CMP-InsertByte-InsertByte- DE: "\377\377"-
thread '<unnamed>' panicked at /home/runner/work/str0m/str0m/src/packet/h264.rs:263:30:
index out of bounds: the len is 42 but the index is 42

https://github.com/algesten/str0m/actions/runs/11769744229/job/32781189675#step:8:916

k0nserv:
> We check for +1, but STAPA_NALU_LENGTH_SIZE is 2. Maybe that's it?

At offset 1 we expect two bytes for the length if the packet is 2 bytes long `1 + 1 < 2 == false` and we don't loop. The minimum is 3 bytes.

Failing test here https://github.com/algesten/str0m/compare/main...feature/stapa-nalu-parsing. I think this condition is wrong https://github.com/algesten/str0m/blob/36ae0ec699ce6e871470f779ac32609e1d22d533/src/packet/h264.rs#L256-L261, should be `>= packet.len()`, but that makes another test fail

algesten:
Is the test wrong_

k0nserv:
It's possible, I'll dig into it

xnorpx:
I will close this one after confirm it's gone next fuzz run.

xnorpx:
<img width="539" alt="image" src="https://github.com/user-attachments/assets/242bbab4-7604-4888-a97e-842bc7fa8d27">

PR #590, opened 2024-11-26
Fix oob fuzz test failure

xnorpx:
One could fix and rewrite this in many ways, but to keep small changes this is my proposal.


xnorpx:
Fix for #586

davibe (approved):
to minimize changes this could be `let Some(b0) = packet.get(curr_offset) else { continue; }`

PR #591, opened 2024-11-26
Add crypto faq

xnorpx:
Suggestion of addition to FAQ section after wincrypto is merged.


algesten:
Added something smaller in #594

PR #593, opened 2024-12-03
Refactor crypto features

lolgesten:
This is a refactor to make less use of the feature flags **openssl** and **wincrypto** for conditional compilation. It plugs in "dummy" impls that panics if the corresponding feature is not turned on. The goal is:

* Able to compile with `--all-features` (no mutually exclusive feature flags)
* Always default to a specific crypto (currently openssl), even if that fails runtime. This is to ensure consumers make the correct config instead of relying on feature flags to get a desired behavior
* Minimize amount of code behind conditional compilation to aid refactoring


lolgesten:
Darn. I need to redo this as my private github user. :S

PR #596, opened 2024-12-05
MSRV 1.65 -> 1.71 and update deps

algesten:


thomaseizinger (approved):
We are on latest stable so don't mind the MSRV bump :)

xnorpx:
> We are on latest stable so don't mind the MSRV bump :)

Same here

xnorpx (approved):

PR #597, opened 2024-12-06
refactor: re-export crypto items under same name to avoid `cfg`s

thomaseizinger:
This is how we handle platform-specific code in Firezone and it works quite well for us. I went ahead and decided that `openssl` can't at all be used on Windows anymore because it makes the `_impl` module nicely platform specific and avoids the issue of `--all-features` not working.

Let me know what you think.

### DIFF
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -49,29 +49,10 @@ pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
     hmac.finalize().into_bytes().into()
 }
 
-/// If openssl is enabled and sha1 is not, it uses `openssl` crate.
-#[cfg(all(feature = "openssl", not(feature = "sha1")))]
-pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
-    use openssl::hash::MessageDigest;
-    use openssl::pkey::PKey;
-    use openssl::sign::Signer;
-
-    let key = PKey::hmac(key).expect("valid hmac key");
-    let mut signer = Signer::new(MessageDigest::sha1(), &key).expect("valid signer");
-
-    for payload in payloads {
-        signer.update(payload).expect("signer update");
-    }
-
-    let mut hash = [0u8; 20];
-    signer.sign(&mut hash).expect("sign to array");
-    hash
-}
-
-/// If wincrypto is enabled and sha1 is not, it uses `wincrypto` crate.
+/// if sha1 is not enabled, delegate to the crypto impl.
 #[cfg(all(feature = "wincrypto", not(feature = "sha1")))]
 pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
-    wincrypto::sha1_hmac(key, payloads)
+    _impl::sha1_hmac(key, payloads)
 }
 
 /// Errors that can arise in DTLS.
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -50,7 +50,7 @@ pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
 }
 
 /// if sha1 is not enabled, delegate to the crypto impl.
-#[cfg(all(feature = "wincrypto", not(feature = "sha1")))]
+#[cfg(not(feature = "sha1"))]
 pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
     _impl::sha1_hmac(key, payloads)
 }
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -21,6 +21,7 @@ mod keying;
 pub use keying::KeyingMaterial;
 
 mod srtp;
+pub use _impl::{AeadAes128Gcm, Aes128CmSha1_80};
 pub use srtp::{aead_aes_128_gcm, aes_128_cm_sha1_80, new_aead_aes_128_gcm};
 pub use srtp::{new_aes_128_cm_sha1_80, srtp_aes_128_ecb_round, SrtpProfile};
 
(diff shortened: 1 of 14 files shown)


thomaseizinger:
@algesten This now passes except for OpenSSL on Windows. Not sure if you want to support that. It wouldn't be too difficult to change the `cfg`s to allow it although conflicting features isn't very clean. For example, it breaks `crater` runs from the Rust team because they run with `--all-features` (AFAIK).

Feel free to take over from here.

algesten:
Hey @thomaseizinger thanks for looking at this!

> This now passes except for OpenSSL on Windows. Not sure if you want to support that. It wouldn't be too difficult to change the cfgs to allow it although conflicting features isn't very clean. For example, it breaks crater runs from the Rust team because they run with --all-features (AFAIK).

Yeah, so OpenSSL on Windows is a valid configuration that we need to support (or at least I think it is. Maybe Windows users never want to use OpenSSL?). Ideally I should be able to run `wincrypto` also on `macOS` and it's just inert and doesn't work.

From my PR, the goals are:

> * Able to compile with --all-features (no mutually exclusive feature flags)
> * Always default to a specific crypto (currently openssl), even if that fails runtime. This is to ensure consumers make the correct config instead of relying on feature flags to get a desired behavior
> * Minimize amount of code behind conditional compilation to aid refactoring

Having said that, there's a clean elegance to what you got here. I wonder if I can iterate on it to fulfill my design goals and have it like this? 🤔

thomaseizinger:
> > * Able to compile with --all-features (no mutually exclusive feature flags)
> > * Always default to a specific crypto (currently openssl), even if that fails runtime. This is to ensure consumers make the correct config instead of relying on feature flags to get a desired behavior
> > * Minimize amount of code behind conditional compilation to aid refactoring
> 
> Having said that, there's a clean elegance to what you got here. I wonder if I can iterate on it to fulfill my design goals and have it like this? 🤔

You can always configure the cfg's in such a way that one takes presedence over the other if both are active. Gets a bit convoluted with the cfgs but definitely doable.

I think that would satisfy all your goals.

thomaseizinger:
Adding a dedicated RUSTFLAGS cfg for the active one would also be a solution.

thomaseizinger:
- Use features like `wincrypto` and `openssl` to make dependencies optional
- Have a `#[cfg(str0m_crypto_backend)]`
- The active impl block is decided as `all()` of the backend and the corresponding feature
- By default, backend is `openssl`

xnorpx:
From our end only wincrypto is allowed on Windows (hence the original pr) so we won't use OpenSSL.

Not use OpenSSL will make str0m smaller binary on Windows, less dependency and simpler to build on Windows platforms.

But that doesn't mean there are no users of str0m out there that "want" to use OpenSSL on Windows.

Maybe only use wincrypto only on Windows for now until someone complains about not having OpenSSL. (But most likely the main complaint why there is no "Rust Only crypto DTLS implementation".

src/crypto/mod.rs
```diff
@@ -44,43 +49,18 @@ pub fn sha1_hmac(key: &[u8], payloads: &[&[u8]]) -> [u8; 20] {
...
-    }
-
-    let mut hash = [0u8; 20];
-    signer.sign(&mut hash).expect("sign to array");
-    hash
-}
-
-/// If wincrypto is enabled and sha1 is not, it uses `wincrypto` crate.
+/// if sha1 is not enabled, delegate to the crypto impl.
 #[cfg(all(feature = "wincrypto", not(feature = "sha1")))]
```
thomaseizinger, line 53:
```suggestion
#[cfg(not(feature = "sha1"))]
```

src/crypto/srtp.rs
```diff
@@ -37,90 +37,31 @@ impl SrtpProfile {
...
-    #[cfg(feature = "wincrypto")]
-    {
-        let ctx = super::wincrypto::WinCryptoSrtpCryptoImpl::new_aes_128_cm_sha1_80(key, encrypt);
-        Box::new(ctx)
-    }
-    #[cfg(not(any(feature = "openssl", feature = "wincrypto")))]
-    {
-        panic!("No SRTP implementation. Enable openssl feature");
-    }
+pub fn new_aes_128_cm_sha1_80(key: AesKey, encrypt: bool) -> super::_impl::Aes128CmSha1_80 {
```
thomaseizinger, line 40: This indirection is pretty redundant now.

thomaseizinger:
> Maybe only use wincrypto only on Windows for now until someone complains about not having OpenSSL. (But most likely the main complaint will be "why there is no Rust only crypto DTLS implementation".

I am in support of this. It isn't too hard to fix in case someone complains and it will make maintenance easier.

algesten:
> > Maybe only use wincrypto only on Windows for now until someone complains about not having OpenSSL. (But most likely the main complaint will be "why there is no Rust only crypto DTLS implementation".
> 
> I am in support of this. It isn't too hard to fix in case someone complains and it will make maintenance easier.

I should probably elaborate on what my worries are.

> * Able to compile with --all-features (no mutually exclusive feature flags)
> * Always default to a specific crypto (currently openssl), even if that fails runtime. This is to ensure consumers make the correct config instead of relying on feature flags to get a desired behavior
> * Minimize amount of code behind conditional compilation to aid refactoring

Number 1 and 3 here are about maintainability. With ureq, I accepted that code had large swaths of code gated behind mutually exclusive feature flags. It became a maintenance nightmare slowing down the rate of change. The wincrypto backend is in some respects worse because I don't have any Windows machine, and I don't particularly fancy needing emulation environments whenever I touch str0m.

Having said that, I am committed to have this in-tree and support it as much as I can. This is why I favor #594 and will try to take that to completion today.

thomaseizinger:
I have no strong position on which way we go here as long as the `IceAgent` can be used standalone.

From experience, have naturally disjoint modules based on target platforms works well. Of course, if we want to offer openssl on Windows too, then this doesn't work that well.

Looking forward to seeing how this pans out :)

algesten:
Closing in favor of #594 now merged to main.

Issue #598, opened 2024-12-20
Panic in `CandidatePair::next_binding_attempt`


thomaseizinger:
We received a crash report of our application that was triggered by what I believe is this `unwrap`: https://github.com/algesten/str0m/blob/8e8f82a08b2a6ca98d298b9d13e1520129c1f9db/src/ice/pair.rs#L330

Here is the stacktrace:

```
OS Version: windows 10.0.26120 (None)
Report Version: 104


Application Specific Information:
called `Option::unwrap()` on a `None` value

Thread 0 Crashed:
0   firezone-client-ipc.exe         0x7ff7e76f8c80      backtrace::capture::Backtrace::new
1   firezone-client-ipc.exe         0x7ff7e76f8b87      backtrace::capture::Backtrace::new
2   firezone-client-ipc.exe         0x7ff7e7d22477      sentry_panic::PanicIntegration::event_from_panic_info
3   firezone-client-ipc.exe         0x7ff7e7d21f1b      sentry_panic::panic_handler
4   firezone-client-ipc.exe         0x7ff7e7d22213      sentry_panic::PanicIntegration::setup
5   firezone-client-ipc.exe         0x7ff7e7da421a      [inlined] alloc::boxed::impl$50::call (boxed.rs:2245)
6   firezone-client-ipc.exe         0x7ff7e7da421a      std::panicking::rust_panic_with_hook (panicking.rs:805)
7   firezone-client-ipc.exe         0x7ff7e7da407e      std::panicking::begin_panic_handler::closure$0 (panicking.rs:664)
8   firezone-client-ipc.exe         0x7ff7e7da1efe      std::sys::backtrace::__rust_end_short_backtrace<T> (backtrace.rs:170)
9   firezone-client-ipc.exe         0x7ff7e7da3ca5      std::panicking::begin_panic_handler (panicking.rs:662)
10  firezone-client-ipc.exe         0x7ff7e7ea7e53      core::panicking::panic_fmt (panicking.rs:74)
11  firezone-client-ipc.exe         0x7ff7e7ea7efc      core::panicking::panic (panicking.rs:148)
12  firezone-client-ipc.exe         0x7ff7e7ea7dbd      core::option::unwrap_failed (option.rs:2015)
13  firezone-client-ipc.exe         0x7ff7e7ddb94c      str0m::ice_::pair::CandidatePair::next_binding_attempt
[...]

PR #602, opened 2024-12-25
chore: add test for #598

thomaseizinger:
In #600, we fixed a bug that triggered a `panic` in `str0m` when certain arithmetic operations on an `Instant` failed. Only after merging did I think of an idea on how this can actually be unit-tested.

`Instant`s are represented in a platform-specific way, with some arbitrary point-in-time as a reference point. To simulate this, we attempt to subtract `Duration`s until we find this earliest point where a previous one can now longer be represented. We then attempt to use this point in time to run a `host - host` ICE handshake, ensuring that `str0m` handles the case gracefully where the provided `Instant` can no longer handle earlier points in time.

### DIFF
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -261,6 +261,34 @@ mod test {
         );
     }
 
+    // str0m performs calculations on `now` internally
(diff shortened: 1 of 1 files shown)


algesten (approved):
Nice! :D

PR #601, opened 2024-12-24
Dedupe acked packets from TwccSendRegister::apply_report()

alexlapa:
Fixes #599 

> The missing invariant here, at least as far as BWE is concerned, is that we don't want to process the same packet(as indicated by transport-cc sequence number) more than one time. We can achieve this either by apply_report to not return a range or, preferably, by changing send_records to only return "new" reports. Alternatively, and perhaps preferably, we add a new layer between TwccSendRegister and the BWE subsystem inspired by libWebRTC's TransportFeedbackAdapter construct as you suggest.

So here is a deduplicating adapter between `TwccSendRegister` and the insides of `SendSideBandwithEstimator`. It should do the job but looks kinda awkward to me. I would probably prefer encapsulating this inside of `TwccSendRegister` by simply removing `TwccSendRecord`s from the storage like libwebrtc [does](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/rtp/transport_feedback_adapter.cc;l=393-400;drc=776866774fd739996b60a6b683f7aa9fa3addd65). But anyway, if that works for you then it works for me too.

> A good first step might be a failing test based on such a sequence of TWCC feedbacks.

I've also added a test from libwebrtc captured data. It fails right now to demonstrate whats wrong, I'l fix it before merge.

PR #605, opened 2025-01-06
Dedupe TwccSendRecord from apply_report()

algesten:


algesten:
@alexlapa @k0nserv here's another take on the BWE deduping.

alexlapa:
So I've logged `acked_packets` right [here](https://github.com/algesten/str0m/blob/bcb9888b0381b5688e41d7a204de987389f446df/src/packet/bwe/mod.rs#L85) when under reordering and it look like this:

```
[0, 1, 3]
[3, 2, 4, 5]
[9, 11]
[9, 11, 10, 7, 8, 13, 6, 14]
[13, 14, 12]
[16, 17, 18, 20, 21]
[16, 17, 18, 20, 21, 22, 15]
[20, 21, 22, 19, 23, 28, 29, 24, 31]
[28, 29, 31, 33, 25, 32, 37, 39, 40, 27, 30]
```

So it does not fix the issue. A possible problem might how you set `TwccRecvReport.apply_report_counter`, right now it is being updated every time packet is mentioned in TWCC, which might happen multiple times. And it is supposed to be set only once, the first time TWCC says it was received by the remote, and the value should be carried over during all subsequent mentions. So something like this:

```rust
let apply_report_counter = if remote_recv_time.is_some() {
  if let Some(rr) = r.recv_report {
      // Was mentioned before so either set current counter or carry over previous value
      rr.apply_report_counter.or(Some(apply_report_counter))
  } else {
      // It is first time packet is mentioned and it is received
      Some(apply_report_counter)
  }
} else {
    // Packet is not received
    None
};

let recv_report = TwccRecvReport {
    local_recv_time: now,
    remote_recv_time,
    apply_report_counter,
};
```

But anyway there are some other problems I'm currently trying to figure out.

algesten:
Ah, of course. Let's fix that!

algesten:
@alexlapa I pushed a commit. I think it might fix it?

alexlapa:
@algesten , it needs #606 first.

algesten:
@alexlapa so is this working now if we combine #605 and #606?

Can we drop #601?

alexlapa:
@algesten ,

> so is this working now if we combine https://github.com/algesten/str0m/pull/605 and https://github.com/algesten/str0m/pull/606?

ce06ba3 is problematic since it makes it so if packet is included as not received in the first TWCC, but second TWCC reports it as received, then `TwccRecvReport.remote_recv_time` wont be set. So we will treat delayed packet as lost.

You can use this test from #601 to reproduce what I'm talking about:

```rust
    #[test]
    fn no_duplicates_when_reordered() {
        let now = Instant::now();
        let mut twcc_gen = TwccRecvRegister::new(1000);
        let mut twcc_handler = TwccSendRegister::new(1000);

        twcc_handler.register_seq(1.into(), now + Duration::from_millis(1), 0);
        twcc_handler.register_seq(2.into(), now + Duration::from_millis(2), 0);
        twcc_handler.register_seq(3.into(), now + Duration::from_millis(3), 0);
        twcc_handler.register_seq(4.into(), now + Duration::from_millis(4), 0);

        {
            let acked_packets = twcc_handler
                .apply_report(
                    {
                        // 3rd packet is delayed
                        twcc_gen.update_seq(1.into(), now + Duration::from_millis(5));
                        twcc_gen.update_seq(2.into(), now + Duration::from_millis(6));
                        twcc_gen.update_seq(4.into(), now + Duration::from_millis(7));
                        twcc_gen.build_report(10_000).unwrap()
                    },
                    now + Duration::from_millis(8),
                )
                .unwrap()
                .filter_map(|sr| sr.remote_recv_time().map(|_| sr.seq.as_u16()))
                .collect::<Vec<_>>();

            assert_eq!(acked_packets, [1, 2, 4]);
        }

        twcc_handler.register_seq(5.into(), now + Duration::from_millis(9), 0);
        twcc_handler.register_seq(6.into(), now + Duration::from_millis(10), 0);
        twcc_handler.register_seq(7.into(), now + Duration::from_millis(11), 0);

        {
            let acked_packets = twcc_handler
                .apply_report(
                    {
                        // So the receipt order is 1, 2, 4, 3, 7
                        twcc_gen.update_seq(3.into(), now + Duration::from_millis(12));
                        twcc_gen.update_seq(7.into(), now + Duration::from_millis(13));
                        twcc_gen.build_report(10_000).unwrap()
                    },
                    now + Duration::from_millis(14),
                )
                .unwrap()
                .filter_map(|sr| sr.remote_recv_time().map(|_| sr.seq.as_u16()))
                .collect::<Vec<_>>();

            // 4 was already returned previously and [5, 6] are not received atm
            assert_eq!(acked_packets, [3, 7]);
        }
    }
```

> Can we drop https://github.com/algesten/str0m/pull/601?

I'l delete it once this PR is merged, or i can merge this PR into #601 and fix everything there, whatever you like.

algesten:
I still get an error with this test. 3 is registered in the first apply_report()

alexlapa:
@algesten ,

> I still get an error with this test.

Well yeah, test is supposed to reproduce the problem I've mentioned [here](https://github.com/algesten/str0m/pull/605#issuecomment-2575764989).

> https://github.com/algesten/str0m/commit/ce06ba3eb8992865fd8205a7685248e820a859b7 is problematic since it makes it so if packet is included as not received in the first TWCC, but second TWCC reports it as received, then TwccRecvReport.remote_recv_time wont be set. So we will treat delayed packet as lost.

> 3 is registered in the first apply_report()

`3` is included in the first TWCC as not received so `TwccRecvReport { remote_recv_time: None }` is created, this is correct. And in the second TWCC `3` is included as received (acked). So you must update `TwccSendRecord.recv_report.remote_recv_time` from `None` to `Some`, cause this packet was not lost, it was delayed.

I'll make code suggestions with more explanations.

lolgesten:
@alexlapa feel free to go the other way and merge into #601 too. I just assumed this was very close.

src/rtp/rtcp/twcc.rs
```diff
@@ -1125,13 +1149,21 @@ impl TwccSendRegister {
...
+            apply_report_counter: u64,
         ) -> bool {
             if r.seq != seq {
                 return false;
             }
+
+            // Reports can double count. We are only interested in the first ever report.
+            if r.recv_report.is_some() {
+                return true;
+            }
```
alexlapa, line 1161: @algesten ,

So to explain why this `if` is problematic as I've mentioned [here](https://github.com/algesten/str0m/pull/605#issuecomment-2575764989) and [here](https://github.com/algesten/str0m/pull/605#issuecomment-2577183327).

```rust
let apply_report_counter = if let Some(rr) = r.recv_report {
    if rr.remote_recv_time.is_some() {
        // This packet was acked by the previous TWCC, so we carry over
        // apply_report_counter, so it won't be included in the
        // apply_report() result. We only dedup acked packets.
        rr.apply_report_counter
    } else {
        // This packet is nacked, and it was nacked before. It may
        // be lost or delayed. If it is delayed, then some
        // other TWCC will ack it later. It is included in apply_report()
        // result, but it won't be converted to `AckedPacket` in
        // BWE module.
        apply_report_counter
    }
} else {
    if instant.is_some() {
        // This is the first time packet is included, it's acked, so 
        // we set current apply_report_counter value and it won't be 
        // changed by any further TWCCs. This packet will be returned
        // in the apply_report() result only in this call, it will
        // be converted into `AckedPacket` in BWE module.
        apply_report_counter
    } else {
        // This is the first time packet is included, it's nacked
        // it may be lost or delayed. If it is delayed, then some
        // other TWCC will ack it later. It is included in apply_report()
        // result, but it won't be converted to `AckedPacket` in
        // BWE module.
        apply_report_counter
    }
};

// Carry over previous remote recv time if it was set before.
let remote_recv_time = r.recv_report.and_then(|rr|r.remote_recv_time()).or(instant);

let recv_report = TwccRecvReport {
    local_recv_time: now,
    remote_recv_time,
    apply_report_counter,
};
r.recv_report = Some(recv_report);
```

algesten:
Closing in favor of #601

algesten:
@alexlapa thank you so much for investing the time in investigating, explaining, fixing etc. Great!!!!

PR #607, opened 2025-01-09
Add EC-DSA, custom certificate common names for DTLS

efer-ms:
- New Features -

Update Str0m to support EC-DSA certification for DTLS. See: https://developer.chrome.com/blog/webrtc-ecdsa/

This updates both the OpenSSL and Wincrypto providers to allow both RSA or EC-DSA certificates. The default remains RSA, although perhaps it should change to EC-DSA at some point.

Took the opportunity to also allow the CN of the certificate to be configured (as this was an existing TODO).

----
- Other changes -

OpenSSL is configured to indicate SHA1 is not acceptable for the DTLS certificate signing, but then proceeded to use SHA1 and for some reason accepts it. This changes it to use SHA256 which matches WinCrypto, since the Windows APIs refused to offer a SHA1 signed cert if the DTLS handshake indicated it wasn't supported.

WinCrypto was updated to use the Owned<Free> struct that windows-rs offers for cleaning up the SrtpKey, rather than having a custom Drop impl.

WinCrypto had an issue when compiled in --release, where DTLS handshake would fail as a server. This was related to buffers being marked as immutable, but then cast to mutable in unsafe code. The compiler (I believe) was using the immutability to avoid reading the values multiple times, so although things were mutated in the unsafe code, the code did not recognize the changes. This just updated those buffers to be marked `mut`, it also eliminated some of the double-casts in unsafe code.


src/crypto/dtls.rs
```diff
@@ -59,12 +89,12 @@ impl DtlsCert {
     ///
     /// * **openssl** (defaults to on) for crypto backed by OpenSSL.
     /// * **wincrypto** for crypto backed by windows crypto.
-    pub fn new(p: CryptoProvider) -> Self {
+    pub fn new(p: CryptoProvider, dtls_cert_options: DtlsCertOptions) -> Self {
```
algesten, line 92: Let's rename `dtls_cert_options` to `opts` and follow that through down. Let the type be the context.

src/crypto/ossl/cert.rs
```diff
@@ -25,16 +26,26 @@ pub struct OsslDtlsCert {
...
         let f4 = BigNum::from_u32(RSA_F4).unwrap();
-        let key = Rsa::generate_with_e(2048, &f4)?;
-        let pkey = PKey::from_rsa(key)?;
+        let pkey = match options.pkey_type {
+            DtlsPKeyType::Rsa => {
+                let key = Rsa::generate_with_e(2048, &f4)?;
+                PKey::from_rsa(key)?
+            }
+            DtlsPKeyType::EcDsa => {
+                let nid = Nid::X9_62_PRIME256V1; // NIST P-256 curve
```
algesten, line 43: Is it understood that this is the curve to always use for ECDSA in this context?

efer-ms:
The public API change is mostly set_dtls_config in RtcConfig to allow changing the default cert generation options. The default behavior matches the current Str0m.

There is a breaking API change for Pregenerated DTLS certs, since to create a cert, you need to supply the certification options. I'm not sure how widespread the use of that API is.

src/lib.rs
```diff
@@ -1939,6 +1957,17 @@ impl RtcConfig {
         self.crypto_provider
     }
 
+    /// Returns the configured DTLS certificate configuration.
+    pub fn dtls_cert_config(&self) -> &DtlsCertConfig {
+        &self.dtls_cert_config
+    }
+
+    /// Set the DTLS certificate configuration for certificate generation.
```
algesten, line 1965: Let's flesh this out a bit. Show an example.

src/lib.rs
```diff
@@ -1939,6 +1957,17 @@ impl RtcConfig {
         self.crypto_provider
     }
 
+    /// Returns the configured DTLS certificate configuration.
```
algesten, line 1960: Let's flesh this out a bit. Say what the default is.

algesten (approved):
Looks good!

A couple of minor comments.

Thanks for pushing ECDSA!

src/crypto/ossl/cert.rs
efer-ms: My understanding isn't great, but that seems to be the curve I've seen for 256bit EcDsa.

We could rename the enum values to EcDsaP256. and RSA to Rsa2048? To be ready when someone was EcDsaP384 or something else.

algesten: Is there any precedence in Pion or similar?

efer-ms: Looks like Pion doesn't allow you to select the key type.. it just uses EC-DSA w/ the NIST P-256 curve. https://github.com/pion/dtls/blob/1e4ae60e28516c475e999b389bcbe96a168c4470/pkg/crypto/selfsign/selfsign.go#L26

efer-ms: you can use other curves but this is the common curve I've seen used. This is one of the FIPS standard curves, so that likely drives the adoption over others of similar strength.

efer-ms: So I do think it would probably be good to rename the values in the enum to Rsa2048 and EcDsaP256 .. so you explicitly know the key size and curve being used.

And then it's simply to add other curves and strength in the future.

src/lib.rs
efer-ms: Should I remove the set_dtls_cert()... it can be done using this same function and passing in DtlsCertConfig::PregeneratedCert instead of DtlsCertConfig::Options ? I kept it for API compat....same goes for get_dtls_cert().. you could just do get_dtls_cert_config() and that returns either Options or a PreeneratedCert

src/crypto/dtls.rs
efer-ms: Done

src/lib.rs
efer-ms: Done

algesten (approved):

algesten:
Thanks!

PR #608, opened 2025-01-10
Align BWE ArrivalGroup calculation with libwebrtc implementation

alexlapa:
So there are some differences in str0m's `ArrivalGroupAccumulator` and libwebrtc's `InterArrivalDelta` implementation that affect trendline estimator bandwidth usage hypothesis thus the final BWE result. This PR alligns str0m's implementation so it would work more like libwebrtc's.

Changes are pretty straightforward, only `ArrivalGroup::belongs_to_group` is modified and it corresponds to [this](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/inter_arrival_delta.cc;l=109-135;drc=c4a205c7fa7d25f2e1c8e66156d353a6a3df0e81) two functions in libwebrtc.

And to show that modified `ArrivalGroup` works like libwebrtc's implementation this PR [includes some tests](https://github.com/alexlapa/str0m/blob/f0f1ea355f2de106178c9afe9c3b7639b3a220ca/src/packet/bwe/arrival_group.rs#L5932) based on the data captured in libwebrtc that run modified `ArrivalGroup` next to the latest libwebrtc implementation (it is [slightly modified](https://www.diffchecker.com/Lsi7QqU3/) but it won't affect results). I will obviously remove this before merge.

#### Also I've got few things to discuss here:
1. Notice that there are three data sets that tests are ran against, and the third one is [disabled](https://github.com/alexlapa/str0m/blob/f0f1ea355f2de106178c9afe9c3b7639b3a220ca/src/packet/bwe/arrival_group.rs#L5052-L5053). This is because it includes negative send deltas in result and I'm not sure whether this is correct, but this is how libwebrtc works. 
2. Notice that data captured in libwebrtc uses millisecond precision while str0m uses `Instant`. This obviously affects how everything in BWE module works, but I'm really not sure in what way exactly. Perhaps, better precision produces results that are more precise? But on the other hand it seems to be an [explicit decision on libwebrtc's side](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/rtp/transport_feedback_adapter.cc;l=197-200?q=TransportFeedbackAdapter::ProcessTransportFeedback&ss=chromium%2Fchromium%2Fsrc) that also has a todo that says that it is not necessary to confuse me even more

I can "fix" both these things in another PR, but I'm really not sure whether they should be fixed in the first place. So let me know what you think about this.


algesten:
Great idea! I looked a the code before I read your comment and was like OMG, we're adding cxx?! :D 

> * Notice that there are three data sets that tests are ran against, and the third one is [disabled](https://github.com/alexlapa/str0m/blob/f0f1ea355f2de106178c9afe9c3b7639b3a220ca/src/packet/bwe/arrival_group.rs#L5052-L5053). This is because it includes negative send deltas in result and I'm not sure whether this is correct, but this is how libwebrtc works.

I think negative send deltas are fine. We should handle them. Notice how the second type, large, is a i16. 

* https://datatracker.ietf.org/doc/html/draft-holmer-rmcat-transport-wide-cc-extensions-01#section-3.1.5
* https://github.com/algesten/str0m/blob/main/src/rtp/rtcp/twcc.rs#L85

> * Notice that data captured in libwebrtc uses millisecond precision while str0m uses `Instant`. This obviously affects how everything in BWE module works, but I'm really not sure in what way exactly. Perhaps, better precision produces results that are more precise? But on the other hand it seems to be an [explicit decision on libwebrtc's side](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/rtp/transport_feedback_adapter.cc;l=197-200?q=TransportFeedbackAdapter::ProcessTransportFeedback&ss=chromium%2Fchromium%2Fsrc) that also has a todo that says that it is not necessary to confuse me even more

I don't care if _at some point_ in our BWE code, we want to match libwebrtc and we use milliseconds instead of `Instant`. However our "API", both external and internal is `Instant`, so I'd say everything in `twcc.rs` is `Instant`, the interactions with `session.rs` are `Instant`, but if it's helpful we could change `packet/bwe/*` to work on milliseconds.

build.rs
thomaseizinger, line 1: I'd appreciate if we can move this to a different crate such that it can be feature-flagged out. I'd like to avoid a gcc dependency when I am only compiling the `IceAgent` (i.e. `str0m` without default-features).

algesten: @thomaseizinger it's not staying around!

It's just brought in to compare the implementations with the tests. Once the str0m impl is adjusted to match, the cxx stuff goes away.

alexlapa:
@algesten ,

So [here is a successful](https://github.com/algesten/str0m/actions/runs/12735410397/job/35494139947) CI run under ubuntu with that libwebrtc captured data (other targets have some cpp compilation issues which I haven't looked at since it does not really matter).

#### I've removed all irrelevant stuff and this is ready to be merged.

> I think negative send deltas are fine. We should handle them. Notice how the second type, large, is a i16.

Okay, I'll take a closer look at this a bit later.

> I don't care if at some point in our BWE code, we want to match libwebrtc and we use milliseconds instead of Instant.

Yea, that would be an internal `packet/bwe` thing only i imagine. Maybe `rtp/rtcp/twcc.rs` also.

> but if it's helpful we could change packet/bwe/* to work on milliseconds.

And that's the main problem - I am not sure at all whether this would be helpful or not :D 
The only reason why I'm bringing this up is that this is how libwebrtc works. I guess I'll let the current version run for a while and maybe return to this question later.

algesten (approved):
I have no idea what this does, but if it fixes our impl to be like libWebRTC it's 👍  with me.

algesten:
Any thoughts from you @k0nserv?

k0nserv:
@alexlapa I'll have a look, did you base this on a recent version of libWebRTC or the same commit that we initially based things on(I believe it's recorded somewhere). I don't recall if diverged from their implementation or if they have improved it.

The concern would be that using more recent versions of this code, but older versions of the surrounding code might not work well.

k0nserv:
Btw I don't see why we shouldn't use the most precise time available to us(so staying with `Instant` seems okay). Agree with @algesten we should handle negative send deltas

src/packet/bwe/arrival_group.rs
```diff
@@ -1,11 +1,14 @@
+use std::cmp::Ordering;
 use std::mem;
 use std::time::{Duration, Instant};
 
 use crate::rtp_::SeqNo;
 
 use super::AckedPacket;
 
-const BURST_TIME_INTERVAL: Duration = Duration::from_millis(5);
+const BURST_TIME_INTERVAL: f64 = 0.005; // 5ms
```
k0nserv, line 9: I think we should use Duration, it's a richer type which is less prone to mistakes

alexlapa:
@k0nserv ,

> did you base this on a recent version of libWebRTC or the same commit(14e2779a6ccdc67038ed2069a5732dd41617c6f0) that we initially based things on?

Based on current master (63334690f216949b3d57e2931c9c4cd28621e500) but there is [little difference](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/congestion_controller/goog_cc/inter_arrival_delta.cc;l=1;bpv=1;bpt=0;drc=63334690f216949b3d57e2931c9c4cd28621e500;dlc=14e2779a6ccdc67038ed2069a5732dd41617c6f0) between these committs.

k0nserv (approved):
Nice!

algesten:
I made a new branch saving away your work for the future. https://github.com/algesten/str0m/tree/bwe-compare-libwebrtc

This is from the commit `run tests` before fiddling with the ci and removing cxx again.

src/packet/bwe/arrival_group.rs
alexlapa: `Duration::as_secs_f64` is not `const` on current MSRV :(

[CI run](https://github.com/algesten/str0m/actions/runs/12735618126/job/35494573846):
```
error: `Duration::as_secs_f64` is not yet stable as a const fn
Error:  --> src/packet/bwe/arrival_group.rs:9:34
  |
9 | const BURST_TIME_INTERVAL: f64 = Duration::from_millis(5).as_secs_f64();
  |                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `str0m` (lib) due to previous error
```

algesten: Ah! Leave a comment for what MSRV we can switch.

k0nserv: I'm not suggesting `Duration::from_millis(5).as_secs_f64()`. We don't need Duration to transform `5` into `0.005`. My point is to try and use Durations rather than converting to seconds at all, if we must convert to seconds we can use `as_secs_f64` at the callsite

alexlapa: Ah, right, it just seemed like a minor inefficency, but whatever, fixed!

src/packet/bwe/arrival_group.rs
```diff
@@ -51,38 +54,47 @@ impl ArrivalGroup {
...
+        let inter_arrival_time = {
+            let last_remote_recv_time = self.remote_recv_time();
 
-        let inter_arrival_time = packet
-            .remote_recv_time
-            .checked_duration_since(self.remote_recv_time());
+            if packet.remote_recv_time >= last_remote_recv_time {
+                (packet.remote_recv_time - last_remote_recv_time).as_secs_f64()
+            } else {
+                (last_remote_recv_time - packet.remote_recv_time).as_secs_f64() * -1.0
```
thomaseizinger, line 71: Is the `* -1.0` the reason we convert to floats? I might not have the full picture so take with a grain of salt but it could potentially be modeled more type-safe if we do:

```rust
enum SignedDuration {
	Positive(Duration),
	Negative(Duration)
}
```

thomaseizinger (approved):

src/packet/bwe/arrival_group.rs
k0nserv: I was confused by this code. Until I remembered that `Duration` cannot be negative :( I think having a wrapper type like this would be good, but probably a good future improvement

algesten: In ureq I did this: https://github.com/algesten/ureq/blob/main/src/unversioned/transport/time.rs#L9-L24

I want to copy this code across to str0m and merge it and replace our similar concepts here: https://github.com/algesten/str0m/blob/main/src/util/time_tricks.rs#L6-L10

This new wrapper type could also handle negative durations.

algesten:
Excellent work. Thanks!

src/packet/bwe/arrival_group.rs
alexlapa: > Is the * -1.0 the reason we convert to floats?

Well it [was](https://github.com/algesten/str0m/blob/b6905402fff3e23b7ea3d1a85ace7bdedad000c7/src/packet/bwe/arrival_group.rs#L76-L77) being converted anyway, but for the pretty much the same reason: so `inter_group_delay_delta` could be negative.

But yeah, difference in time calculations is a problem: I can imagine that there might be other cases when `Instant`s subtraction returns `Duration::ZERO` when in the libwebrtc's implementation that would be a negative [`TimeDelta`](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/units/time_delta.h;drc=e0a524b5e02f293a0c13fd687334f7e6eedb971d).

algesten: @alexlapa sounds great!

You can join our Zulip if you want a simpler way to chat ;)

Issue #611, opened 2025-01-16
`rids_rx()` and `rids_tx()` return the wrong value when updating an existing non-simulcast track


OxleyS:
I'm still investigating this one, but it's a severe enough bug to break our project, so I wanted to get this issue out ahead of any potential `str0m` releases.

This seems to be a regression introduced by https://github.com/algesten/str0m/commit/425a851a2de2d3af378c15b55d7e9ed16ca6a1b7. When a WebRTC client changes the direction of their audio track which does not use simulcast, `rids_rx()` starts returning `Rids::Specific(vec![])` instead of `Rids::Any`.

[This code block](https://github.com/algesten/str0m/commit/425a851a2de2d3af378c15b55d7e9ed16ca6a1b7#diff-5fd5a802394cab016425ddca33d909c07672077403390ccafa359c7d55dcbed0R1084-R1092) seems to not account for the non-RIDs case, but I'm still looking into what exactly needs to be done instead.

algesten:
Doh! Direction changes!

I understand why this is happening.

OxleyS:
I'm writing up a fix now :+1:

PR #612, opened 2025-01-16
Fix RIDs bug on updating non-simulcast media

OxleyS:
Fixes https://github.com/algesten/str0m/issues/611

https://github.com/algesten/str0m/commit/425a851a2de2d3af378c15b55d7e9ed16ca6a1b7 added a `From` impl that converted from a slice of RIDs into a `Rids` instance, and unconditionally used it when updating media via SDP. However, this impl did not take into account empty slices, which are naturally produced when simulcast is not in play. This caused non-simulcast tracks to start reporting `Rids::Specific(vec![])` once they were updated by e.g. a direction change.

As the correct `Rids` value on an empty slice depends on whether you're sending or receiving, this PR removes the `From` impl entirely and replaces it with manual conversion while updating media via SDP.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -8,7 +8,7 @@ use crate::crypto::Fingerprint;
 use crate::format::CodecConfig;
 use crate::format::PayloadParams;
 use crate::io::Id;
-use crate::media::{Media, Simulcast};
+use crate::media::{Media, Rids, Simulcast};
 use crate::packet::MediaKind;
 use crate::rtp_::MidRid;
 use crate::rtp_::Rid;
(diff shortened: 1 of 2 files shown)


algesten (approved):

algesten:
Let's land it.

There's a bigger question here about the semantic meaning of Rids `Specific`, `Any` and `None`. Before `None` the meaning was

* `Any` - used by direct api since we don't have SDP to specify which specific RIDs to expect as incoming. That means the streams are allocated without needing to check they were declared up front
* `Specific` - used in SDP api to know what has been negotiated

I think I introduced `None` to be able to separate the case where in an SDP negotiation one side says I want x, ,y, z and the other responds with empty vs there being no rids in question at all.

Issue #610, opened 2025-01-15
Unbounded memory growth handling reception reports with stats disabled


OxleyS:
Upon getting a reception report, [this function](https://github.com/algesten/str0m/blob/ecf1e78b3989342c8923cc24f157e58c9130d482/src/streams/send.rs#L1106-L1107) is called, which stores the loss fraction in the `self.losses` Vec. This Vec is later consumed [here](https://github.com/algesten/str0m/blob/ecf1e78b3989342c8923cc24f157e58c9130d482/src/streams/send.rs#L1110) when generating a stats report.

However, stats reports are only generated if [configured to do so](https://github.com/algesten/str0m/blob/ecf1e78b3989342c8923cc24f157e58c9130d482/src/lib.rs#L1756-L1764). If not configured as such, `self.losses` will never be cleared and will grow unbounded (until the instance is destroyed, of course).

I wasn't sure how to best address this. While it would probably be optimal to avoid appending to `self.losses` at all, that would require up-front knowledge about whether stats are being collected, which is pretty hard to thread through. We could unconditionally generate stats and throw them out, but of course that's a waste of CPU for users that don't need them. We could add an alternative function that's called instead of `StreamTxStats.fill()` when stats is disabled, but that has a chance to bitrot if stats-related code changes.

While we're here, it may be good to audit for other uses of `Vec` in stats-related stuff.

thomaseizinger:
I think what we should be able to do is change the stream to also have an `Option` of stats only if the `Rtc` instance has an option of stats (or a dummy instance that performs no-op's).

Whether or not stats are enabled is known through the config as far as I can see!

algesten:
I don't think it should be too bad to pass down whether stats gathering should be enabled or not.

PR #613, opened 2025-01-16
Don't retain Receiver Report loss info when stats are disabled

OxleyS:
Fixes https://github.com/algesten/str0m/issues/610

`StreamTxStats` cannot be `Option`'d out entirely if stats reports are disabled, because it's still used for other purposes, e.g. RTX ratio calculation. Trying to split it out into a stats-reports and non-stats-reports portion was also awkward, due to overlap.

Instead, we give it a hint that `fill()` will never be called, which will prevent it from creating a `losses` array at all. Fortunately, bad/missing loss data was already accounted for, so even if it does end up being called, the behavior will still be sane (i.e. no panics, no bogus data).

I also considered a solution that would just put an upper bound on the `losses` array instead, but that had a few downsides:
1. We'd need to decide on overflow behavior, of which I don't have a good opinion.
2. Users that don't need stats would always eat the overhead of a full array, plus overflow behavior.

### DIFF
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -153,7 +153,7 @@ pub(crate) struct StreamTxStats {
     /// Can be null in case of missing or bad reports
     rtt: Option<f32>,
     /// losses collecter from RR (known packets, lost ratio)
-    losses: Vec<(u64, f32)>,
+    losses: Option<Vec<(u64, f32)>>,
 
     /// `None` if `rtx_ratio_cap` is `None`.
     bytes_transmitted: Option<ValueHistory<u64>>,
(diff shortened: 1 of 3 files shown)


src/streams/send.rs
```diff
@@ -153,7 +153,7 @@ pub(crate) struct StreamTxStats {
     /// Can be null in case of missing or bad reports
     rtt: Option<f32>,
     /// losses collecter from RR (known packets, lost ratio)
-    losses: Vec<(u64, f32)>,
+    losses: Option<Vec<(u64, f32)>>,
```
thomaseizinger, line 156: What I would suggest is, to make a newtype called `Losses` that has two constructors:

- `new`
- `no_op` / `dummy` etc

We then pass `Losses` as a parameter through all layers that don't yet know whether capturing `Losses` will be required. At the point where we have the config available, we can then decide which constructor to call.

This newtype can also encapsulate some of the functionality on how the data is actually processed and inserted, i.e. provide a more specific interface for the other components to interact with.

If you wanted to go full on type-safe, we could also define `Losses` as an enum such that in the no-op mode, there isn't even an underlying data-structure to operate on / fill.

src/streams/send.rs
```diff
@@ -1097,41 +1103,42 @@ impl StreamTxStats {
...
         self.rtt = rtt;
 
-        let ext_seq = {
-            let prev = self.losses.last().map(|s| s.0).unwrap_or(r.max_seq as u64);
-            let next = (r.max_seq & 0xffff) as u16;
-            extend_u16(Some(prev), next)
-        };
+        if let Some(losses) = self.losses.as_mut() {
+            let ext_seq = {
+                let prev = losses.last().map(|s| s.0).unwrap_or(r.max_seq as u64);
```
thomaseizinger, line 1108: All this logic could go onto the `Losses` newtype.

src/streams/send.rs
```diff
@@ -1097,41 +1103,42 @@ impl StreamTxStats {
...
+        }
     }
 
     pub(crate) fn fill(&mut self, snapshot: &mut StatsSnapshot, midrid: MidRid, now: Instant) {
         if self.bytes == 0 {
             return;
         }
 
-        let loss = {
+        let loss = self.losses.as_mut().and_then(|losses| {
```
thomaseizinger, line 1122: Similar here, this computation can be moved onto the `Losses` newtype.

thomaseizinger (commented):
Thanks for tackling this!

I left an idea on how I think this could be solved in a more type-safe manner. Before you tackle it, let's hear what @algesten thinks of it :)

PR #614, opened 2025-01-16
Don't retain Receiver Report loss info when stats are disabled (take 2)

### LINKED ISSUE #610: Unbounded memory growth handling reception reports with stats disabled

algesten:
See description in #610

Fixes #610
Fixes #613

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -128,10 +128,12 @@ impl Session {
             (PacerImpl::Null(NullPacer::default()), None)
         };
 
+        let enable_stats = config.stats_interval.is_some();
+
         Session {
             id,
             medias: vec![],
-            streams: Streams::default(),
+            streams: Streams::new(enable_stats),
             app: None,
             reordering_size_audio: config.reordering_size_audio,
             reordering_size_video: config.reordering_size_video,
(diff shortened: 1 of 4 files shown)


algesten:
@OxleyS @thomaseizinger i took your respective ideas and built a bit on it.

I wasn't allowed to push to #610, so I made this PR.

algesten:
Thanks for both your help and input!

PR #615, opened 2025-01-17
Support negative time deltas in BWE

alexlapa:
So first of all we should be able to handle negative send deltas in [`ArrivalGroup`](https://github.com/algesten/str0m/blob/25d51d539097df71c593553d00e9b6693db24107/src/packet/bwe/arrival_group.rs#L196) as discussed before: [0](https://github.com/algesten/str0m/pull/608#issue-2780561380), [1](https://github.com/algesten/str0m/pull/608#issuecomment-2585142506), [2](https://github.com/algesten/str0m/pull/608#issuecomment-2589567660)

So this is pretty much what this PR does so far, but it might be extended as discussed [here](https://github.com/algesten/str0m/pull/608#discussion_r1914662503).

So, I can leave it as it is, right now it only affects `packet/bwe` and these new types are only used where it's expected that negative time deltas are possible. Alternatively:
1. Always prefer using these new types instead of std's in BWE? They work pretty much like libwebrtc's `Timestamp` and `TimeDelta` so they should fit.
2. Also use it to cover similar concepts as `not_happening()`, `already_happened()`? I haven't really though about it but first thing that comes to mind is just using `SuperInstant` instead of those `Instant`s that can be `already_happened` or `not_happening` with some simple conversion `SuperInstant::as_instant() -> Instant` that will just return `already_happened` || `Exact(Instatnt)` || `not_happening`. Might also just put a `Deref -> &Instant` on it.

So let me know what you think about this.

Also I've returned that `ArrivalGroupAccumulator` test on libwebrtc captured data, but [now it also runs on negative send deltas](https://github.com/algesten/str0m/blob/negative-time-deltas/src/packet/bwe/arrival_group.rs#L5451). Ill remove it before merge.


algesten:
Looks good.

Yeah. So I want to replace as much as possible of the `time_hacks.rs`. Ideally I'd like to call these things `Instant` and `Duration` (instead of `SuperInstant` and `TimeDelta`). The names deliberately shadow the std-lib ones because they should function 99% like the std-lib does. That means a casual reader of the code should not be surprised by new time types.

Our external API is however std-lib `Instant` and `Duration`, which means there's work to go through exactly all the internal code where this would affect.

I'd prefer if the `SuperInstant`/`TimeDelta` could be moved to a separate PR that handles that and leave this PR to only deal with the BWE consequences.

thomaseizinger:
> Our external API is however std-lib `Instant` and `Duration`, which means there's work to go through exactly all the internal code where this would affect.

Might be worth adopting something like https://github.com/obi1kenobi/cargo-semver-checks to ensure changes to the public API are automatically linted!

alexlapa:
> Might be worth adopting something like https://github.com/obi1kenobi/cargo-semver-checks to ensure changes to the public API are automatically linted!

Well in this specific case [`E0446`](https://doc.rust-lang.org/error_codes/E0446.html) should be enough I suppose.

algesten:
Given the findings in #617. Let's go further with this PR by making the negative-time handling variants local to the BWE module. `SuperInstant` is maybe a questionable name, but `TimeDelta` is good. Let's isolate them to bwe module so they don't spread across the rest of the code base for now.

And again, sorry for making you doing a wild goose chase in a direction that doesn't work.

src/packet/bwe/time.rs
```diff
@@ -0,0 +1,739 @@
...
+    MinusInf,
+
+    /// An exact negative duration.
+    Negative(Duration),
+
+    /// An exact positive duration.
+    Positive(Duration),
+
+    /// Time delta to some event in distant future that will never happen.
+    PlusInf,
```
algesten, line 35: `PositiveInfinity`?

src/packet/bwe/time.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// A time in the future that will never happen.
+    DistantFuture,
+}
+
+/// Wrapper for [`Duration`] that can be negative and provides a duration to a
+/// distant future or past.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(super) enum TimeDelta {
+    /// Time delta to some event in distant past that already happened.
+    MinusInf,
```
algesten, line 26: `NegativeInfinity`?

src/packet/bwe/time.rs
```diff
@@ -0,0 +1,739 @@
...
+impl TimeDelta {
+    pub(super) const ZERO: Self = Self::Positive(Duration::ZERO);
+
+    /// Returns the number of seconds contained by this [`TimeDelta`] as `f64`.
+    pub fn as_secs_f64(&self) -> f64 {
+        match self {
+            Self::Negative(d) => d.as_secs_f64().neg(),
+            Self::Positive(d) => d.as_secs_f64(),
+            Self::PlusInf => f64::INFINITY,
+            Self::MinusInf => f64::NEG_INFINITY,
```
algesten, line 47: Reorder these to the order of the enum values.

src/packet/bwe/time.rs
```diff
@@ -0,0 +1,739 @@
...
+    /// Indicates whether this [`SuperInstant`] is [`SuperInstant::Exact`].
+    pub const fn is_finite(&self) -> bool {
+        matches!(self, Self::Exact(_))
+    }
+
+    /// Indicates whether this [`SuperInstant`] is [`SuperInstant::DistantPast`]
+    /// or [`SuperInstant::DistantFuture`].
+    pub const fn is_not_finite(&self) -> bool {
+        !self.is_finite()
+    }
```
algesten, line 83: I think the word "finite" here is introducing another name for something that is already called "exact" in the enum.

I suggest we stick with one of them. My preference is `is_exact` (do we need `is_not_exact`?)

src/packet/bwe/time.rs
```diff
@@ -0,0 +1,739 @@
...
+
+impl Add<TimeDelta> for SuperInstant {
+    type Output = Self;
+
+    fn add(self, rhs: TimeDelta) -> Self::Output {
+        match (self, rhs) {
+            (Self::DistantFuture, _) | (_, TimeDelta::PlusInf) => Self::DistantFuture,
+            (Self::DistantPast, _) | (_, TimeDelta::MinusInf) => Self::DistantPast,
+            (Self::Exact(i), TimeDelta::Negative(d)) => Self::Exact(i - d),
+            (Self::Exact(i), TimeDelta::Positive(d)) => Self::Exact(i + d),
```
algesten, line 94: Sadly Rust doesn't seem to have `Instant::checked_sub(Duration)`, which would fit neatly into this mechanic. Going via u128 to do it manually seems expensive, so let's leave it like this. Maybe with a comment to look out for future Rust std API.

algesten (approved):
LGTM! Some minor comments

src/packet/bwe/time.rs
alexlapa: There is an [`Instant::checked_sub(Duration)`](https://doc.rust-lang.org/std/time/struct.Instant.html#method.checked_sub). And what you mean is going `DistantFuture` on overflow and `DistantPast` on underflow instead of panicking? Well, I thought that panicking is a better option since if overflow/underflow happens than this is most likely an unexpected error. But that probably won't ever happen so Im okay with changing it if you want

algesten: Alright! You thought about it. Let's leave it.

algesten (approved):

algesten:
@alexlapa Thanks!!!

PR #616, opened 2025-01-21
Remove restriction on EC to use on ECDH exchange

efer-ms:
Str0m cannot complete a DTLS handshake with a peer using a certificate signed with an secp384r1 key. The DTLS server hello indicates that that key is supported, however, Str0m (when using openssl) will reject fail the DTLS handshake. The issue is that Str0m is restricting OpenSSL to only use/accept the secp256r1 EC.

---- More details ----

set_tmp_ecdh() is deprecated in OpenSSL, and under the covers will call set_groups() with the EC group of the provided key. This restricts our session to the one given EC group (secp256r1/PRIME256V1 in the code) [See https://github.com/openssl/openssl/discussions/23070#discussioncomment-7885148]. This prevents session establishment with DTLS implementations using secp384r1 or another valid EC.

By simply removing the call, we are allowing OpenSSL to decide which ECs to support. Which is at least: x25519, secp256r1 (prime256v1), secp521r1, and secp384r1.

The upside is that we let OpenSSL manage which EC are "safe" to use. The downside is that we rely on OpenSSL maintaining the default as a safe list, should weaknesses be discovered in the ECs.

### DIFF
--- a/src/crypto/ossl/dtls.rs
+++ b/src/crypto/ossl/dtls.rs
@@ -2,8 +2,6 @@ use std::collections::VecDeque;
 use std::io::{self, Read, Write};
 use std::time::{Duration, Instant};
 
-use openssl::ec::EcKey;
-use openssl::nid::Nid;
 use openssl::ssl::{Ssl, SslContext, SslContextBuilder, SslMethod, SslOptions, SslVerifyMode};
 
 use crate::crypto::dtls::DtlsInner;
@@ -15,8 +13,7 @@ use super::io_buf::IoBuffer;
 use super::stream::TlsStream;
 use super::CryptoError;
 
-const DTLS_CIPHERS: &str = "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
-const DTLS_EC_CURVE: Nid = Nid::X9_62_PRIME256V1;
+const DTLS_CIPHERS: &str = "ECDHE+AESGCM:DHE+AESGCM:ECDHE+AES256:DHE+AES256";
 
 pub struct OsslDtlsImpl {
     /// Certificate for the DTLS session.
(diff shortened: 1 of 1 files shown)


src/crypto/ossl/dtls.rs
```diff
@@ -15,8 +13,9 @@ use super::io_buf::IoBuffer;
 use super::stream::TlsStream;
 use super::CryptoError;
 
-const DTLS_CIPHERS: &str = "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH";
-const DTLS_EC_CURVE: Nid = Nid::X9_62_PRIME256V1;
+// We restrict cipher suites to those that include ephermeral Diffie-Hellman or ephemeral
+// Elliptical Curve Diffie-Hellman AND AES-256 or AES-GCM.
+const DTLS_CIPHERS: &str = "ECDHE+AESGCM:DHE+AESGCM:ECDHE+AES256:DHE+AES256";
```
algesten, line 18: * EECDH -> ECDHE
* EDH -> DHE

I'm trying to Google and don't quite understand this change. We already included ephemeral Diffie-Hellmans but named a bit differently. What is this change?

algesten (changes requested):

src/crypto/ossl/dtls.rs
efer-ms: Just consistency in order of the key exchange and cipher... and having the acronym orders match the typical naming I've seen. I can revert, but this seemed more consistent to me.

efer-ms: Just to clarify a bit more

The Key exchange and cipher had different orders in the old string. Where it had KEY_EXCHANGE+CIPHER for 2 of the selectors, and CIPHER_KEY_EXCHANGE for the other two. So that was making the line consistent in ordering.

The EDH vs DHE and EECDH vs ECDHE is also about being consistent, but here it's about being consistent with the typical order used. DHE and ECDHE is what I've seen in openssl (eg. openssl ciphers), as well as Wireshark when it decodes packets.

This change is purely about consistency, there is no functional difference.

algesten (approved):

PR #619, opened 2025-02-02
Support AV1-specific media format parameters

alexlapa:
This PR adds AV1-specific parameters to `FormatParams` struct and `FormatParam` enum + parsing and serialization.


algesten (approved):
The cleanup of parsing the existing parameters could have been its own commit, but let's roll with it.

algesten:
Thanks!

PR #621, opened 2025-02-10
Use spoofed raddr in srflx and relay ICE candidates

OxleyS:
A PR arising from the discussion in a [Zulip thread](https://str0m.zulipchat.com/#narrow/channel/377845-general/topic/Another.20.28.3F.29.20Firefox.20connectivity.20problem).

Firefox demands that all non-host ICE candidates (that is, server-reflexive, peer-reflexive, and relayed candidates) contain `raddr` and `rport` fields, or it fails to parse. Putting the real related address here is a privacy issue, and remote clients don't use the value anyway, so in practice browsers often put a dummy address there instead.

Here, we do the same for server-reflexive and relay candidates generated by us. I left out peer-reflexive candidates because they're not typically signaled anyway, and I didn't want to affect any existing ICE logic.


algesten:
Looks good! Just fix the MSRV since we don't want to move that.

OxleyS:
Oh, sorry, I wasn't paying attention to that! I've amended the commit accordingly.

PR #624, opened 2025-03-19
Fix for very long media pauses

lolgesten:


src/streams/receive.rs
```diff
@@ -381,8 +381,41 @@ impl StreamRx {
...
+        // If we're coming out of a paused state and the timestamp would go backwards,
+        // incrementally adjust until we find a value that moves time forward
+        if self.paused && previous_time.is_some() && time_u32 < previous_time.unwrap() {
+            // In 32-bit RTP timestamps, adding 2^31 (MAX/2) flips to the other half of timestamp space
+            // This forces extend_u32 to produce a value in the next cycle
+            let half_cycle = 1u32 << 31;
+            let mut adjusted_ts = header.timestamp;
+
+            // Try at most twice (should be sufficient to find a suitable timestamp)
+            for _ in 0..2 {
```
k0nserv, line 399: Adding `1 << 31` twice just brings you back to the same value so this doesn't need to be a loop

src/streams/receive.rs
```diff
@@ -381,8 +381,41 @@ impl StreamRx {
...
+                // If this adjusted timestamp moves time forward, use it
+                if adjusted_time_u32 > previous_time.unwrap() {
+                    time_u32 = adjusted_time_u32;
+                    break;
+                }
+            }
+
+            // If we still couldn't find a timestamp that moves forward, force time ahead
+            if time_u32 < previous_time.unwrap() {
+                time_u32 = previous_time.unwrap() + 1;
```
k0nserv, line 415: Won't this cause lots of weirdness. If we end up here, chances are we'll end up here again for the next frame and the resulting effect will be frames advancing extremely quickly e.g. instead of having 2700 ticks per frame for 90kHZ video we'll have 1 tick per frame until we can escape this condition

k0nserv (approved):

src/streams/receive.rs
algesten: True

src/streams/receive.rs
algesten: Right. Let's just reset in this case.

algesten:
See #625

PR #626, opened 2025-03-19
Direct API reset_stream_tx() call

algesten:


src/change/direct.rs
```diff
@@ -281,4 +281,28 @@ impl<'a> DirectApi<'a> {
...
+        rid: Option<Rid>,
+        new_ssrc: Ssrc,
+    ) -> Option<&mut StreamTx> {
+        let midrid = MidRid(mid, rid);
+
+        // Find the stream by mid/rid
+        let stream = self.rtc.session.streams.stream_tx_by_midrid(midrid)?;
+
+        // Reset the stream with the new SSRC
+        stream.reset_ssrc(new_ssrc);
```
algesten, line 303: I should check the SSRC is not the same as the old one.

src/streams/send.rs
```diff
@@ -1016,6 +1016,37 @@ impl StreamTx {
         self.padding = 0;
     }
 
+    /// Reset this stream to use a new SSRC.
+    ///
+    /// This updates the SSRC and resets all relevant internal fields.
+    pub(crate) fn reset_ssrc(&mut self, new_ssrc: Ssrc) {
+        // Update the SSRC
+        self.ssrc = new_ssrc;
```
k0nserv, line 1024: I'm wondering if we should reset the RTX SSRC to, you could imagine the remote associating the RTX SSRC with the main SSRC such that any future sends on the RTX end up going to the old one

src/streams/send.rs
```diff
@@ -1016,6 +1016,37 @@ impl StreamTx {
         self.padding = 0;
     }
 
+    /// Reset this stream to use a new SSRC.
+    ///
+    /// This updates the SSRC and resets all relevant internal fields.
+    pub(crate) fn reset_ssrc(&mut self, new_ssrc: Ssrc) {
```
k0nserv, line 1022: Stats too maybe?

k0nserv (approved):

src/streams/send.rs
algesten: Yeah, makes sense. Did you observe that with FF?

k0nserv: I'll check

k0nserv: image

I did get new RTX streams (look at the Payload, SSRC, and start time)

algesten: 👍 Super! Thanks!

PR #630, opened 2025-03-23
Avoid calls std::env::set_var to configure EnvFilter.

mstyura:
Rust 2024 treats `std::env::set_var` as unsafe function and it is possible to avoid it using direct log filter configuration API.


algesten (commented):
Looks good!

Thanks!

algesten (approved):

PR #631, opened 2025-04-02
Feature/turn messages

k0nserv:


src/io/mod.rs
```diff
@@ -11,11 +11,10 @@ use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
 mod stun;
-pub use stun::StunMessage;
 pub(crate) use stun::{
-    Class as StunClass, Method as StunMethod, StunError, StunTiming, TransId,
-    DEFAULT_MAX_RETRANSMITS,
+    Class as StunClass, Method as StunMethod, StunError, StunTiming, DEFAULT_MAX_RETRANSMITS,
 };
+pub use stun::{StunMessage, StunMessageBuilder, TransId as StunTransId};
```
lolgesten, line 17: I don't think we need this renaming of TransId -> StunTransId. It causes unnecessary code churn.

src/ice/agent.rs
```diff
@@ -984,6 +984,19 @@ impl IceAgent {
...
+                StunMethod::Allocate
+                | StunMethod::Refresh
+                | StunMethod::Send
+                | StunMethod::Data
+                | StunMethod::CreatePermission
+                | StunMethod::ChannelBind,
+                _,
+            ) => {
+                // Unexpected TURN related message
+                trace!("Message rejected, TURN method unexpected in this context");
```
lolgesten, line 997: Log the method?

src/io/stun.rs
```diff
@@ -406,6 +507,12 @@ impl Method {
         use Method::*;
         match self {
             Binding => 0b0000_0000_0000_0001,
+            Allocate => 0b0000_0000_0000_0011,
+            Refresh => 0b0000_0000_0000_0100,
+            Send => 0b0000_0000_0000_0110,
+            Data => 0b0000_0000_0000_0111,
+            CreatePermission => 0b0000_0000_0000_1000,
+            ChannelBind => 0b0000_0000_0000_1001,
```
lolgesten, line 515: Turn off rustfmt here and align the numbers.

src/io/stun.rs
```diff
@@ -503,21 +630,28 @@ use crate::util::NonCryptographicRng;
...
     const ICE_CONTROLLED: u16 = 0x8029;
     const ICE_CONTROLLING: u16 = 0x802a;
+
     const MAPPED_ADDRESS: u16 = 0x0001;
+    const USERNAME: u16 = 0x0006;
     const MESSAGE_INTEGRITY: u16 = 0x0008;
-    const NETWORK_COST: u16 = 0xc057;
-    const NONCE: u16 = 0x0015;
-    const PRIORITY: u16 = 0x0024;
+    const ERROR_CODE: u16 = 0x0009;
```
lolgesten, line 640: why does this one move?

src/io/stun.rs
```diff
@@ -550,19 +684,63 @@ impl<'a> Attributes<'a> {
...
     fn to_bytes(self, vec: &mut dyn Write, trans_id: &[u8]) -> io::Result<()> {
         if let Some(v) = self.username {
             vec.write_all(&Self::USERNAME.to_be_bytes())?;
-            vec.write_all(&(v.len() as u16).to_be_bytes())?;
-            vec.write_all(v.as_bytes())?;
-            let pad = 4 - (v.len() % 4) % 4;
-            for _ in 0..pad {
-                vec.write_all(&[0])?;
-            }
+            encode_str(Self::USERNAME, v, vec)?;
```
lolgesten, line 743: let's rename `vec` to `out` here.

src/io/stun.rs
```diff
@@ -737,6 +998,22 @@ impl<'a> Attributes<'a> {
     }
 }
 
+fn encode_str(typ: u16, s: &str, vec: &mut dyn Write) -> io::Result<()> {
```
lolgesten, line 1001: same here `vec`

src/io/stun.rs
```diff
@@ -815,11 +1092,256 @@ impl<'a> fmt::Debug for StunMessage<'a> {
...
+                class: self.state.class,
+                trans_id,
+                attrs: self.attrs,
+                integrity: &[],   // Calculated during serialization
+                integrity_len: 0, // Calculated during serialization
+            }
+        }
+    }
+}
+
```
lolgesten, line 1340: Nice builder!

lolgesten (approved):
Looks great!

algesten (approved):

src/io/mod.rs
k0nserv: My thinking was that all the other exports are prefixed Stun so for consistency it should also have the prefix. However, we could keep just `TransId` internally

src/ice/agent.rs
k0nserv: Sounds good

src/io/stun.rs
k0nserv: I re-ordered these to be in the order of their identifier

k0nserv: That's all in 3fcdd83c4a28d2b86f0b7dd4c162b1a425189d04

src/io/stun.rs
k0nserv: Sure

src/io/mod.rs
lolgesten: Let's keep it for now and make a separate PR to address that.

thomaseizinger:
Is str0m getting a TURN client? :)

algesten:
> Is str0m getting a TURN client? :)

Hehehe… We are implementing one, but no decision on whether to keep it internal yet.

You interested?

thomaseizinger:
Let's just say I have opinions on it.

I think what exactly you need from a TURN client can be quite application specific so I am not fully convinced it makes sense to build a generic one.

Ours is here: https://github.com/firezone/firezone/blob/main/rust/connlib/snownet/src/allocation.rs

It has grown into a monster and I don't like it anymore 😂

Couple of features / bugfixes that we realised we wanted but didn't know from the beginning:

- Happy-eyeballs for picking either the IPv4 or IPv6 socket if the TURN server is dual-stack
- On-demand channel bindings (not just ahead of time reserved channels)
- Think about possible clock-drift between client and TURN server in regards to re-binding expired channels
- Implement STUN binding keep-alives with the TURN server to prevent the NAT binding from expiring when your software is idling

There is probably more that I missed 😁

algesten:
@thomaseizinger Thanks! Very good pointers!

thomaseizinger:
Happy to chat anytime if you want to run by some designs :)

PR #632, opened 2025-04-03
Optional integrity on STUN messages

k0nserv:
In some situations, for example before TURN allocation it's not possible to construct a valid `MESSAGE-INTEGRITY` attribute, therefore we need to make this possible.

Designwise I've initially made the password optional on `to_bytes`, but an alternate approach that I prefer is

```rust
    pub fn to_bytes(self, password: &str, buf: &mut [u8]) -> Result<usize, StunError> {
      self.to_bytes_internal(Some(password), buf)
    }

    pub fn to_bytes_no_auth(buf: &mut [u8]) -> Result<usize, StunError> {
      self.to_bytes_internal(Some(password), buf)
    }
```

The detailed example section from the TURN RFC is useful as an example of the flow https://datatracker.ietf.org/doc/html/rfc5766#autoid-55

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -145,24 +145,40 @@ impl<'a> StunMessage<'a> {
 
         let attrs = Attributes::parse(&buf[20..], trans_id, &mut message_integrity_offset)?;
 
-        // message-integrity only includes the length up until and including
-        // the message-integrity attribute.
-        if message_integrity_offset == 0 {
-            return Err(StunError::Parse("No message integrity in incoming".into()));
-        }
-
         // length including message integrity attribute
-        let integrity_len = (message_integrity_offset + 4 + 20) as u16;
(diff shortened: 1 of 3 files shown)


src/io/stun.rs
```diff
@@ -145,11 +145,13 @@ impl<'a> StunMessage<'a> {
...
         let attrs = Attributes::parse(&buf[20..], trans_id, &mut message_integrity_offset)?;
 
-        // message-integrity only includes the length up until and including
-        // the message-integrity attribute.
-        if message_integrity_offset == 0 {
-            return Err(StunError::Parse("No message integrity in incoming".into()));
-        }
+        // TODO: Determine when/how to perform message integrity checks.
+
+        // // message-integrity only includes the length up until and including
```
k0nserv, line 150: NB: We need a resolution on this too

algesten (approved):

PR #636, opened 2025-04-10
Fix incorrect padding in STUN attributes

k0nserv:
Some of the logic around calculating padding for variable length attributes resulted in extra padding when none was necessary.

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -653,6 +653,7 @@ use std::{io, str};
 
 use crate::util::NonCryptographicRng;
 
+const PAD: [u8; 4] = [0, 0, 0, 0];
 impl<'a> Attributes<'a> {
     const ALTERNATE_SERVER: u16 = 0x8023;
     const FINGERPRINT: u16 = 0x8028;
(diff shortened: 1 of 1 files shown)


src/io/stun.rs
```diff
@@ -1456,6 +1457,24 @@ mod test {
...
+            ..Default::default()
+        };
+        let mut buf = vec![];
+        let trans_id = TransId::new();
+        attrs
+            .to_bytes(&mut buf, &trans_id.0)
+            .expect("To serialize attributes");
+        assert_eq!(
+            buf.len(),
+            8,
```
k0nserv, line 1473: Prior to this, we would incorrectly add 4 extra 0 bytes in this case

src/io/stun.rs
```diff
@@ -683,10 +684,7 @@ impl<'a> Attributes<'a> {
 
         let username = self
             .username
-            .map(|v| {
-                let pad = 4 - (v.len() % 4) % 4;
```
k0nserv, line 687: This should've been `(4 - (v.len() % 4)) % 4`. An alternative to this PR would be something smaller that just fixes this in the relevant places. 

Previously this would cause us to add 4 extra zero bytes for already aligned attributes.

src/io/stun.rs
```diff
@@ -1044,6 +1041,14 @@ impl<'a> Attributes<'a> {
     }
 }
 
+fn calculate_pad(len: usize) -> usize {
+    if len & 3 != 0 {
```
k0nserv, line 1045: Actually, `(4 - (len % 4)) % 4` generates better assembly than this, it's just

```asm
        neg     eax
        and     eax, 3
```

algesten:
Woohaaa. Lot's of changes. I have no idea if it's correct. But you typically know what' you're doing :)

PR #638, opened 2025-04-14
Add stats for the selected candidate pair

k0nserv:


src/lib.rs
```diff
@@ -727,7 +727,10 @@ mod session;
 use session::Session;
 
 pub mod stats;
-use stats::{MediaEgressStats, MediaIngressStats, PeerStats, Stats, StatsEvent, StatsSnapshot};
+use stats::{
+    CandidatePairStats, CandidateStats, MediaEgressStats, MediaIngressStats, PeerStats, Stats,
+    StatsEvent, StatsSnapshot,
+};
```
algesten, line 733: :'(

algesten (approved):

src/lib.rs
k0nserv: What's the problem?

PR #639, opened 2025-04-17
feat(ice): prefer relayed connections within the same IP version

thomaseizinger:
When calculating preferences for candidates, we currently always prefer IPv6 over IPv4. This can lead to sub-optimal situations when a connection ends up using a TURN server.

TURN allows a client to allocate an IPv4 and an IPv6 address in the same allocation. This makes it possible for e.g. an IPv4-only client to connect to an IPv6-only peer as long as the TURN server runs in dual-stack AND the client requests an IPv6 address in addition to an IPv4 address with the `ADDITIONAL-ADDRESS-FAMILY` attribute.

Assume that a client sits behind symmetric NAT and therefore needs to rely on a TURN server to communicate with its peers. The TURN server as well as all the other peers operate in dual-stack mode.

The current priority calculation will yield a communication path that uses IPv4 to talk to the TURN server (as that is the only one available) but due to the preference ordering of IPv6 over IPv4, will use an IPv6 path to the peer, despite the peer also supporting IPv4.

This isn't a problem per-se but makes (our) life unnecessarily difficult. We operate a fleet of TURN servers that use eBPF to efficiently deal with TURN's channel-data messages. Channel-data messages are just UDP packets with a 4 byte header. Adding / removing that is easy enough to do within an eBPF program, allowing for much greater throughput than copying bytes from kernel space to user space and back.

This doesn't work though if we need to also convert from IPv4 to IPv6 or vice versa. It is not technically impossible but more challenging to implement because we also need to translate the IP header from one version to the other instead of just updating source and destination addresses.

This patch extends the `Candidate::relayed` constructor to also take a `local_turn_socket` address which is - similar to the `base` address of other candidate types - the address the client is sending from in order to use this candidate. In the context of relayed candidates, this is the address the client is using to talk to the TURN server. We can use this information in the candidate's priority calculation to prefer candidates that allow traffic to remain within one IP version, i.e. if the client talks to the TURN server over IPv4, the allocated IPv4 address on the TURN server will have a higher priority than the allocated IPv6 address.

In addition to this priority calculation, we also patch the `add_local_candidate`. In there, we currently compute a fixed `local_preference` which - when set on the candidate - opts out of the priority calculation within `Candidate`.

Staying within the same IP version whilst relaying traffic allows our TURN servers to use their eBPF kernel which results in a much better UX due to lower latency and higher throughput.

In general, I'd consider the ability of TURN servers to translate between IP versions as a fallback mechanism. A kind of "last resort" way of establishing a connection if all else fails. But for that to be true, we need to take into account the base address that we are using to talk to the TURN server, otherwise IPv6 network paths will always win over the IPv4 one.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -528,7 +528,7 @@ impl IceAgent {
                 ServerReflexive => 32_767,
                 Relayed => 16_383,
             };
-            x - (if ip.is_ipv6() { 0 } else { 1 })
+            x - if ip.is_ipv6() { 0 } else { 1 }
         };
 
         // Count the number of existing candidates of the same kind.
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1766,8 +1766,6 @@ impl IceAgent {
 #[cfg(test)]
 mod test {
 
-    use crate::ice_::test::relay;
-
     use super::*;
     use std::net::SocketAddr;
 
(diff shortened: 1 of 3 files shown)


thomaseizinger:
@algesten I am still testing this change within our application but I'd like to get early feedback on this idea. As I've written in the description, it is (strongly) motivated by our current infrastructure. I do think that it is generally a worthwhile change to make. In other words, I don't think this situation is necessarily unique to us. I think we should avoid routing traffic "across" IP versions where possible and only prefer IPv6 if we can have a native IPv6 connection.

I am open to other ideas on how to implement this. An alternative could be to expose a more general "local preference offset" that applications can set on candidates in order to have an influence on their priority. I was going to implement that at first but figured it is would be nice to actually have this first-class in str0m so we can also write a test against it.

thomaseizinger:
Ah using the `base` field here is buggy because we use that one to set `Transmit::source`.

thomaseizinger:
Perhaps this should be modelled using `NETWORK-COST`? I don't see any integration of that beyond serialising and deserialising the attribute though.

algesten:
I think  `local_turn_socket` is leaking a bit too much into the abstraction.

* The candidate type is called "relay" presumably because there could be other relay types than TURN.
* We don't actually want the local socket address, only know if it's ipv4 or 6.

Would an alternative be to maybe configure `prefer_ipv4` on the IceAgent and this would change the order to be v4/v6/v4/v6... instead of the current order?

thomaseizinger:
> Would an alternative be to maybe configure `prefer_ipv4` on the IceAgent and this would change the order to be v4/v6/v4/v6... instead of the current order?

I don't want to prefer IPv4 though, I want to prefer allocations made on the same IP version. For example, if the client is using IPv6 to talk to the relay, I want the IPv6 candidate from the allocation to have a higher priority than the IPv4 one and vice versa.

> I think `local_turn_socket` is leaking a bit too much into the abstraction.
> 
>     * The candidate type is called "relay" presumably because there could be other relay types than TURN.
> 
>     * We don't actually want the local socket address, only know if it's ipv4 or 6.

I am open to other ideas. Would you be willing to expose a more general "local preference offset" mechanism on the `Candidate`?

The candidate type might be relay which is an abstraction I agree. The constructor of an abstraction must be something concrete though. Should we add another constructor or change the current one to be explicitely a TURN relay candidate?

algesten:
@thomaseizinger in #640 i did another take, which doubles down on using `base` to represent the local ip address. I've arranged it commit-by-commit to finally bring in your penalty change.

algesten:
Close in favor of #640

PR #640, opened 2025-04-18
Prefer relayed candidates with matching IP versions

algesten:
Relates to #639

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1416,7 +1416,7 @@ impl IceAgent {
 
         let local = pair.local_candidate(&self.local_candidates);
         let proto = local.proto();
-        let local_addr = local.base();
+        let local_addr = local.source_addr();
         let remote = pair.remote_candidate(&self.remote_candidates);
         let remote_addr = remote.addr();
 
@@ -1658,7 +1658,7 @@ impl IceAgent {
             self.nominated_send = Some(best_prio.id());
             self.emit_event(IceAgentEvent::NominatedSend {
                 proto: local.proto(),
-                source: local.base(),
+                source: local.source_addr(),
                 destination: remote.addr(),
             })
         }
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -539,7 +539,22 @@ impl IceAgent {
             .filter(|v| v.addr().is_ipv6() == ip.is_ipv6())
             .count() as u32;
 
-        let pref = counter_start - same_kind * 2;
+        // For relayed candidates, we add a "punishment" to the local preference
+        // if the base address differs in the IP version from the allocated address
+        // of the candidate.
+        // This punishment ensures that we prefer relayed within the same IP version,
+        // e.g. IPv4 <> IPv4 over ones that translate between IP version, e.g. IPv4 <> IPv6.
+        let relay_across_ip_version_punishment = if c.kind() == CandidateKind::Relayed {
+            if c.base().is_ipv4() != ip.is_ipv4() {
+                1000
+            } else {
+                0
+            }
+        } else {
+            0
+        };
+
+        let pref = counter_start - same_kind * 2 - relay_across_ip_version_punishment;
         trace!("Calculated local preference: {}", pref);
 
         c.set_local_preference(pref);
@@ -1752,6 +1767,8 @@ impl IceAgent {
 
 #[cfg(test)]
 mod test {
+    use crate::ice_::test::relay;
+
     use super::*;
     use std::net::SocketAddr;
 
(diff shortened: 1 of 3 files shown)


algesten:
I guess this opens up for that discussion we had long ago about knowing which local address was used for a relayed candidate? 🤔

thomaseizinger:
I was originally using `base` which I thought was quite clean! I didn't really think of introducing something like `source_addr` to fix what should go into `Transmit`, nicely done!

Should the priority calculation within the candidate also be patched?

src/ice/candidate.rs
```diff
@@ -205,7 +205,11 @@ impl Candidate {
     ///
     /// Relayed candidates are server sockets relaying traffic to a local socket.
     /// Allocate a TURN addr to use as a local candidate.
-    pub fn relayed(addr: SocketAddr, proto: impl TryInto<Protocol>) -> Result<Self, IceError> {
+    pub fn relayed(
+        addr: SocketAddr,
+        base: SocketAddr,
```
thomaseizinger, line 210: I think this function should document, what `base` and `addr` are.

src/ice/agent.rs
```diff
@@ -2178,4 +2195,36 @@ mod test {
...
+    fn relayed_candidates_across_ip_versions_have_lower_priority() {
+        let mut agent = IceAgent::new();
+
+        let relay_ipv4_ipv4 = relay("1.1.1.1:5000", "2.2.2.2:5000", "udp");
+        let relay_ipv4_ipv6 = relay("1.1.1.1:5001", "[::1]:5000", "udp");
+        let relay_ipv6_ipv4 = relay("[::1]:5000", "1.1.1.1:5000", "udp");
+        let relay_ipv6_ipv6 = relay("[::1]:5001", "[::2]:5000", "udp");
+
+        agent.add_local_candidate(relay_ipv4_ipv4.clone());
+        agent.add_local_candidate(relay_ipv6_ipv6.clone());
```
thomaseizinger, line 2209: Something worth noting is that the order of these candidates here is important to make the test pass due to priority ordering performed in `add_local_candidate`. So we need to add this IPv6 candidate first before the other one to ensure this priority is higher than the IPv4 <> IPv4 one.

I couldn't find a way of working around that.

src/ice/agent.rs
```diff
@@ -1752,6 +1767,8 @@ impl IceAgent {
 
 #[cfg(test)]
 mod test {
+    use crate::ice_::test::relay;
```
thomaseizinger, line 1770: This import annoyed me which is why I ended up moving the test to `mod.rs`. YMMV

src/ice/candidate.rs
algesten: check

src/ice/agent.rs
algesten: I just tried to add them in another order and it still works.

src/ice/agent.rs
algesten: Right. I just used LLM to fix it. I didn't even notice :)

algesten:
> Should the priority calculation within the candidate also be patched?

I decided to not do that. The `local_priority` is about _the local priority_ meaning the context of how this candidate is used. That the candidate has some default is not relevant. I'm half considering just panicking if we want the prio without first setting the `local_priority`. Nothing in str0m relies on this default.

algesten:
> Did we have a discussion on this? What I do remember is that we (at Firezone) at some point talked about optimising for candidates on the same relay to avoid double-hops. That problem has since gone away for us because we now always only pick a single relay per connection.

I think the problem was along the lines of having to iterate through the potential candidates to figure out which was actually used when looking at `Transmit`.

src/ice/agent.rs
thomaseizinger: Interesting, why does this suddenly work? I need to check again.

thomaseizinger:
> > Did we have a discussion on this? What I do remember is that we (at Firezone) at some point talked about optimising for candidates on the same relay to avoid double-hops. That problem has since gone away for us because we now always only pick a single relay per connection.
> 
> I think the problem was along the lines of having to iterate through the potential candidates to figure out which was actually used when looking at `Transmit`.

Ah yes. We still need to do that. We just ask each `Allocation` now: https://github.com/firezone/firezone/blob/a3c221d7c28a25f6c33dc8c7997e90d584435694/rust/connlib/snownet/src/node.rs#L1875-L1906

thomaseizinger:
> > Should the priority calculation within the candidate also be patched?
> 
> I decided to not do that. The `local_priority` is about _the local priority_ meaning the context of how this candidate is used. That the candidate has some default is not relevant. I'm half considering just panicking if we want the prio without first setting the `local_priority`. Nothing in str0m relies on this default.

I found this override a bit confusing. Could we unify this by making the default priority the algorithm that is used inside the `IceAgent` but have an additional offset that is applied to it, i.e. only compute what we are deducting? Essentially, this would mean we only need to pass in, how many other candidates of the same kind there are. Perhaps this could even be a parameter that you always have to pass to the priority calculation? `Candidate` being so stateful is a bit odd.

I don't mind either way though, also happy to go with the current model.

src/ice/agent.rs
```diff
@@ -2178,4 +2195,36 @@ mod test {
 
         buf
     }
+
+    #[test]
+    fn relayed_candidates_across_ip_versions_have_lower_priority() {
```
thomaseizinger, line 2200: I think I want to change this test to actually use two `IceAgent`s and see what ends up getting nominated. That seems like a more robust test to me because it doesn't rely so much on what the internals look like.

thomaseizinger (approved):
Works for me, thank you for iterating on it.

I am happy to merge this as is, I can iterate on the tests in a separate PR or push it here, whichever you prefer.

algesten:
Let's merge. You approve and we merge it.

thomaseizinger (approved):
Let's go!

thomaseizinger:
I did check the candidate ordering in the test and it will still fail if you add the IPv4 <> IPv4 candidate before all others because the first IPv4 <> IPv6 candidate will cause the IPv6 <> IPv6 candidate to be ranked lower and therefore have a lower priority than the IPv4 <> IPv4 one.

PR #643, opened 2025-04-19
Avoid overwriting acked SCTP association

### LINKED ISSUE #642: Multiple association on late packets on datachannel setup.
I put it here for placeholder as I never get around to do a fix for it.

My memory is vague but here is what I remember.

We noticed in slow or unreliable networks from browsers (use 3g or slow 4g) settings.

That the browser resends a new associations and str0m now overrides the previously acked association.

Webrtc should use only 1 association so our private hack looks like this.

https://github.com/pthatcher/str0m/commit/2a84703b3e714c180ae66088a9cb92874d033e5b

algesten:
Close #642

### DIFF
--- a/src/sctp/mod.rs
+++ b/src/sctp/mod.rs
@@ -382,6 +382,12 @@ impl RtcSctp {
 
         match event {
             DatagramEvent::NewAssociation(a) => {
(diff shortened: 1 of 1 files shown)


xnorpx (approved):
lgtm!

PR #644, opened 2025-04-19
refactor(ICE): improve same-IP-version preference test

thomaseizinger:
Follow-up to #640. Doing so uncovered a bug where we were still using `base` instead of `source_addr`.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1767,8 +1767,6 @@ impl IceAgent {
 
 #[cfg(test)]
 mod test {
-    use crate::ice_::test::relay;
-
     use super::*;
     use std::net::SocketAddr;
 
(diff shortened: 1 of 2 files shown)


thomaseizinger:
I looked for other usages of `base` and also found https://github.com/algesten/str0m/blob/6d9e5c69f9b793158a4aeaa7b4c1478502dcffbc/src/ice/agent.rs#L1591.

Do we need to change that too? I haven't thought about peer-reflexive candidates in detail.

algesten (approved):

algesten:
> I looked for other usages of `base` and also found https://github.com/algesten/str0m/blob/6d9e5c69f9b793158a4aeaa7b4c1478502dcffbc/src/ice/agent.rs#L1591
> 
> 
> 
> Do we need to change that too? I haven't thought about peer-reflexive candidates in detail.

i don't think so. but i haven't thought deeply about it.

algesten:
> I looked for other usages of `base` and also found https://github.com/algesten/str0m/blob/6d9e5c69f9b793158a4aeaa7b4c1478502dcffbc/src/ice/agent.rs#L1591
> 
> 
> 
> Do we need to change that too? I haven't thought about peer-reflexive candidates in detail.

i don't think so. but i haven't thought deeply about it.

algesten (approved):

PR #645, opened 2025-04-20
Add start of talkspurt for sample writer

xnorpx:
This is a proposal for implementing "start of talkspurt" support in the sample writer for use with DTX (Discontinuous Transmission) or silence suppression.

**Usage:**  
When running a client with str0m and using Opus with DTX enabled, the `start_of_talkspurt` (marker bit) should be set after periods of DTX. This indicates to the receiver that the next packet marks the beginning of a new talkspurt.

This signaling is important for jitter buffers, as it allows them to detect the end of silence periods and adjust their size accordingly.

### DIFF
--- a/src/packet/payload.rs
+++ b/src/packet/payload.rs
@@ -32,6 +32,7 @@ impl Payloader {
             wallclock,
             rtp_time,
             data,
+            start_of_talk_spurt,
             ext_vals,
             ..
         } = to_payload;
@@ -43,7 +44,8 @@ impl Payloader {
             let last = idx == len - 1;
 
             let previous_data = stream.last_packet();
-            let marker = self.pack.is_marker(data.as_slice(), previous_data, last);
+            let marker = self.pack.is_marker(data.as_slice(), previous_data, last)
+                || (is_audio && start_of_talk_spurt);
 
             let seq_no = stream.next_seq_no();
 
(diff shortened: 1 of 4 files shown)


xnorpx:
For #125

src/packet/payload.rs
```diff
@@ -43,7 +44,8 @@ impl Payloader {
             let last = idx == len - 1;
 
             let previous_data = stream.last_packet();
-            let marker = self.pack.is_marker(data.as_slice(), previous_data, last);
+            let marker = self.pack.is_marker(data.as_slice(), previous_data, last)
+                || (is_audio && start_of_talk_spurt);
```
algesten, line 48: Is using the marker for `start_of_talk_spurt` something that is the same across all audio codecs?

xnorpx: Yes I believe we should use start of talkspurt for all audio codecs.

One could argue that you can add logic to figure out when to add a marker bit for other codecs, if str0m notices a silent period. (timestamp gap, but no seq number gap or comfort noise payload types)

But imho I think it's fine leaving it to the caller for all codecs.

References that I found:
https://datatracker.ietf.org/doc/html/rfc3551#section-4.1
https://datatracker.ietf.org/doc/html/rfc3389#section-5.1

algesten: cool

algesten (approved):

PR #647, opened 2025-04-28
Fix off-by-one in SDES word size calculation

kc5nra:
When calculating the SDES word size, the `SdesType::END` written in `write_to` isn't considered in the `word_size()` method, leading to an assertion when generating the rtcp fb.

This adds the missing byte and a test that checks all possible boundary conditions for a cname (1-255)


algesten (approved):

algesten:
Thanks!

PR #649, opened 2025-05-02
fix(ice): bust cached timeout values on timing config change

thomaseizinger:
In Firezone, we dynamically update the timing config of the `IceAgent` to move into a "low-power" mode when we are idling. As soon as we see packets that we need to tunnel, we reset the timing config to ensure we detect connectivity problems in a timely manner.

At present, str0m caches when it needs to next send a binding request and these caches are not cleared when the timing config changes. As such, it may take up to how ever long MAX RTO was set to before str0m actually sends another STUN binding.

### DIFF
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -854,6 +854,62 @@ mod test {
         }));
     }
 
+    #[test]
+    fn changed_timing_config_takes_effect_immediately() {
+        let _guard = tracing_subscriber::fmt()
+            .with_env_filter("trace")
+            .with_test_writer()
+            .set_default();
+
+        const IDLE_RTO: Duration = Duration::from_secs(60);
+        const NORMAL_RTO: Duration = Duration::from_secs(3);
+
+        let mut a1 = TestAgent::new(info_span!("L"));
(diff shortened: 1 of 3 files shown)


src/ice/mod.rs
```diff
@@ -854,6 +854,59 @@ mod test {
...
+        }
+
+        let now = a1.poll_timeout().unwrap() - IDLE_RTO;
+
+        a1.set_initial_stun_rto(NORMAL_RTO);
+        a1.set_max_stun_rto(NORMAL_RTO);
+
+        let timeout_after = a1.poll_timeout().unwrap();
+
+        assert_eq!(timeout_after, now + NORMAL_RTO);
```
thomaseizinger, line 907: I wasn't sure what the best assertion here is. Calculating `now` based on the timeout and IDLE_RTO is "insider" knowledge about the ICE agent.

An alternative to this would be to calculate the difference between the two `poll_timeout` calls. That has to be the difference between the two RTOs.

algesten: I suppose it works until it doesn't. Make a code comment about it being dodgy so when it breaks we don't have to scratch our heads.

algesten (approved):

PR #650, opened 2025-05-02
refactor(ice): return `Candidate` from `add_local_candidate`

thomaseizinger:
At present, `IceAgent::add_local_candidate` consumes the `Candidate` and returns a `bool` whether the candidate was accepted by the `IceAgent`. In most applications, one needs to test for this boolean to know, whether the candidate should be signaled to the remote party. For example, a redundant candidate doesn't need to be signaled to the remote.

The current API is sub-optimal for this usecase because one always needs to `clone` the candidate in order to have it available to sending it via the signalling layer. We therefore refactor this API to return an Option`` instead with a reference to the candidate if it got accepted.

In addition, as of #640, str0m has a feature where it computes the local preference of a `Candidate` such that relay candidates allocated via the same IP version are preferred over others: An IPv4 relay candidate allocated over an IPv4 network is preferred over the IPv6 relay candidate allocated on the same network. By returning a reference to the accepted candidate from `add_local_candidate`, this local preference is preserved and reflected accordingly in the priority communicated to the remote.

As a result, the controlling agent (even if that isn't str0m), will prefer such a candidate and nominate a network path that doesn't need any IP stack translation on the relay. Prior to this change, this would only work for the relay candidates of the controlling agent as local preferences are only computed once a candidate is added to an agent.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1824,10 +1824,18 @@ mod test {
     fn local_preference_host() {
         let mut agent = IceAgent::new();
 
-        agent.add_local_candidate(Candidate::host(ipv4_1(), "udp").unwrap());
-        agent.add_local_candidate(Candidate::host(ipv6_1(), "udp").unwrap());
-        agent.add_local_candidate(Candidate::host(ipv6_2(), "udp").unwrap());
-        agent.add_local_candidate(Candidate::host(ipv4_2(), "udp").unwrap());
+        agent
+            .add_local_candidate(Candidate::host(ipv4_1(), "udp").unwrap())
+            .unwrap();
+        agent
+            .add_local_candidate(Candidate::host(ipv6_1(), "udp").unwrap())
+            .unwrap();
+        agent
+            .add_local_candidate(Candidate::host(ipv6_2(), "udp").unwrap())
+            .unwrap();
+        agent
+            .add_local_candidate(Candidate::host(ipv4_2(), "udp").unwrap())
+            .unwrap();
 
         let v: Vec<_> = agent
             .local_candidates
@@ -1859,7 +1867,7 @@ mod test {
         let host = Candidate::host(ipv4_1(), "udp").unwrap();
         let srflx = Candidate::server_reflexive(ipv4_1(), ipv4_1(), "udp").unwrap();
 
-        agent.add_local_candidate(host.clone());
+        agent.add_local_candidate(host.clone()).unwrap();
         let invalidated = agent.invalidate_candidate(&srflx);
         assert!(!invalidated);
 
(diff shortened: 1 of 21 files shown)


src/ice/agent.rs
```diff
@@ -466,17 +467,35 @@ impl IceAgent {
...
+    /// match agent
+    ///     .add_local_candidate(Candidate::host("1.1.1.1:1000".parse().unwrap(), "udp").unwrap())
+    ///     .map_err(|e| *e) // Dereference boxed error to allow pattern matching.
+    /// {
+    ///     Ok(_) => {}
+    ///     Err(IceError::RedundantCandidate { .. }) => {}
+    ///     Err(IceError::NonHostCandidateInIceLite(..)) => {}
+    /// }
+    /// ```
+    pub fn add_local_candidate(&mut self, mut c: Candidate) -> Result<&Candidate, Box<IceError>> {
```
thomaseizinger, line 492: Having to `Box` the error here is not ideal. Clippy was complaining that it is a very large error variant (168 bytes). Boxing it was the easiest to fix that as otherwise, I'd have to put an `#[allow]` on all call-sites that return `IceError` and `RtcError` (which embeds an `IceError`).

I added a doc-test to explain how to still pattern match on it. Open to other solutions!

algesten: I'd hugely prefer `Option<&Candidate>`

algesten (commented):
Looks good! Though I don't like `Result` here. 

Adding redundant candidates is not an error, it's expected. The idea is that you should be able to continuously add candidates that you discover without needing to remember which are added already.

It silently discards any that are not relevant.

thomaseizinger:
> Looks good! Though I don't like `Result` here.
> 
> Adding redundant candidates is not an error, it's expected. The idea is that you should be able to continuously add candidates that you discover without needing to remember which are added already.
> 
> It silently discards any that are not relevant.

Fair enough. For the record though, I don't think that just because something return an error means that something fatal happened. I see `Result` as a way of expressing "did this operation succeed" and when a candidate is not added, then it didn't succeed. What `Result` opens up though is the flexibility of allowing the host application to decide e.g. which level they want to log this at. With `Option`, this information is lost. So in a way, `Result` is just a richer API.

Perhaps we should turn `Result` into `Option` on the `Rtc` level?

I don't mind either way though, happy to go with `Option`!

algesten:
> I don't mind either way though, happy to go with `Option`!

With `Result` you have to handle it. And for this type, you really don't. If we need to convey meaning I'd prefer to return an enum, but we don't have any other meaning right now than "was/nt added",

thomaseizinger:
@algesten I figured that the `Result` might be controversial so I dropped that commit. Everything else is the same (and now rebased on top of `main`).

algesten (approved):

algesten:
Looks good! Thanks!

PR #652, opened 2025-05-04
Reformat doc

algesten:
- **Github action to check line lengths**
- **Reformat doc**


algesten:
@k0nserv Instead of passive-aggressively complain about my very special ideas around formatting, I decided to channel my Inner Hugo and automate the checks instead.

k0nserv (approved):
For Rust code I think it'd be better to configure Rustfmt and then use `cargo fmt --check` as the CI pass, this way contributors don't have to think too much about the rules since rustfmt will just enforce them.

There's [`max_width`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#max_width) for line lengths and then a bunch of other options that might be able to express the preference for stuff like use statements split over several lines

thomaseizinger:
> For Rust code I think it'd be better to configure Rustfmt and then use `cargo fmt --check` as the CI pass, this way contributors don't have to think too much about the rules since rustfmt will just enforce them.
> 
> There's [`max_width`](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md?rgh-link-date=2025-05-04T11%3A07%3A40Z#max_width) for line lengths and then a bunch of other options that might be able to express the preference for stuff like use statements not being split over several lines

Seconding this. It would be nice if we could achieve the desired formatting by configuring `rustfmt`. There is a neat "trick" where you can use unstable formatting features on stable rustfmt if you pass the config on the commandline. Bundled with a cargo-alias, in a repository-specific `.cargo/config.toml`, this can then be easily used by contributors.

algesten:
I should have pointed out that I already looked at rustfmt, and I don't believe it formats doc. `max_width` defaults to 100 which is less than the 110 the script is enforcing.

I don't see any way we can configure the tools here.

algesten:
And there is no option to avoid multi-line use apart from `Item` (making each import on a separate line like Java) – which is nightly and not exactly what I want.

PR #653, opened 2025-05-09
Fix vp8 depacketize for small packets

LWCoder:
# Description

I'm successfully using a GStreamer pipeline to send a VP8 encoded test pattern into str0m, but occasionally I'm seeing a `PacketError::ErrShortPacket` back from `Vp8Depacketizer.depacketize()`. Upon tracing the issue, I determined that the cause is a total packet length check at the top of the method, which I believe is incorrect.

According to the [VP8 Payload Descriptor documentation](https://datatracker.ietf.org/doc/html/rfc7741#section-4.2), there are two variations of this payload descriptor, with the more compact version only requiring a single octet. I regularly see valid packets in my tests with a length of less than four, which result in corrupted video when returned as an error instead of being processed.

I've recorded a short video demonstration of what I'm seeing with an example of the current code version with extra logging, and an example after the change has been made. In both cases, the VP8 encoder only generates keyframes on request, and both tests run for more than 5 minutes: https://www.youtube.com/watch?v=Po4v8QGKKcM

# Changes Made

The primary change is the removal of the packet length check. The rest of the code below will naturally detect a short packet if that were to occur.

Additionally, I fixed an issue with the payload descriptor comment, as the last row was offset incorrectly, and included the URL to the documentation mirroring the "VP8 Payload Header" comment found later in the code.

### DIFF
--- a/src/packet/vp8.rs
+++ b/src/packet/vp8.rs
@@ -182,9 +182,9 @@ impl Depacketizer for Vp8Depacketizer {
         extra: &mut CodecExtra,
     ) -> Result<(), PacketError> {
         let payload_len = packet.len();
-        if payload_len < 4 {
-            return Err(PacketError::ErrShortPacket);
-        }
+        // VP8 Payload Descriptor
+        // https://datatracker.ietf.org/doc/html/rfc7741#section-4.2
+        //
         //    0 1 2 3 4 5 6 7                      0 1 2 3 4 5 6 7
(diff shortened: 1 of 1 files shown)


algesten:
Thanks!

@davibe this is your area. Does this look right?

LWCoder:
Apologies, it didn't occur to me to run the tests after this change.

algesten:
No worries!

davibe (approved):

Issue #635, opened 2025-04-08
`handle_timeout()`: unlimited feedback buffer growth


alnyan:
If left to run for a while, I started observing out-of-memory crashes in my SFU implementation. I've decided to run heap profiling on it and found out that the stack trace related to most memory consumtion (and which keeps growing over the program's lifespan) is:

* `-- Frames for the SFU code itself --`
* `str0m::Rtc::handle_input`
* `str0m::Rtc::do_handle_timeout`
* `str0m::session::Session::handle_timeout`
* `str0m::streams::Streams::handle_timeout`
* `str0m::streams::send::StreamTx::create_sr_and_update`
* `-- VecDeque impl + some malloc/profiling frames --`

In the code, I see str0m push SenderReport + SourceDescription feeback items if a stream in question is in the `mids_to_report` list, which is built in the `streams::Streams::handle_timeout()` function.

For some context, I am polling the outputs in a loop, where I call `rtc.poll_output()` for each of the streams, and if a `Timeout` is returned, I sleep and call `rtc.handle_input(Instant::now())` to advance the internal timer of the RTC instance, or break out of the loop with the data if `Transmit` or `Event` is returned.

I am not yet sure if this is an actual issue with str0m itself or my specific usage of it, but maybe capping the amount of SenderReport feedback items would be helpful?

alnyan:
Here is a flamegraph of my attempts at profiling the memory leak, I clipped it to the frame which seems to be "leaking" memory:
![Image](https://github.com/user-attachments/assets/5f8ed70a-1229-454f-ae00-374fff1cf11f)

algesten:
Hi @alnyan!

Thanks for reporting this. It's very strange.

alnyan:
I cannot reproduce the issue by running the code locally, unfortunately. I first observed this problem when running a deployment with ~800 streams, of which most are paused and ~10 are active (streaming actual data) at a time. I'll try to reproduce it tomorrow and will let you know if I come up with any results.

alnyan:
I've managed to reproduce the issue: the feedback items keep piling up if either:
 
* Transmit of `Output::Transmit` fails, which I assume is unlikely, because I'm using UDP sockets as a transport
* Reception of input UDP datagram fails (e.g. remote client loses connection or goes offline for some reason)

[...]

PR #654, opened 2025-05-15
Drop feedback until ICE is established

OxleyS:
Potential solution to https://github.com/algesten/str0m/issues/635.

Quite simply, it just clears any generated feedback if ICE has not yet been established. This prevents it from buffering indefinitely if ICE just never establishes at all.

I thought this might be a simpler solution than trying to have extra logic to gate generating that feedback in the first place.

### DIFF
--- a/tests/no-pre-ice-feedback.rs
+++ b/tests/no-pre-ice-feedback.rs
@@ -1,0 +1,72 @@
+use std::net::Ipv4Addr;
+use std::time::Duration;
+
+use str0m::media::{Direction, MediaKind};
+use str0m::rtp::RawPacket;
+use str0m::{Rtc, RtcError};
+use tracing::info_span;
+
+mod common;
+use common::{init_crypto_default, init_log, negotiate, progress, TestRtc};
+
+#[test]
+pub fn no_pre_ice_feedback() -> Result<(), RtcError> {
+    init_log();
+    init_crypto_default();
+
+    let l_rtc = Rtc::builder().enable_raw_packets(true).build();
+    let r_rtc = Rtc::builder().enable_raw_packets(true).build();
+
+    let mut l = TestRtc::new_with_rtc(info_span!("L"), l_rtc);
+    let mut r = TestRtc::new_with_rtc(info_span!("R"), r_rtc);
+
+    l.add_host_candidate((Ipv4Addr::new(1, 1, 1, 1), 1000).into());
+    r.add_host_candidate((Ipv4Addr::new(2, 2, 2, 2), 2000).into());
+
+    negotiate(&mut l, &mut r, |change| {
+        change.add_media(MediaKind::Video, Direction::SendOnly, None, None, None)
+    });
+
+    // Before ICE is established, introduce a large time delta that spans at least one sender/receiver report
+    // interval
+    let mut t = l.last;
+    l.handle_input(str0m::Input::Timeout(t))?;
+    t += Duration::from_secs(10);
+    l.handle_input(str0m::Input::Timeout(t))?;
+
+    /// Some things are gated not only on DTLS being connected, i.e. `Rtc.is_connected()`, but on that
+    /// connected event being propagated internally. This includes the outputting of raw packets sometimes.
+    fn is_fully_connected(rtc: &TestRtc) -> bool {
+        rtc.events
+            .iter()
+            .any(|(_, e)| matches!(e, str0m::Event::Connected))
+    }
+
+    loop {
+        if is_fully_connected(&l) && is_fully_connected(&r) {
+            break;
+        }
+
+        progress(&mut l, &mut r)?;
+    }
+
+    for (_, event) in l.events.iter() {
+        if let Some(RawPacket::RtcpTx(tx)) = event.as_raw_packet() {
+            panic!(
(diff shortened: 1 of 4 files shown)


thomaseizinger:
> I thought this might be a simpler solution than trying to have extra logic to gate generating that feedback in the first place.

Would it be that difficult? It can only be sending feedback right? That happens in the `Session::handle_timeout` function on line 236:

https://github.com/algesten/str0m/blob/9b0125388b4c54775da392c8e46e7e5d526f636c/src/session.rs#L236

So we would "only" need to pass the information about whether we have an ICE connection into that `handle_timeout` function. But yeah, it is a bit brittle.

Alternatively, the call to `Session::poll_datagram` could be split off from the call to `Dtls::poll_datagram` and only the latter would be conditional on having an ICE connection. We could unconditionally poll for packets from the `Session` and simply drop them if we don't have an ICE connection. Maybe the same should be done for DTLS as well? I guess that could generally be a good approach to avoid buffers from building up? I don't know the details though, maybe that would lead to timeouts?

algesten:
@OxleyS looks good! If you fix the test error we can merge. It's the rust doc that is too long on row 660.

thomaseizinger:
Would it be possible to add a unit test for this? Like, create two agents but delay ICE yet trigger feedback being generated, finish ICE, assert that no feedback packets are being emitted.

OxleyS:
Thanks for the feedback, I'll work through all this on Monday!

@algesten Is that a Clippy check? My tools didn't give me any feedback on line length, although maybe something was wonky with my setup.

algesten:
@OxleyS i don't think i can configure clippy and/or rustfmt to warn on this, so I made my own github CI job. But I'd be super happy if it was possible.

OxleyS:
I've amended the commit to fix the line length issue and to rename the method to `clear_feedback`. I didn't know what the length limit was until I dug into the PR adding the CI job. Should we add `"editor.rulers": [110]` to `.vscode/settings.json`, to give a little more hinting to VSCode users?

I've also added a test that I hope adequately covers this issue. At the very least, it fails without the `clear_feedback` call and passes with it. The only part that bothers me is the `is_fully_connected()` function. Streams will not output anything, including raw packets, until SRTP keying material has been propagated. On the sender side in particular, this happens one step _after_ `dtls.is_connected()` starts returning `true`, requiring one more round of `progress()`. I don't know if this behavior is a bug or not.

> So we would "only" need to pass the information about whether we have an ICE connection into that handle_timeout function

It's not just TWCC we need to prevent, it's also things like Sender Reports. This means we'd need to pass whether we have ICE at least one level deeper than that, into `Streams::handle_timeout()` as well. Alternatively, we could gate the entire `Streams::handle_timeout()` call, but at that point I'd start to worry about accidentally gating something else out that actually should have been run.

> Alternatively, the call to Session::poll_datagram could be split off from the call to Dtls::poll_datagram and only the latter would be conditional on having an ICE connection. We could unconditionally poll for packets from the Session and simply drop them if we don't have an ICE connection.

I do like this solution as well. The one snag is that in tests, if you poll a packet from the `Session`, the corresponding raw packet is generated and output, even if you drop the regular packet. This makes our new test fail! It doesn't strictly have to be that way, I suppose, but it would require some non-trivial refactoring to fix.

> Maybe the same should be done for DTLS as well?

I could be reading the code wrong, but I think data doesn't get fed into `Rtc::dtls` anyway until DTLS is established (for which ICE is a precondition). For the memory growth I mentioned [here](https://github.com/algesten/str0m/issues/635#issuecomment-2882483239), things were being buffered before that point.

We could adopt this drop-if-no-ICE as a general rule to apply across everything, though. I could see arguments both for and against it. The obvious benefit is stopping memory bloat in a less regression-prone way, but the downside is the surprising behavior of dropping media packets that we let the user write without error. For video in particular, dropping a keyframe in this manner would be a fairly big deal.

thomaseizinger:
> we let the user write without error

Ultimately, this is the issue, isn't it? We don't have backpressure to the caller that the stream isn't ready yet to take on data.

tests/no-pre-ice-feedback.rs
```diff
@@ -0,0 +1,72 @@
...
+    // Before ICE is established, introduce a large time delta that spans at least one sender/receiver report
+    // interval
+    let mut t = l.last;
+    l.handle_input(str0m::Input::Timeout(t))?;
+    t += Duration::from_secs(10);
+    l.handle_input(str0m::Input::Timeout(t))?;
+
+    /// Some things are gated not only on DTLS being connected, i.e. `Rtc.is_connected()`, but on that
+    /// connected event being propagated internally. This includes the outputting of raw packets sometimes.
+    fn is_fully_connected(rtc: &TestRtc) -> bool {
```
thomaseizinger, line 39: I'd suggest to move this to `TestRtc`.

thomaseizinger (approved):
Looks great!

algesten:
@OxleyS 

>  Should we add "editor.rulers": [110] to .vscode/settings.json, to give a little more hinting to VSCode users?

Good idea! I'll add that in another PR.

I think this looks ready to land. We can move the `is_fully_connected` when we start using it somewhere else.

PR #655, opened 2025-05-18
fix/media event

### LINKED ISSUE #567: `Event::MediaAdded` is not fired for medias added in local offer
The `MediaAdded` event documentation states that 

> This event fires both for negotiations triggered by a remote or local offer.

Which is not the case, since it is emitted only for media lines added via remote offer, and it seems that firing these event for local offer is explicitly disabled right here: [0](https://github.com/algesten/str0m/blob/ed9b9eb9da09ba102b0ce06303e3ca7377a4a6b3/src/change/sdp.rs#L800), [1](https://github.com/algesten/str0m/blob/ed9b9eb9da09ba102b0ce06303e3ca7377a4a6b3/src/change/sdp.rs#L962-L969).

And, yeah, while this is an `apply_answer()` function it is also when local offer is being applied too, so i guess it should also cause `MediaAdded` event to fire.

So is this an expected behaviour and outdated documentation or a bug? I can prepare a PR if its the second case.

algesten:
Close #567

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Document behavior of MediaAdded #655
   * Fix growing feedback bug before ICE #654
   * Fix vp8 depacketize for small packets #653
   * Bust cache on config change #649
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -912,9 +912,13 @@ pub enum Event {
 
     // =================== Media related events ==================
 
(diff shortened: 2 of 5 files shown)


thomaseizinger (approved):

thomaseizinger:
Can't speak for the docs changes as we don't use those parts of str0m :)

OxleyS (approved):

PR #656, opened 2025-05-26
Pii newtype to redact pii data as feature

xnorpx:
**This is an opt-in feature flag and will be disabled by default to not change current behavior.**

It is important to be able to avoid logging PII (Personally Identifiable Information) because improper handling or exposure of such data can lead to privacy violations, regulatory penalties, reputational damage, and loss of customer trust.

IP is a clear example of PII, and this has been the focus of this PR. However there might be other information that we also want to put under PII such as ufrag, mids etc but that can be done in a follow up PR in an iterative manner.

- Introduced a new `Pii<T>` wrapper type in pii.rs that, when the `pii` feature is enabled, redacts sensitive values (shows `{REDACTED}` for `Display`/`Debug`).
- Updated logging throughout the codebase (notably in ICE and DTLS modules) to wrap sensitive values (such as credentials, candidates, addresses, and events) in `Pii`, ensuring they are redacted in logs when the `pii` feature is enabled.
- Added pii.rs with a test that:
  - Only runs if the `pii` feature is enabled.
  - Installs a custom tracing subscriber that panics if any IPv4 or IPv6 address is found in logs (using the `regex` crate).
  - Runs a media session and asserts that no IP addresses are leaked in logs.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -8,7 +8,7 @@ use crate::io::{Id, StunClass, StunMethod, StunTiming, DATAGRAM_MTU_WARN};
 use crate::io::{Protocol, StunPacket};
 use crate::io::{StunMessage, TransId};
 use crate::io::{Transmit, DATAGRAM_MTU};
-use crate::util::NonCryptographicRng;
+use crate::util::{NonCryptographicRng, Pii};
 
 use super::candidate::{Candidate, CandidateKind};
 use super::pair::{CandidatePair, CheckState, PairId};
@@ -318,7 +318,7 @@ impl IceAgent {
     /// Sets the local ice credentials.
     pub fn set_local_credentials(&mut self, r: IceCreds) {
         if self.local_credentials != r {
-            info!("Set local credentials: {:?}", r);
+            info!("Set local credentials: {:?}", Pii(&r));
             self.local_credentials = r;
         }
     }
@@ -400,7 +400,7 @@ impl IceAgent {
     /// Sets the remote ice credentials.
     pub fn set_remote_credentials(&mut self, r: IceCreds) {
         if self.remote_credentials.as_ref() != Some(&r) {
-            info!("Set remote credentials: {:?}", r);
+            info!("Set remote credentials: {:?}", Pii(&r));
             self.remote_credentials = Some(r);
         }
     }
@@ -603,7 +603,8 @@ impl IceAgent {
                     // The new candidate is not better than what we already got.
                     debug!(
                         "Reject redundant candidate, current: {:?} rejected: {:?}",
-                        other, c
+                        Pii(&other),
+                        Pii(&c)
                     );
                     return None;
                 }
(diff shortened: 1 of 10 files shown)


src/lib.rs
```diff
@@ -1421,7 +1421,7 @@ impl Rtc {
             },
             Output::Transmit(t) => {
                 self.peer_bytes_tx += t.contents.len() as u64;
-                trace!("OUT {:?}", t)
+                trace!("OUT {:?}", Pii(t))
```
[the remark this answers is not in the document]
xnorpx: sure

algesten:
@xnorpx looks good.

I wonder about messages on the `trace` level though. In `ureq` I decided that trace is free-for-all in terms of logging. "Wire-level" logging even. I'm interested in adopting such a stance in str0m to avoid having to have the `Pii` on every single trace message.

algesten:
For reference: https://github.com/algesten/ureq?tab=readme-ov-file#log-levels

I also got told off for having stuff on the INFO level. Apparently many people expect INFO to be fairly quiet.

xnorpx:
> For reference: https://github.com/algesten/ureq?tab=readme-ov-file#log-levels
> 
> I also got told off for having stuff on the INFO level. Apparently many people expect INFO to be fairly quiet.

We turn off str0m info logging in our production so it would align to how we use it.

thomaseizinger:
> In `ureq` I decided that trace is free-for-all in terms of logging. "Wire-level" logging even. I'm interested in adopting such a stance in str0m to avoid having to have the `Pii` on every single trace message.

FWiW, we have a similar policy in Firezone and it has served us well.

algesten:
Yeah, @xnorpx I'd like to have parity between my projects here, perhaps even increase the verbosity of `trace` to be wire level. If this PR could focus on debug level and up be great.

We should also visit the use of INFO, WARN and ERROR to be similar to ureq. That could be beyond the scope of this PR, so I can do that in a follow-up.

src/ice/agent.rs
```diff
@@ -1041,7 +1048,7 @@ impl IceAgent {
     ///
     /// Will not be used if [`IceAgent::accepts_message`] returns false.
     pub fn handle_packet(&mut self, now: Instant, packet: StunPacket) -> bool {
-        trace!("Handle receive: {:?}", &packet.message);
+        trace!("Handle receive: {:?}", packet.message);
```
algesten, line 1051: Does this need changing?

src/ice/agent.rs
```diff
@@ -1237,7 +1244,7 @@ impl IceAgent {
     pub fn poll_event(&mut self) -> Option<IceAgentEvent> {
         let x = self.events.pop_front();
         if x.is_some() {
-            trace!("Poll event: {:?}", x);
+            trace!("Poll event: {:?}", &x);
```
algesten, line 1247: And this?

wincrypto/Cargo.toml
```diff
@@ -14,4 +14,9 @@ description = "Supporting crate for str0m"
 [dependencies]
 thiserror = { version = "1.0.38" }
 tracing = "0.1.37"
-windows = { version = "0.58", features = ["Win32_Security_Cryptography", "Win32_Security_Authentication_Identity", "Win32_Security_Credentials", "Win32_System_Rpc"] }
+windows = { version = "0.58", features = [
+    "Win32_Security_Cryptography",
+    "Win32_Security_Authentication_Identity",
+    "Win32_Security_Credentials",
+    "Win32_System_Rpc",
+] }
```
algesten, line 22: Was this a deliberate change?

algesten (approved):
Looks good. Some things that look like unintended changes. I approve and when you fix this you can merge.

src/ice/agent.rs
xnorpx: ugh no, let me fix

src/ice/agent.rs
xnorpx: will fix

wincrypto/Cargo.toml
xnorpx: Yes, Martin snowflake triggered on this line.

xnorpx: Discussed offline will push this

PR #659, opened 2025-06-01
H265 support

showier-drastic:
These commits add full support for H265.

Including tests adapted from the `pion/webrtc` project. All tests passed.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -106,8 +106,6 @@ pub struct CodecSpec {
 pub enum Codec {
     Opus,
     H264,
-    // TODO show this when we support h265.
-    #[doc(hidden)]
     H265,
     Vp8,
     Vp9,
@@ -503,6 +501,7 @@ impl CodecConfig {
         let mut c = Self::empty();
         c.enable_opus(true);
 
+        c.enable_h265(true);
         c.enable_vp8(true);
         c.enable_h264(true);
         // c.add_default_av1();
@@ -635,6 +634,23 @@ impl CodecConfig {
         for p in PARAMS {
             self.add_h264(p.0.into(), Some(p.1.into()), p.2, p.3)
         }
+    }
+
+    /// Enable H265 codec.
+    pub fn enable_h265(&mut self, enabled: bool) {
+        self.params.retain(|c| c.spec.codec != Codec::H265);
+        if !enabled {
+            return;
+        }
+
+        self.add_config(
+            116.into(),
+            Some(117.into()),
+            Codec::H265,
+            Frequency::NINETY_KHZ,
+            None,
+            FormatParams::default(),
+        );
     }
 
     // TODO: AV1 depacketizer/packetizer.
--- a/src/packet/h265.rs
+++ b/src/packet/h265.rs
@@ -807,9 +807,18 @@ impl Depacketizer for H265Depacketizer {
     }
 
     /// is_partition_head checks if this is the head of a packetized nalu stream.
-    fn is_partition_head(&self, _payload: &[u8]) -> bool {
-        //TODO:
-        true
+    fn is_partition_head(&self, packet: &[u8]) -> bool {
+        if packet.len() < 3 {
+            return false;
+        }
+
+        let header = H265NALUHeader::new(packet[0], packet[1]);
+
+        if header.is_fragmentation_unit() {
+            H265FragmentationUnitHeader(packet[2]).s()
+        } else {
+            true // AP and Single NALU packets are always partition heads
+        }
     }
 
     fn is_partition_tail(&self, marker: bool, _payload: &[u8]) -> bool {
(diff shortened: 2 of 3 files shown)


src/format.rs
```diff
@@ -503,6 +501,7 @@ impl CodecConfig {
         let mut c = Self::empty();
         c.enable_opus(true);
 
+        c.enable_h265(true);
```
k0nserv, line 516: We need to be careful that we don't end up breaking SDP negotiation due to the ongoing issue of PT exhaustion and resulting collisions. Worth doing some testing here especially with differing orderings of who makes the first offer (browser or str0m) and the effect of subsequent offers with different kinds of media (cameras, screens, and audio)

showier-drastic: I am not actually familiar with how the payload types are assigned or negotiated between different partners. Could you please give some advice on how do I perform the tests that you mentioned?

src/packet/h265.rs
```diff
@@ -1,7 +1,7 @@
 #![allow(clippy::all)]
```
xnorpx, line 1: Remove these clippy suppressions and then look through the implementations and see what warnings there is, if any then figure out why they are there.

xnorpx:
I would generate PCAP files from Safari and Chrome senders with h265 payloads and run them through the depacketizer.

I would encode a couple of payloads, then write a send recv integration tests that you tests both the packetizer and depacketizer.

algesten:
@showier-drastic thank you for looking at this!

I think removing the directives and fix any potential problems would be good.

showier-drastic:
> I would generate PCAP files from Safari and Chrome senders with h265 payloads and run them through the depacketizer.
> 
> I would encode a couple of payloads, then write a send recv integration tests that you tests both the packetizer and depacketizer.

This seems to be a bit big work for me. It seems that we don't have test suites like this available? I think I'm currently unable to writing these test suites. But if we have existing test suites, I'm more than happy to run and debug them.

Anyway, I successfully run it in on my local code to send stream from a local h265 file to the browser. I can tidy this code up and contribute this to the examples.

> @showier-drastic thank you for looking at this!
> 
> I think removing the directives and fix any potential problems would be good.

Thanks, I will do this.

algesten:
> This seems to be a bit big work for me. It seems that we don't have test suites like this available? I think I'm currently unable to writing these test suites. But if we have existing test suites, I'm more than happy to run and debug them.

We do have tests like that for the other protocols. Check out [tests/data](https://github.com/algesten/str0m/tree/main/tests/data). I made the first of these pcaps using wireshark. Each protocol has `*.pcap` and a `*.txt` both are saved out from wireshark.

I believe we use `SSLKEYLOGFILE` with chrome, right @k0nserv?

k0nserv:
It's not `SSLKEYLOGFILE` for webrtc. Here's how I do it on macOS, but this can be tweaked for other operating systems

## Chrome

Run chrome as `$PATH_TO_CHROME  --enable-logging -v=3 --force-fieldtrials=WebRTC-Debugging-RtpDump/Enabled/`

Then use this script to turn it into a pcap

```bash
#!/bin/sh
function yes_or_no {
    while true; do
        read -p "$* [y/n]: " yn
        case $yn in
            [Yy]*) return 0  ;;  
            [Nn]*) return  1 ;;
        esac
    done
}
NOW=`date +"%Y-%m-%dT%H:%M"`

CHROME_LOG_PATH="${1:-$HOME/Library/Application Support/Google/Chrome Dev/chrome_debug.log}"
TEXT_2_PCAP="${2:-/Applications/Wireshark.app/Contents/MacOS/text2pcap}"
OUT_BASE_PATH="${3:-$HOME/Documents/chrome-rtp-dump}"
OUT_PATH="$OUT_BASE_PATH/rtp-dump-$NOW.pcap"

mkdir -p $OUT_BASE_PATH

grep RTP_DUMP "$CHROME_LOG_PATH" > "$OUT_BASE_PATH/rtp-dump-$NOW.log"
$TEXT_2_PCAP -D -u 5443,62132 -t %H:%M:%S.%f "$OUT_BASE_PATH/rtp-dump-$NOW.log" $OUT_PATH

echo "Open the pcap?"
yes_or_no && open $OUT_PATH
```

## Firefox

Run using

```bash
#!/usr/bin/env sh
BINARY="/Applications/Firefox\ Developer\ Edition.app/Contents/MacOS/firefox"
export MOZ_LOG=timestamp,signaling:5,jsep:5,RtpLogger:5,SCTP:5
export MOZ_LOG_FILE=/tmp/moz_logs/log
export MOZ_DISABLE_CONTENT_SANDBOX=1
mkdir -p /tmp/moz_logs
echo "Starting $BINARY"
/Applications/Firefox\ Developer\ Edition.app/Contents/MacOS/firefox
```

Then convert to pcap with 

```bash
#!/bin/sh
function yes_or_no {
    while true; do
        read -p "$* [y/n]: " yn
        case $yn in
            [Yy]*) return 0  ;;  
            [Nn]*) return  1 ;;
        esac
    done
}
NOW=`date +"%Y-%m-%dT%H-%M"`

LOG_PATH="${1:-/tmp/moz_logs/log.child-4.moz_log}"
TEXT_2_PCAP="${2:-/Applications/Wireshark.app/Contents/MacOS/text2pcap}"
OUT_BASE_PATH="${3:-$HOME/Documents/firefox-rtp-dump}"
OUT_PATH="$OUT_BASE_PATH/rtp-dump-$NOW.pcap"

mkdir -p $OUT_BASE_PATH

grep -o -E "(RTP_PACKET|RTCP_PACKET)(.*$)" $LOG_PATH | cut -d ' ' -f 2- > "$OUT_BASE_PATH/rtp-dump-$NOW.log"
$TEXT_2_PCAP -D -u 5443,62132 -t %H:%M:%S.%f "$OUT_BASE_PATH/rtp-dump-$NOW.log" $OUT_PATH

echo "Open the pcap?"
yes_or_no && open $OUT_PATH
```

src/format.rs
k0nserv: It might be tricky to test if you don't already have a system that can initiate offers from both a browser or str0m (forwarded from some other browser). Given that you try various sequences of turning on and off video media (cameras and screens).

Relevant bug for Chromium https://issues.webrtc.org/issues/41480892

algesten:
Closing due to inactivity

PR #660, opened 2025-06-01
Change error log

xnorpx:
@algesten not sure how you want this one, 

error -> debug log
error -> warn log
assert
just remove it


wincrypto/src/dtls.rs
```diff
@@ -631,12 +631,12 @@ impl Drop for Dtls {
...
             if let Some(ctx_handle) = self.security_ctx {
                 if let Err(e) = DeleteSecurityContext(&ctx_handle) {
-                    error!("DeleteSecurityContext on Drop failed: {:?}", e);
+                    debug!("DeleteSecurityContext on Drop failed: {:?}", e);
                 }
             }
             if let Some(cred_handle) = self.cred_handle {
                 if let Err(e) = FreeCredentialsHandle(&cred_handle) {
-                    error!("FreeCredentialsHandle on Drop failed: {:?}", e);
+                    debug!("FreeCredentialsHandle on Drop failed: {:?}", e);
```
algesten: So what does this error message mean?

xnorpx: Looks like the only error codes those functions can return is  SEC_E_INVALID_HANDLE which in theory should be handled by the option.

xnorpx: If the logic is correct that is, I would probably have an assert(FreeCredenti.... .is_ok())

algesten: The wincrypto is more you guys' domains. Asserts sounds good to me!

xnorpx: Changed to expect valid handle

algesten (approved):

PR #661, opened 2025-06-06
Add some remote stats from SR/RR stats into Ingress/EgressStats

efer-ms:
Fill in RemoteIngressStats, and RemoteEgressStats using the most recent RR or SR respectively.

I made the naming within the remote stats structs consistent with the local stats, so renamed "bytes_tx" to "bytes" .. since we don't differentiate that way in MediaIngressStats or MediaEgressStats.

Did not pass the ntp/rtp time values in the SR nor the last SR time + delay in RR. Not sure if we feel they're needed or not, figure they can be added when needed.

PR #662, opened 2025-06-07
First attempt at adding SR/RR stats into Ingress/EgressStats

algesten:
* First attempt at adding SR/RR stats into Ingress/EgressStats

* fix doctest

* ok... better I think

### DIFF
--- a/src/stats.rs
+++ b/src/stats.rs
@@ -4,6 +4,7 @@ use std::collections::{HashMap, VecDeque};
 use std::net::SocketAddr;
 use std::time::{Duration, Instant};
 
+use crate::rtp::SeqNo;
 use crate::rtp_::{Mid, Rid};
 use crate::Bitrate;
 use crate::{io::Protocol, rtp_::MidRid};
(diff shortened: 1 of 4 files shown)


algesten:
This is #661

algesten:
@efer-ms I have added two commits. The first is to make the RemoteEgressStat calculation clearer to a reader. The second commit is to address the extended_seq calculation of the RR max_seq. Reading the code I realized we had two problems here:

1. We were already extended it in an incorrect way (relying on loss reports), so the new code was doubling up on something already happening. However, relying on loss reports seems dodgy – what if there are none?
2. The new code relied on `Sender.seq_no`, and there's a crucial clue in the doc to the field `"If we are doing seq_no ourselves (when writing sample mode)"`. This field is only in use for the "Media API". If we're doing RTP Mode, the `SeqNo` is provided in the incoming RTP packet.

To solve this properly, I introduced a new `last_sent_seq_no`, which we update when we poll the next packet to send, and it's a regular packet (not resend or padding). This should be correct for both Media API and RTP Mode.

efer-ms:
Looks good to me! Thank you for improving it!

xnorpx (approved):
lgtm

PR #663, opened 2025-06-11
fix(ice): don't reject relay candidate as redundant with host

thomaseizinger:
In #640 and #644, we broke ICE for network setups where both peers are behind symmetric NAT because relay candidates now have a base and when we check for redundant candidate pairs, we only compare the base. Instead, I believe we need to compare the `source_addr` here as that is what we use to send packets from.

The notion of "base" is getting pretty muddied here but I also can't think of a much better model here that doesn't involve storing yet-another `SocketAddr` in the `Candidate`.

Co-authored-by: Antoine Labarussias <antoinelabarussias@gmail.com>

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -779,7 +779,7 @@ impl IceAgent {
                     let check_local = check.local_candidate(&self.local_candidates);
                     let check_remote = check.remote_candidate(&self.remote_candidates);
 
-                    let redundant = local.base() == check_local.base()
+                    let redundant = local.source_addr() == check_local.source_addr()
                         && remote.addr() == check_remote.addr()
                         && local.proto() == check_local.proto()
                         && remote.proto() == check_remote.proto();
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -265,6 +265,41 @@ mod test {
         );
     }
 
+    #[test]
+    pub fn relay_relay() {
+        let _guard = tracing_subscriber::fmt()
+            .with_test_writer()
+            .with_env_filter("trace")
+            .set_default();
+
+        let mut a1 = TestAgent::new(info_span!("L"));
+        let mut a2 = TestAgent::new(info_span!("R"));
+
+        // Add unreachable host candidates (simulating symmetric NAT)
+        a2.add_remote_candidate(a1.add_host_candidate("1.1.1.1:9999"));
+        a1.add_remote_candidate(a2.add_host_candidate("2.2.2.2:9999"));
+
+        // Add relay candidates
+        let r1 = a1
+            .add_local_candidate(relay("5.5.5.5:1000", "1.1.1.1:9999", "udp"))
+            .unwrap();
+        a2.add_remote_candidate(r1.clone());
+        let r2 = a2
+            .add_local_candidate(relay("6.6.6.6:1000", "2.2.2.2:9999", "udp"))
+            .unwrap();
+        a1.add_remote_candidate(r2.clone());
+
+        a1.set_controlling(true);
+        a2.set_controlling(false);
+
+        loop {
+            if a1.state().is_connected() && a2.state().is_connected() {
+                break;
+            }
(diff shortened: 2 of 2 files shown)


thomaseizinger:
> Co-authored-by: Antoine Labarussias [antoinelabarussias@gmail.com](mailto:antoinelabarussias@gmail.com)

@Intuinewin I've added you as the co-author because you've discovered the bug and essentially also provided me with the fix. Let me know if you are happy with this attribution.

@algesten Please include the `Co-authored-by` trailer in the commit message when you squash-merge this to ensure proper attribution.

algesten:
@thomaseizinger Maybe I'm missing something, but this seems to directly [contradict spec](https://www.rfc-editor.org/rfc/rfc8445#section-6.1.2.4). 

> Two candidate pairs are redundant if their local candidates have the same base 
> and their remote candidates are identical.

The way I see it, "base" is a normalization that makes it possible to compare say a host and a relay candidate, because the host will be a better choice.

thomaseizinger:
> The way I see it, "base" is a normalization that makes it possible to compare say a host and a relay candidate, because the host will be a better choice.

But the host candidate is unusable if the peer is behind symmetric NAT. I am open to other solutions but in the current form, ICE is broken for network setups with two symmetric NAT (see the failing the test).

thomaseizinger:
> Two candidate pairs are redundant if their local candidates have the same base
> and their remote candidates are identical.

I am aware of this. Maybe we are wrong about what the base of a relay candidate is?

algesten:
> I am aware of this. Maybe we are wrong about what the base of a relay candidate is?

Yeah, that could be it.

>      The base of a server reflexive candidate is the host candidate
>      from which it was derived.  A host candidate is also said to have
>      a base, equal to that candidate itself.  Similarly, the base of a
>      relayed candidate is that candidate itself.

So

* host - base: host
* relay - base: relay
* peer/server reflexive – base: host

?

thomaseizinger:
That means in order to support #640 and #644, we need store an additional field?

thomaseizinger:
> That means in order to support #640 and #644, we need store an additional field?

I am overall not very fond of the current design where `Candidate` is being mutated whilst it is stored inside the `IceAgent`.

Would you be open to a refactor here that aims to reduce the mutability? One thing we could try is creating an explicit `LocalCandidate` and `RemoteCandidate`. Perhaps that would open up some possibilities?

algesten:
> I am overall not very fond of the current design where `Candidate` is being mutated whilst it is stored inside the `IceAgent`.
> 
> Would you be open to a refactor here that aims to reduce the mutability? One thing we could try is creating an explicit `LocalCandidate` and `RemoteCandidate`. Perhaps that would open up some possibilities?

We can explore a refactor here. I think the public API surface of `Candidate` is pretty good as it is. I don't particularly want to split that in two against the user (this could have weird consequences for SDP serialization). I have no problem exploring an immutable version.

I think the ipv6-ipv6 preference is a secondary concern compared to relayed candidates being completely broken. If keeping this function requires us to add additional fields to keep track of the local address at all times (and public API to support that), it's a bit meh. Is it possible to just change the local preference values before adding the candidates?

thomaseizinger:
> If keeping this function requires us to add additional fields to keep track of the local address at all times (and public API to support that), it's a bit meh.

I need to sit down and iterate on a few designs to say for sure but my thinking was that we can perhaps avoid the additional field if we can split `Candidate` such that the priority calculation can be moved into the constructor or even be completely external.

For example, if we strip the type to just serialise to and from SDP then we could consider a constructor where you just set the priority. Maybe it can remain crate-private though and all we need as public API are a type (`NewCandidate`, `LocalCandidate`, etc?) that hold the information about gathered candidates (just the address and type) to be passed to `add_local_candidate`. As part of adding them, their priority can be calculated and we return an immutable `Candidate` that can be serialised.

thomaseizinger:
Closed in favor of #665. Will open a new PR with a refactoring later.

PR #664, opened 2025-06-11
Simplified implementation of SDP ser/de.

mstyura:
If I'm not missing anything crucial the proposed implementation should be must easier to understand while not changing the properties of original implementation.


algesten (approved):

algesten:
Thanks!

PR #665, opened 2025-06-13
Remove base from relay candidates (follow spec)

algesten:
Relates to #663 

The ICE spec says that relay candidates should have their base set to the same as the relay address itself. This makes it impossible to preserve the behavior where the agent prefers to maintain IPv4/IPv6 between local address and relay. This needs a solution, but we are looking at a bigger refactor (see #663)

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -209,14 +209,8 @@ impl Candidate {
     ///
     /// * `addr` - The TURN server's allocated address that will be used for relaying traffic.
     ///            This is the address that will be used for communication with the peer.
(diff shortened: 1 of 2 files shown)


algesten:
@thomaseizinger this PR is to get the public API of `Candidate::relay` in line with spec. It of course breaks the ipv4-ipv4, preference, but let's tackle that in follow-up refactoring.

thomaseizinger:
> This needs a solution, but we are looking at a bigger refactor (see #663)

I'll start on this early next week, currently travelling without a computer.

PR #667, opened 2025-06-16
feat: introduce `created_via` parameter to `Candidate::relayed`

thomaseizinger:
This PR is a follow-up to #665 that re-enables the feature of computing the preference of a relay candidate such that we stay within the same IP version. To achieve this without adding additional fields to `Candidate`, almost all of the preference calculation is moved into `Candidate` except for the adjustment based on the number of already added candidates of the same type. This information is only present in the `IceAgent`.

Instead of allowing to simply set the local preference, `Candidate` now exposes a narrower API that accepts a delta that is applied to the preference.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1,6 +1,5 @@
 use std::collections::{HashSet, VecDeque};
 use std::net::SocketAddr;
-use std::ops::SubAssign;
 use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
@@ -524,23 +523,9 @@ impl IceAgent {
         // ================
         //
         // The above presupposes that we know all the candidates when we start
-        // the ice agent. That doesn't work for us, so we deliberately do not
-        // follow spec. We assign the following intervals for the different
-        // types of candidates:
-        //
-        // 0     - 16384 => relay
-        // 16384 - 32768 => srflx
-        // 32768 - 49152 => prflx
-        // 49152 - 65536 => host
-        //
-        // And furthermore we subdivide these to interleave IPv6 with IPv4
-        // so that odd numbers are ipv6 and even are ipv4.
-        //
-        // For host candidates this means:
-        // 65535 - first ipv6
-        // 65534 - first ipv4
-        // 65533 - second ipv6
-        // 65432 - second ipv4
+        // the ice agent. That doesn't work for us. The best we can do is adjust
+        // the computed local preference for each new candidate such that it is
+        // always less than the previous one.
 
         // Count the number of existing candidates of the same kind.
         let same_kind = self
@@ -550,10 +535,7 @@ impl IceAgent {
             .filter(|v| v.addr().is_ipv6() == ip.is_ipv6())
             .count() as u32;
 
-        let pref = same_kind * 2;
-        trace!("Calculated local preference adjustment: {}", pref);
-
-        c.local_preference_mut().sub_assign(pref);
+        c.adjust_local_preference(same_kind * 2);
 
         // Tie this ufrag to this ICE-session.
         c.set_ufrag(&self.local_credentials.ufrag);
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1,5 +1,6 @@
 use std::collections::{HashSet, VecDeque};
 use std::net::SocketAddr;
+use std::ops::SubAssign;
 use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
@@ -540,16 +541,6 @@ impl IceAgent {
         // 65534 - first ipv4
         // 65533 - second ipv6
         // 65432 - second ipv4
-        let counter_start: u32 = {
-            use CandidateKind::*;
-            let x = match c.kind() {
-                Host => 65_535,
-                PeerReflexive => 49_151,
-                ServerReflexive => 32_767,
-                Relayed => 16_383,
-            };
-            x - if ip.is_ipv6() { 0 } else { 1 }
-        };
 
         // Count the number of existing candidates of the same kind.
         let same_kind = self
@@ -574,10 +565,10 @@ impl IceAgent {
             0
         };
 
-        let pref = counter_start - same_kind * 2 - relay_across_ip_version_punishment;
-        trace!("Calculated local preference: {}", pref);
+        let pref = same_kind * 2 - relay_across_ip_version_punishment;
+        trace!("Calculated local preference adjustment: {}", pref);
 
-        c.set_local_preference(pref);
+        c.local_preference_mut().sub_assign(pref);
 
         // Tie this ufrag to this ICE-session.
         c.set_ufrag(&self.local_credentials.ufrag);
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1942,14 +1942,14 @@ mod test {
 
         // we expect:
         // (host/udp host/udp) - (0, 1)
-        // (host/udp rflx/udp) - (0, 1)
         // (rflx/udp host/udp) - (1, 1)
+        // (host/udp rflx/udp) - (0, 0)
         // (rflx/udp rflx/udp) - (1, 0)
         // (host/tcp host/tcp) - (2, 2)
 
         assert_eq!(
             agent.pair_indexes(),
-            [(0, 1), (0, 0), (1, 1), (1, 0), (2, 2)]
+            [(0, 1), (1, 1), (0, 0), (1, 0), (2, 2)]
         );
     }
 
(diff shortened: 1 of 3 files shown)


src/ice/agent.rs
```diff
@@ -1984,14 +1942,14 @@ mod test {
 
         // we expect:
         // (host/udp host/udp) - (0, 1)
-        // (host/udp rflx/udp) - (0, 1)
```
thomaseizinger, line 1987: This comment was already wrong, I think this should have said `(0,0)`

src/ice/agent.rs
```diff
@@ -1984,14 +1942,14 @@ mod test {
...
         // (rflx/udp host/udp) - (1, 1)
+        // (host/udp rflx/udp) - (0, 0)
         // (rflx/udp rflx/udp) - (1, 0)
         // (host/tcp host/tcp) - (2, 2)
 
         assert_eq!(
             agent.pair_indexes(),
-            [(0, 1), (0, 0), (1, 1), (1, 0), (2, 2)]
+            [(0, 1), (1, 1), (0, 0), (1, 0), (2, 2)]
```
thomaseizinger, line 1952: I am not quite sure why this order changes. Does that break anything @algesten?

src/ice/agent.rs
```diff
@@ -559,25 +535,7 @@ impl IceAgent {
...
-            }
-        } else {
-            0
-        };
-
-        let pref = counter_start - same_kind * 2 - relay_across_ip_version_punishment;
-        trace!("Calculated local preference: {}", pref);
-
-        c.set_local_preference(pref);
+        c.adjust_local_preference(same_kind * 2);
```
thomaseizinger, line 538: It is a bit smelly that this code knows that it needs to adjust by `same_kind * 2`. I contemplated about narrowing the API even further to just say `adjust_local_preference_for_nth_candidate` but I thought that was a bit too specific. Open to other ideas.

thomaseizinger:
@algesten Commit-by-commit review is encouraged. There are failing tests in-between though so I think we should still squash it at the end.

algesten:
@thomaseizinger I've looked through the commits, and if I summarize what's happening.

* Re-enable ip-version preferences tests
* Move most preference calculations to `Candidate`
* Add a new `Candidate::adjust_local_preference` (relative) and remove `set_local_preference` (absolute)
* Add a new `created_via` parameter to `relayed` to support the ip-version preference

Overall I think it's a bit sideways move rather than improvement. I'm not loving moving all priority calculations to the `Candidate`. How an agent prioritize candidates is situation dependent and more to do with `Agent`, than `Candidate` – for example, when doing WebRTC over certain mobile networks, IPv6 is categorically a better choice since that has lets NAT translations by the network operator.

If we are to move the calculations, I think they should be modular and configurable on `Agent` (with some reasonable default). The ip-version preference is in some respect "your" problem (I don't mean that in an aggressive way) – an elegant solution would be to make it possible to maintain that specific logic on your side.

# Idea

* Introduce `Extensions` on `Candidate`. https://docs.rs/http/1.3.1/http/struct.Extensions.html This would let users put anything they want on the Candidate struct.
* Make the calculation pluggable on `Agent`, and you could introduce the ip-version-preference as logic local to your code base
* Make it possible for the user to decide whether an extension should be cloned over to a reflexive candidate.
* Explore exposing the `Candidate` in question  on `Transmit`

thomaseizinger:
> The ip-version preference is in some respect "your" problem (I don't mean that in an aggressive way) – an elegant solution would be to make it possible to maintain that specific logic on your side.

I was under the impression that this is something that you agree with is useful to be the case for all users of str0m, not just for Firezone. So far I did not explore any pluggable solution because I thought we want to keep the logic within str0m.

As a first step towards this pluggable solution, we could make `adjust_local_preference` public. Then I would be able to call it on our relay candidates before passing them to the `IceAgent`. That would also remove the need for the `created_via` parameter so users who don't care about maintaining the same IP version across the connection don't need to bother setting this.

algesten:
> I was under the impression that this is something that you agree with is useful to be the case for all users of str0m, not just for Firezone. So far I did not explore any pluggable solution because I thought we want to keep the logic within str0m.

I don't know if it's useful for all users, but I also can't see the behavior have any adverse side effect. Since all of this was internal details in `IceAgent`, I didn't see a reason to _not_ have it in tree.

> As a first step towards this pluggable solution, we could make `adjust_local_preference` public. Then I would be able to call it on our relay candidates before passing them to the `IceAgent`. That would also remove the need for the `created_via` parameter so users who don't care about maintaining the same IP version across the connection don't need to bother setting this.

Does it? Isn't the problem that you only know how to adjust once you see the local-remote pair?

Could you maybe elaborate on why moving the preference calculation to `Candidate` is better arch than what we have?

thomaseizinger:
> Does it? Isn't the problem that you only know how to adjust once you see the local-remote pair?

No, the adjustment is local to the relay candidate! The IP versions that need to align are the one used to manage the allocation and the allocation address. A client can manage TURN allocations over IPv4 or IPv6 (TURN uses 3-tuples to identify the client so they are different allocations) and in both cases, it can allocate an IPv4 and an IPv6 address on the relay. The IP translation happens when e.g. a client only has IPv4 connectivity and therefore needs to manage its allocation over IPv4 but allocates an IPv4 and IPv6 relay address and then attempts to talk to a peer over IPv6.

What happens is:

- the client binds a channel to the IPv6 peer
- sends a channel data message with the to-be-relayed data over IPv4
- relay receives the channel data message over IPv4, strips the header and sends it back out over IPv6

> Could you maybe elaborate on why moving the preference calculation to `Candidate` is better arch than what we have?

Of course!

If we look at str0m as a whole, then how we calculate the local preference is almost entirely local to each candidate except for the adjustment of same-kind candidates, that is information that is only present in the `IceAgent`.

By moving the calculation to the candidate, we can directly incorporate the adjustment for relay candidates as well. An alternative design would be to make local-preference an i32 and start with a negative number for those relay candidates that describe a candidate with an IP version different from the one used for the TURN protocol.

I can't say that it is strictly _better_. As I commented above, it is a bit of a code smell that the `IceAgent` knows by how much it needs to adjust the preference for same-kind candidates. However, given that it is still a private API, I thought it is okay because it is internal to str0m. In a way, it is not very different to the `discarded` flag.

If we keep the current design of `adjust_local_preference` but make it public, then I can adjust it myself and we don't need to add the `created_via` parameter. Of course it is then public API and we need to commit to it. I think the most dangerous part here is that in a way, this also exposes the internal scheme str0m uses to compute the preference as one needs to know these bands to use them effectively. Considering that, I don't think it is a good idea to make it public.

algesten:
> By moving the calculation to the candidate, we can directly incorporate the adjustment for relay candidates as well.

Do you see, or disagree with my point that preference calculation is situation dependent and more a concern for the `IceAgent` than the `Candidate`?

thomaseizinger:
> > By moving the calculation to the candidate, we can directly incorporate the adjustment for relay candidates as well.
> 
> Do you see, or disagree with my point that preference calculation is situation dependent and more a concern for the `IceAgent` than the `Candidate`?

I don't disagree with it but I don't see them as strictly separate components. Everything around computing the local-preference are crate-private APIs so it doesn't matter that much where the computation happens. You can't really do anything with a `Candidate` other than add it to an `IceAgent`.

The `discarded` flag is also something that only the `IceAgent` cares about, yet it is stored in the `Candidate`.

Would you prefer a design where certain relay candidates start out with a negative local preference and the `IceAgent` then adjusts it further?

thomaseizinger:
>     * Introduce `Extensions` on `Candidate`. [docs.rs/http/1.3.1/http/struct.Extensions.html](https://docs.rs/http/1.3.1/http/struct.Extensions.html) This would let users put anything they want on the Candidate struct.
>     * Make the calculation pluggable on `Agent`, and you could introduce the ip-version-preference as logic local to your code base
>     * Make it possible for the user to decide whether an extension should be cloned over to a reflexive candidate.
>     * Explore exposing the `Candidate` in question  on `Transmit`

I don't mind this as a vision to work towards but it seems like a larger refactoring to me with lots of bike-shed opportunities. If you say that a design like the above is what is required to merge into main-line str0m then that is okay. I don't think I'll have time to work on it in the upcoming weeks though.

The fix on our end would very likely be to depend on a fork with this PR to restore the current functionality. Current mainline str0m without this optimisation is a massive performance regression for us that we cannot ship and the release prior to the fix breaks double symm-NAT setups.

if you don't want to commit to additional public APIs, we could make a 2nd constructor for relayed candidates that is `#[doc(hidden)]` as a compromise. That one would get removed again once the extension-based API is ready.

algesten:
@thomaseizinger I don't think what I proposed is that massive change. However, I've made a stab at a small PR going in the direction I think we should go in #668 

The one thing I omitted was the `Extensions` dynamic properties, and instead re-introduced a `local` address that is set for all Candidate kinds (not just relayed).

thomaseizinger:
Fixed in #668

PR #668, opened 2025-06-17
Make local preference calculation pluggable

algesten:
- **Candidate::local to track local interface**
- **Re-enable disabled tests**
- **Candidate::source_addr to use local instead of base**
- **Make local preference calculation pluggable**

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -565,7 +565,7 @@ impl IceAgent {
         // This punishment ensures that we prefer relayed within the same IP version,
         // e.g. IPv4 <> IPv4 over ones that translate between IP version, e.g. IPv4 <> IPv6.
         let relay_across_ip_version_punishment = if c.kind() == CandidateKind::Relayed {
-            if c.base().is_ipv4() != ip.is_ipv4() {
+            if c.local().is_ipv4() != ip.is_ipv4() {
                 1000
             } else {
                 0
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -397,6 +397,13 @@ impl IceAgent {
         debug!("max_retransmits = {num}");
     }
 
+    /// Sets the local preference calculation.
+    ///
+    /// This must be used before adding any local candidates.
+    pub fn set_local_preference(&mut self, p: impl LocalPreference) {
+        self.local_preference = LocalPreferenceHolder(Arc::new(p));
+    }
+
     fn bust_candidate_pair_timeout_caches(&mut self) {
         for pair in self.candidate_pairs.iter_mut() {
             pair.reset_cached_next_attempt_time();
@@ -1430,7 +1437,7 @@ impl IceAgent {
 
         let local = pair.local_candidate(&self.local_candidates);
         let proto = local.proto();
-        let local_addr = local.source_addr();
+        let local_addr = local.base();
         let remote = pair.remote_candidate(&self.remote_candidates);
         let remote_addr = remote.addr();
 
@@ -1516,7 +1523,7 @@ impl IceAgent {
 
         let trans = Transmit {
             proto: local.proto(),
-            source: local.source_addr(),
+            source: local.base(),
             destination: remote.addr(),
             contents: buf.into(),
         };
@@ -1672,7 +1679,7 @@ impl IceAgent {
             self.nominated_send = Some(best_prio.id());
             self.emit_event(IceAgentEvent::NominatedSend {
                 proto: local.proto(),
-                source: local.source_addr(),
+                source: local.base(),
                 destination: remote.addr(),
             })
         }
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1,9 +1,13 @@
 use std::collections::{HashSet, VecDeque};
+use std::fmt;
 use std::net::SocketAddr;
+use std::panic::RefUnwindSafe;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 use serde::{Deserialize, Serialize};
 
+use crate::ice_::preference::default_local_preference;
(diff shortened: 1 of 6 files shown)


src/ice/candidate.rs
```diff
@@ -151,6 +163,12 @@ impl Candidate {
             kind,
             raddr,
             ufrag,
+            match kind {
+                CandidateKind::Host => addr,
+                CandidateKind::PeerReflexive
+                | CandidateKind::ServerReflexive
+                | CandidateKind::Relayed => Self::arbitrary_raddr(addr),
+            },
```
algesten, line 175: This is a bit weird. Since this only applies to parsed (i.e. _remote_) candidates, let's just go with `Self::arbitrary_raddr` for all cases.

src/ice/agent.rs
```diff
@@ -1796,6 +1764,12 @@ impl IceAgent {
     }
 }
```
algesten, line 1772: Missing a `set_local_preference` to be able to configure the preference calculation.

src/ice/candidate.rs
```diff
@@ -74,6 +74,16 @@ pub struct Candidate {
...
 
+    /// The address of the local interface.
+    ///
+    /// * host - same as `addr`
+    /// * peer/server reflexive - same as `base`
+    /// * relay - set explicitly
+    ///
+    /// This is an extension to the ICE spec that we use to track
+    /// local interfaces also for relayed candidates.
+    local: SocketAddr,
```
thomaseizinger, line 89: If adding this field is an option, then we can start with a much smaller solution :)

With this information available, we can leave everything as is and simply compare the relay candidate address with this local address to calculate the preference punishment.

src/ice/candidate.rs
thomaseizinger: This is where a dedicated type that is solely responsible for deserialising from SDP (and feeding into `add_remote_candidate`) would pay off.

src/ice/candidate.rs
thomaseizinger: What I don't like so much about this design (but is not a blocker for me) is the semantic overlap with `base`.

src/ice/agent.rs
thomaseizinger: If the current default one stays as it is then I don't think there is a need to expose it right away.

src/ice/candidate.rs
```diff
@@ -414,12 +443,12 @@ impl Candidate {
     }
 
     pub(crate) fn source_addr(&self) -> SocketAddr {
```
thomaseizinger: I think `source_addr` can now be deleted again and we can just use `base`.

thomaseizinger (approved):
This design works for me, thank you for implementing it. I think I was optimising for different aspects in my iterations :)

src/ice/agent.rs
algesten: True

src/ice/candidate.rs
algesten: > If adding this field is an option, then we can start with a much smaller solution :)

"Eller hur?" as we say in Sweden ;)

> What I don't like so much about this design (but is not a blocker for me) is the semantic overlap with `base`.

Yeah, it's not ideal, but I think it's the best compromise between two things:

1. Spec. Base has a very specific meaning, and it's important to stick to spec ensure the dedupe etc works correctly (i.e. the start of this whole debacle).
2. Track local address. I think we should eventually expose the IP address to send from in the `Transmit`. I thought we had cracked that with the previous use of base, but alas.

So `local` is almost like `base`, apart from the relay case.

src/ice/candidate.rs
algesten: Oh. Interesting!

src/ice/candidate.rs
thomaseizinger: > 2\. Track local address. I think we should eventually expose the IP address to send from in the `Transmit`. I thought we had cracked that with the previous use of base, but alas.

Bear in mind that just the local address isn't enough. The traffic to relays needs to be wrapped in channel-data messages so needs special handling in any case.

thomaseizinger:
Thank you for the quick fix!

src/ice/candidate.rs
```diff
@@ -210,7 +234,13 @@ impl Candidate {
     /// * `addr` - The TURN server's allocated address that will be used for relaying traffic.
     ///            This is the address that will be used for communication with the peer.
     /// * `proto` - The transport protocol to use (UDP, TCP, etc.).
-    pub fn relayed(addr: SocketAddr, proto: impl TryInto<Protocol>) -> Result<Self, IceError> {
+    /// * `local` - The local interface address that corresponds to this candidate. This is the
+    ///             address from which the TURN allocation request was sent.
+    pub fn relayed(
+        addr: SocketAddr,
+        proto: impl TryInto<Protocol>,
+        local: SocketAddr,
```
thomaseizinger, line 242: I just updated to latest `main` and in the previous version, these two parameters were flipped. Do we want to flip them again? Then the upgrade would be a tad easier.

algesten: Hm. I deliberately flipped them since I thought of this as a breaking change. But is it? 🤔

thomaseizinger: I don't think it is a breaking change. The return value from `base` is breaking but behaviour of this parameter isn't.

thomaseizinger: `base` isn't public API so I think it isn't a breaking change at all.

algesten: Cool. Lets restore the order.

thomaseizinger: I'll send a PR.

PR #669, opened 2025-06-18
Flip parameters of `Candidate::relayed`

thomaseizinger:
Flips the order of these parameters to revert the breaking change from #668.


algesten (approved):
Remove the breaking note from the change log too please. 🙏

thomaseizinger:
> Remove the breaking note from the change log too please. 🙏

@algesten Done!

PR #670, opened 2025-07-05
fix(ice): don't disconnect if we are missing local candidates

thomaseizinger:
It is possible that an ICE agent receives remote candidates prior to discovering its own local candidates. For example, if the agent roams networks and therefore clears all its local candidates, it first needs to e.g. communicate with TURN servers to learn its new host and server-reflexive candidates. The signalling layer may be quicker in sending across the remote candidates than the TURN servers are in responding to the STUN bindings.

In this case, `str0m` will currently report the connection immediately as "Disconnected" even though it hasn't even formed any pairs.

The code already contains a special-case for this where we don't disconnect if we don't have any remote candidates yet. The same reasoning given there is also true for the local candidates though. Therefore, we extend this check to also include local candidates.

Related: https://github.com/firezone/firezone/pull/9793

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -1779,8 +1779,10 @@ impl fmt::Debug for LocalPreferenceHolder {
 
 #[cfg(test)]
 mod test {
+    use crate::ice_::test::host;
+
     use super::*;
-    use std::net::SocketAddr;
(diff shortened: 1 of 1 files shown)


thomaseizinger:
I've stacked this onto #671 for easier merging.

PR #671, opened 2025-07-05
ci: fix build

thomaseizinger:
This fixes several issues with CI.

1. Rust 1.88 has been released and extended a clippy lint that wants variables to be directly put into the format string. We disable this lint for the entire crate.
2. The latest Windows image from GitHub includes a newer VisualStudio installation with a buggy linker that cannot handle symbols above a certain size. See https://github.com/actions/runner-images/issues/12432 for details. To fix this, we switch to `rust-lld.exe` as the linker.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -104,11 +104,11 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets -- -D warnings
+          args: -- -D warnings
       - uses: actions-rs/cargo@v1
         with:
           command: clippy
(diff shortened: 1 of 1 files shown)


algesten (changes requested):
I want us to turn off this clippy lint. I don't agree `"{foo}"` is better than `"{}", foo` since there are many cases you can't use the inline notation, such as `"{foo.bar}"`.

When Rust introduced this, they should have taken it all the way. I want all of this to be uniform, so the clippy is not correct in my code bases.

thomaseizinger:
Okay no worries, we can turn it off for the entire workspace :)

Away from my computer now but will fix up tomorrow.

thomaseizinger:
@algesten I've dropped the previous commits and allowed the lint for the entire crate.

PR #672, opened 2025-07-07
fix(ice): disallow mixed IP versions for srflx candidates

thomaseizinger:
It is impossible to have a server-reflexive IPv4 candidate that sends from an IPv6 base and vice versa.


thomaseizinger:
This is stacked onto #671.

algesten:
Seems reasonable

algesten (approved):

PR #674, opened 2025-07-17
Add methods to configure STUN retransmission settings

### LINKED ISSUE #673: Expose STUN timing setters (set_max_stun_*) so that ICE disconnects can be detected sooner
### Problem
When a peer disappears it takes about 27 seconds before IceConnectionState::Disconnected is raised. The delay comes from the default STUN timing in IceAgent:

• initial RTO: 250 ms
• max RTO: 3 s
• max retransmits: 9

These values lead to roughly 19–27 s of retries before a candidate pair is declared failed.

What we need
Latency-sensitive applications must detect a lost peer in ≤ 2 s. IceAgent already provides the right knobs:

set_initial_stun_rto(Duration)
set_max_stun_rto(Duration)
set_max_stun_retransmits(usize)

But these setters are not reachable through the public API; Rtc hides the underlying IceAgent.

### Request
Expose the three setters on Rtc / RtcConfig, or offer another public way to override STUN-timing parameters while building an Rtc.

### Why it matters
Being able to tune STUN retransmission lets applications trade network traffic for faster failure detection. In real-time media or control channels, a 20–30 s blackout is unacceptable, whereas a 1–2 s detection time is fine.

Sanity check of facts
• Default max retransmits = 9
• Default max RTO = 3 s
• The three setter functions exist on IceAgent
• They are not exposed by the high-level API

All statements match the current str0m source.

yjhong79:
Resolves #673

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+  * Make STUN timers configurable on RtcConfig #674
+  * ice: stricter srflx candidate checks #672
+  * ice: Dont report Disconnected when no local candidates #670
   * Make ICE agent local preference pluggable #668
   * Simplified serde of SDP #664
   * SR/RR stats in Ingress/EgressStats #661 #662
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2059,6 +2059,8 @@ impl RtcConfig {
     ///
     /// This is the initial wait time before a STUN request is retransmitted.
     /// The timeout will double with each retry, starting from this value.
+    ///
+    /// Defaults to 250ms.
     pub fn set_initial_stun_rto(&mut self, rto: Duration) {
         self.initial_stun_rto = Some(rto);
     }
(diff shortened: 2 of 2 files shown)


thomaseizinger (approved):
@algesten any objections?

yjhong79:
> This should be config, not runtime configuration for two reasons.
> 
> It's important to keep the `Rtc` instance clean to make the API easy to understand.
> 
> While there maybe situations where you want to change these values runtime,that's a secondary usage.
> 
> We should move this to `RtcConfig` further down the file.

Considering that IceAgent already has default values for these fields, do you think it’s better to only override them when explicitly set in RtcConfig options?
Or should we instead source the default values from RtcConfig regardless?

algesten:
I think Option in config and set if configured.

yjhong79:
Applied.

algesten (approved):

algesten:
Thanks!

PR #675, opened 2025-07-18
feat(ice): expose remote ICE credentials

thomaseizinger:
Firezone employs an idempotent signalling protocol where we don't do anything, if the ICE credentials are still the same. Currently, we decide this based on the local ICE credentials only but it would be better to also compare the remote ones. For this, the ICE agent should expose the remote credentials just like it exposes the local ones.


algesten (approved):

algesten:
@thomaseizinger looks good! Add a changelog row and merge!

PR #677, opened 2025-07-27
test(ice): simulate port-restricted-cone and symmetric NATs

thomaseizinger:
I was investigating our hole-punch behaviour with different NATs and thought it would be useful to have some tests in str0m that actually simulate different kinds of NATs to see what gets nominated.

### DIFF
--- a/src/ice/mod.rs
+++ b/src/ice/mod.rs
@@ -970,7 +970,7 @@ mod test {
     }
 
     #[test]
-    pub fn symmetric_nat_one_side_uses_peer_reflexive_candidate() {
+    pub fn symmetric_nat_one_side() {
         let _guard = tracing_subscriber::fmt()
             .with_test_writer()
             .with_env_filter("trace")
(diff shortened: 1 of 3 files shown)


thomaseizinger:
Hold on, I think some of the NAT implementation is still wrong ..

src/ice/mod.rs
```diff
@@ -1146,4 +1226,171 @@ mod test {
...
+        let (from, to) = if let Some(nat) = &mut from_agent.nat {
+            // If our traffic is "from" a relay candidate, the NAT does not apply.
+            if outgoing_is_from_relay {
+                (from, to)
+            } else {
+                nat.transform_outbound(from, to)
+            }
+        } else {
+            (from, to)
+        };
```
thomaseizinger, line 1380: let-chains would make this a lot nicer but we also run our tests for our MSRV so we can't use it here :(

src/ice/mod.rs
```diff
@@ -1038,16 +1056,64 @@ mod test {
...
+            let local = local.parse().unwrap();
+            let addr = addr.parse().unwrap();
+
+            self.agent
+                .add_local_candidate(Candidate::relayed(addr, local, "udp").unwrap())
+                .unwrap()
+                .clone()
+        }
+
+        fn server_reflexive_candidate(&mut self, stun_server: &str, base: &str) -> Candidate {
```
algesten, line 1097: Why not `add_xxx`? here for consistency?

src/ice/mod.rs
algesten: Yeah, it's ok. One day we raise MSRV :)

algesten (approved):
LGTM!

src/ice/mod.rs
thomaseizinger: Because server-reflexive candidates are usually not added to the local client, they are just sent to the remote.

algesten: ah

thomaseizinger: In a way, this just performs STUN. I'll add a comment to clarify.

PR #683, opened 2025-07-31
chore(ice): don't log last remote binding attempt time

thomaseizinger:
A raw `Instant` isn't particularly useful because it is a platform-dependent implementation detail, what this `Instant` is relative to. Therefore, remove this from the debug output.

### DIFF
--- a/src/ice/pair.rs
+++ b/src/ice/pair.rs
@@ -455,7 +455,7 @@ impl fmt::Debug for CandidatePair {
         write!(
             f,
             "CandidatePair(\
-                {}-{} ({}-{}) prio={} state={:?} attempts={} unanswered={} remote={} last={:?} nom={:?}\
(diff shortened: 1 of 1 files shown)


algesten:
Can't we just not debug print the field?

thomaseizinger:
> Can't we just not debug print the field?

That is what we currently do. It debug prints the `Instant` which is pretty useless. Unless you have another `Instant` somewhere in the same log stream, you have no idea what it refers to.

algesten:
Yeah, I mean we drop the field altogether from the debug output.

thomaseizinger:
> Yeah, I mean we drop the field altogether from the debug output.

I you meant remove it altogether. Yeah we can do that too if you don't think the output is useful.

thomaseizinger:
Done! I've force-pushed a different change.

algesten (approved):
A mi, me gusta!

thomaseizinger:
> A mi, me gusta!

Que bien, vamos!

PR #685, opened 2025-08-07
Add PCM support

HunterNL:
Adds support for PCMU and PCMA, payload types 0 and 8. I've written the bare minimum to make this functional and I'd be happy to flesh this out if desired.

### DIFF
--- a/src/packet/mod.rs
+++ b/src/packet/mod.rs
@@ -8,7 +8,7 @@ use crate::format::Codec;
 use crate::sdp::MediaType;
 
 mod g7xx;
-use g7xx::{G711Packetizer, G722Packetizer};
+use g7xx::{G711Depacketizer, G711Packetizer, G722Packetizer};
 
 mod h264;
 pub use h264::H264CodecExtra;
@@ -199,6 +199,7 @@ pub(crate) enum CodecPacketizer {
 
 #[derive(Debug)]
 pub(crate) enum CodecDepacketizer {
+    G711(G711Depacketizer),
     H264(H264Depacketizer),
     H265(H265Depacketizer),
     Opus(OpusDepacketizer),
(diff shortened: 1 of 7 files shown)


src/packet/mod.rs
```diff
@@ -229,6 +231,8 @@ impl From<Codec> for CodecDepacketizer {
     fn from(c: Codec) -> Self {
         match c {
             Codec::Opus => CodecDepacketizer::Opus(OpusDepacketizer),
+            Codec::PCMU => CodecDepacketizer::Opus(OpusDepacketizer),
+            Codec::PCMA => CodecDepacketizer::Opus(OpusDepacketizer),
```
algesten: These packetizers look dodgy. G711 in one direction and Opus in the other?

How does PCM packetize?

algesten (commented):
Thanks for looking at this! One question.

src/packet/mod.rs
xnorpx: Yeah probably worth just copy the few lines of depacketizer from Opus.

G711 usually runs on 20 ms packet interval/time which corresponds to 160 bytes per packet. so will fit in one MTU

algesten: Well. 160 _ought_ to be within the MTU :)

HunterNL: It does look very dodgy haha. As far as I understand the payloader does most of the work here, already splitting the samples into 160 byte packages. With a sufficiently high MTU the packetizer logic doesn't come into play. The Opus depacketizer then doesn't do anything Opus specific and just reassembles the payload.

I'll still move this to a dedicated depacketizer though

HunterNL:
I've plainly duplicated the Opus depacketizer and its sole relevant test(?). Wasn't sure what to do about the partition head/tail methods.

Looking at the code more it did struck me this might be better of as a single G.711 codec with a FormatParam/CodecExtra determining the u-law/a-law encoding. What do you think of that approach?

xnorpx:
> @HunterNL thanks!
> 
> @xnorpx this looks ready to ship to me. Do you agree?

Just nits but lgtm

algesten:
Thanks!

PR #686, opened 2025-08-13
Implement buffered amount and buffered amount low event

UE2020:
I exposed sctp-proto's buffered amount API under str0m's `Channel` struct and added a new event that fires when sctp-proto's `BufferedAmountLow` event fires.

I wasn't sure what to do when the user calls the buffered amount functions before the sctp `state` field it `Established`. I added a new variant to the SctpError enum. Let me know if changes need to be made there.

### DIFF
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -59,6 +59,11 @@ impl<'a> Channel<'a> {
     pub fn buffered_amount(&mut self) -> Result<usize, RtcError> {
         Ok(self.rtc.sctp.buffered_amount(self.sctp_stream_id)?)
     }
+
+    /// Set the threshold to emit an [`Event::ChannelBufferedAmountLow`][crate::Event::ChannelBufferedAmountLow]
+    pub fn set_buffered_amount_low_threshold(&mut self, threshold: usize) -> Result<(), RtcError> {
+        Ok(self.rtc.sctp.set_buffered_amount_low_threshold(self.sctp_stream_id, threshold)?)
+    }
 }
 
 impl fmt::Debug for ChannelData {
--- a/src/sctp/mod.rs
+++ b/src/sctp/mod.rs
@@ -108,6 +108,9 @@ pub(crate) enum SctpEvent {
         binary: bool,
         data: Vec<u8>,
     },
+    BufferedAmountLow {
+        id: u16
+    }
 }
 
 /// These are the possible paths:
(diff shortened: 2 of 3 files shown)


src/channel.rs
```diff
@@ -54,6 +54,16 @@ impl<'a> Channel<'a> {
...
+
+    /// Get the amount of buffered data.
+    pub fn buffered_amount(&mut self) -> Result<usize, RtcError> {
+        Ok(self.rtc.sctp.buffered_amount(self.sctp_stream_id)?)
+    }
+
+    /// Set the threshold to emit an [`Event::ChannelBufferedAmountLow`][crate::Event::ChannelBufferedAmountLow]
+    pub fn set_buffered_amount_low_threshold(&mut self, threshold: usize) -> Result<(), RtcError> {
+        Ok(self.rtc.sctp.set_buffered_amount_low_threshold(self.sctp_stream_id, threshold)?)
+    }
```
lolgesten, line 66: This shouldn't be runtime config on `Channel`, but instead options set in the `ChannelConfig` struct.

src/sctp/mod.rs
```diff
@@ -369,6 +376,57 @@ impl RtcSctp {
...
+        if rec.state != StreamEntryState::Open {
+            return Err(SctpError::BufferedAmountQueryBeforeEstablished);
+        }
+
+        let mut stream = assoc.stream(id)?;
+        stream.set_buffered_amount_low_threshold(threshold)?;
+
+        Ok(())
+    }
+
```
lolgesten, line 429: By using `ChannelConfig`, we can simplify this since I don't think we need to consider the "Before Established" situation. The `Channel::buffered_amount()` could return `0` until the SCTP is established.

src/channel.rs
UE2020: Unless I'm misunderstanding str0m's channel API, isn't putting the buffered amount low threshold on the ChannelConfig struct restrictive, because then you have to add the channel using SdpApi or DirectApi to set it? To be honest, I don't have much experience with datachannels in str0m.

algesten: I see what you mean. We need to support channels that are started on by the other side.

algesten:
@UE2020 I couldn't push additional commits to this PR, so I took your changes and worked on them further in #688

PR #688, opened 2025-08-17
Add buffered_amount methods to Channel

algesten:


algesten:
@UE2020 I didn't want to gate the functions `buffered_amount()` and `get_buffered_amount()` on associations etc being up, since parts of that are not observable states for the user. Instead we allow configuring the threshold at any time and propagate it to the actual SCTP stream when we can.

algesten:
@k0nserv or @davibe can I have a review here?

src/sctp/mod.rs
```diff
@@ -344,7 +371,7 @@ impl RtcSctp {
 
         let rec = self
             .entries
-            .iter()
+            .iter_mut()
```
k0nserv: This shouldn't be needed

src/sctp/mod.rs
```diff
@@ -86,6 +86,30 @@ struct StreamEntry {
...
+    /// If the queued outgoing data drops below this threshold, Rtc is to emit an
+    /// event to the user.
+    buffered_threshold: BufferedThresholdConfig,
+}
+
+#[derive(Debug)]
+enum BufferedThresholdConfig {
+    Unconfigured,
+    Desired(usize),
+    Configured(usize),
```
k0nserv, line 106: Can we have some doc comments to describe these variants?

src/sctp/mod.rs
algesten: Very true

src/sctp/mod.rs
algesten: Sure

PR #690, opened 2025-08-21
fix: panic on STUN Binding Indication

lherman-cs:
With certain clients such as OBS WHIP, the client may send a STUN binding indication without USERNAME attribute. There's no need to process the message per
https://datatracker.ietf.org/doc/html/rfc8489#section-6.3.2.

```
Session Traversal Utilities for NAT
    Message Type: 0x0011 (Binding Indication)
    Message Length: 8
    Message Cookie: 2112a442
    Message Transaction ID: 039f587e6a860fcac95f37bc
    [STUN Network Version: RFC-5389/8489 (3)]
    Attributes
        FINGERPRINT
            Attribute Type: FINGERPRINT
            Attribute Length: 4
            CRC-32: 0xc8507de3 [correct]
            [CRC-32 Status: Good]
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
 
 [[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


thomaseizinger (approved):
Nice find! Could you add a unit-test?

lherman-cs:
@thomaseizinger @algesten added a unit test using the hex dump from wireshark.

algesten:
Thanks!

PR #695, opened 2025-09-04
Count STUN messages in received bytes

### LINKED ISSUE #694: Receive Stun bytes not counted
Ok, so this has bit us twice now, we don't see any recv bytes counted when we have no active streams. So figure it worth having an issue to track it.

There is no quick clean way to fix this, which is probably the reason it has a todo.

We switched to track the bytes outside str0m instead for now.

```rust
   let bytes_rx = match r.contents.inner {
            // TODO: stun is already parsed (depacketized) here
            Stun(_) => 0,
            Dtls(v) | Rtp(v) | Rtcp(v) => v.len(),
        };
```

https://github.com/algesten/str0m/blob/783b020ff7c859372884889e205c5c838792e97a/src/lib.rs#L1824C9-L1828C11

algesten:
Close #694

### DIFF
--- a/src/io/stun.rs
+++ b/src/io/stun.rs
@@ -121,6 +121,7 @@ pub struct StunMessage<'a> {
     attrs: Attributes<'a>,
     integrity: &'a [u8],
     integrity_len: u16,
+    unparsed_len: Option<usize>,
 }
 
 impl<'a> StunMessage<'a> {
@@ -129,6 +130,8 @@ impl<'a> StunMessage<'a> {
         if buf.len() < 4 {
             return Err(StunError::Parse("Buffer too short".into()));
         }
+
+        let unparsed_len = buf.len();
 
         let typ = (buf[0] as u16 & 0b0011_1111) << 8 | buf[1] as u16;
         let len = (buf[2] as u16) << 8 | buf[3] as u16;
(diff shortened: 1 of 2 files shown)


algesten:
@xnorpx looks ok?

k0nserv:
I think the reason we don't do this is to somewhat match the spec, but I don't recall exactly. In the spec you have stats counted at the ICE level which do include STUN, but then others that are more media focused.

I think we modelled it after [RTCTransportStats](https://www.w3.org/TR/webrtc-stats/#dom-rtctransportstats)'s bytesReceived which says

>  bytesReceived of type unsigned long long

>    Represents the total number of payload bytes received on this RTCIceTransport, i.e., not including headers, padding or ICE connectivity checks.

algesten:
@k0nserv exciting!

So config option? (lol)

k0nserv:
I dunno. I think it's useful to match the spec because folks might end up comparing with what the browser reports. I'm not sure if there's a value in the spec that just reports raw bytes received that would be suitable.

algesten:
Ok! Let's hear what @xnorpx says

xnorpx:
Ok that's fair, but then I would argue that we should not report it on send :)

xnorpx:
We can use the raw packet sizes and track it outside, but we should make sure send and receive does the same thing.

algesten:
> Ok that's fair, but then I would argue that we should not report it on send :)

Fair. Let's make sure they are reported the same way.

PR #697, opened 2025-09-10
Add AES_256_GCM support for SRTP

efer-ms:
Add AEAD_AES_256_GCM support for SRTP. This is the "preferred" crypto, listed ahead of AEAD_AES_128_GCM and Aes128CmSha1_80

Most of the work was to declare similar functions to the AEAD_AES_256_GCM cases. The biggest shared code change was in SrtpKey::derive() to make it able to take in a 256-bit master key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+ * dtls: support (and prefer) AES_256_GGCM #697
  * ice: fix panic on STUN Binding Indication #690
 
 # 0.10.0


efer-ms:
@algesten, we ran into a desire to use AES_256_GCM with SRTP. Let me know if this looks good or if you want changes. I tested with OpenSSL and Wincrypto in our app, and was able to negotiate and use AES_256_GCM with libwebrtc clients.

src/crypto/srtp.rs
```diff
@@ -11,13 +10,17 @@ pub enum SrtpProfile {
...
 #[allow(dead_code)]
 impl SrtpProfile {
     // All the profiles we support, ordered from most preferred to least.
-    pub(crate) const ALL: &'static [SrtpProfile] =
-        &[SrtpProfile::AeadAes128Gcm, SrtpProfile::Aes128CmSha1_80];
+    pub(crate) const ALL: &'static [SrtpProfile] = &[
+        SrtpProfile::AeadAes256Gcm,
+        SrtpProfile::AeadAes128Gcm,
+        SrtpProfile::Aes128CmSha1_80,
+    ];
```
algesten, line 23: @k0nserv any thoughts on changing the order here to prefer 256?

k0nserv: It'll for sure be slower, but it will also be an increasing regulatory requirement I presume. I don't know enough about the crypto world to say anything smart, 128 bit still fine AFAIK

xnorpx: Should we allow the "advanced" user to provide an order?

k0nserv: > Should we allow the "advanced" user to provide an order?

Yes, but let's not add it until requested. Also, I was thinking, we should check what default libWebRTC uses

algesten: @xnorpx @efer-ms can we mirror the preference order of libWebRTC for now?

efer-ms: Found where libwebrtc has it's default. https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/crypto/crypto_options.cc;l=36;drc=48d6f7175422b2c969c14258f9f8d5b196c28d18

It actually has the default ordering of....
AES128_CM_SHA1_32, AES128_CM_SHA1_80, AEAD_AES256_GCM, AEAD_AES128_GCM

The rationale they give is that the GCM modes require more bytes, so even though they are more secure, the others are preferred, unless disabled. But considering they in the same breath say AES128_CM_SHA1_80 is required.. it doesn't quite make sense to me.

Str0m currently is AEAD_AES128_GCM, followed by AES128_CM_SHA1_80 .. which seems reasonable. So I think we have two choices, match libwebrtc: AES128_CM_SHA1_80, AEAD_AES256_GCM, AEAD_AES128_GCM  ... which is different than str0m today in that CM is preferred over GCM.

Or continue with the most secure as the preferred item: AEAD_AES256_GCM, AEAD_AES128_GCM, AES128_CM_SHA1_80 ...

TL;DR: libwebrtc prefers AEAD_AES256_GCM over AEAD_AES128_GCM, however, it prefers AES128_CM_SHA1_80 over either GCM mode which is different than Str0m currently.

algesten: @efer-ms do you have requirements that mean you want a specific order, or should we follow libWebRTC?

k0nserv: @algesten Keep in mind we have specific requirements, in particular we want `AEAD_AES128_GCM` above `AES128_CM_SHA1_80` because the former is cheaper. I think @xnorpx is correct that this should probably be use configurable

xnorpx: @algesten if we plumb the order we are good to go?
cc: @efer-ms

algesten: @xnorpx i much prefer to have sensible defaults than make this configurable.

algesten: @k0nserv ok. I guess we can agree on `AEAD_AES256_GCM` first then.

k0nserv: Relevant performance data here https://github.com/lelegard/aesbench/blob/main/RESULTS.txt

PR #701, opened 2025-09-19
Fix Vp9Depacketizer failing on high scalability modes

### LINKED ISSUE #693: `ErrTooManyPDiff` error when testing using VP9 + SVC
Changing the chat example to use VP9 + SVC makes str0m library to return the following warning and the chat example has no video:

`WARN chat: Client (0) poll_output failed: Packet(Mid(1), Pt(98), ErrTooManyPDiff)`

To reproduce this issue just apply the following patch:

[chat.patch](https://github.com/user-attachments/files/22084254/chat.patch)

It seems that the N bit from the last P_DIFF is not valid. Limiting the loop to 3 times worked for me:

```
diff --git a/src/packet/vp9.rs b/src/packet/vp9.rs
index 6f5b40e..87477ca 100644
--- a/src/packet/vp9.rs
+++ b/src/packet/vp9.rs
@@ -502,17 +502,17 @@ impl Vp9Depacketizer {
         reader: &mut dyn BitRead,
         mut payload_index: usize,
     ) -> Result<usize, PacketError> {
-        let mut b = 1u8;
-        while (b & 0x1) != 0 {
+        while self.pdiff.len() < MAX_VP9REF_PICS {
             if reader.remaining() == 0 {
                 return Err(PacketError::ErrShortPacket);
             }
-            b = reader.get_u8().ok_or(PacketError::ErrShortPacket)?;
+
+            let b = reader.get_u8().ok_or(PacketError::ErrShortPacket)?;
             payload_index += 1;
 
             self.pdiff.push(b >> 1);
-            if self.pdiff.len() >= MAX_VP9REF_PICS {
-                return Err(PacketError::ErrTooManyPDiff);
+            if (b & 0x1) == 0 {
+                break;
             }
         }
```

Could you please confirm if this is an issue or not?

alexlapa:
Fixes #693 

Fixes `ErrTooManyPDiff` and `ErrVP9CorruptedPacket` for VP9 in L3T3 or S3T3 calability modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Unreleased
 
+ * srtp: support (and prefer) AES_256_GCM #697
  * ice: fix panic on STUN Binding Indication #690
+ * bwe: fix panic on history monotonicity assert in `TrendlineEstimator` #700
  * fix `Vp9Depacketizer` failing on high scalability modes #701
 
 # 0.10.0
--- a/src/crypto/mod.rs
+++ b/src/crypto/mod.rs
@@ -86,7 +86,7 @@ mod keying;
 pub use keying::KeyingMaterial;
 
 mod srtp;
-pub use srtp::{aead_aes_128_gcm, aes_128_cm_sha1_80, SrtpCrypto, SrtpProfile};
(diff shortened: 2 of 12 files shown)


algesten:
The array is a nice improvement!

PR #703, opened 2025-09-22
Update thiserror to 2.0.16

xnorpx:
To avoid duplicated crates, update thiserrror to 2.0.

I will then update sctp-proto once this change is in.

PR #704, opened 2025-09-22
Remove thiserror

algesten:
Alternative to #703 where we remove thiserror completely. procmacros can't be cached effectively in recompilation, hence they slow down compile times. It's better to not use them.

Close #703


algesten:
@xnorpx this is ready to ship I think.

PR #705, opened 2025-09-24
feat: allow ordering for Mid

lherman-cs:
this is mainly to improve DX so developers can use #[derive(Ord)] when Mid is wrapped in another container


algesten:
Sounds good!

PR #706, opened 2025-09-25
Remove use of thiserror from wincrypto

efer-ms:
Change WinCryptoError to be an Enum with different wrapped values. Simplify most of the code emitting error from WinCrypto to just use .into() on the Windows result object or string.

### DIFF
--- a/wincrypto/src/dtls.rs
+++ b/wincrypto/src/dtls.rs
@@ -195,10 +195,10 @@ impl Dtls {
             }
             EstablishmentState::Handshaking => self.handshake(datagram),
             EstablishmentState::Failed => {
-                Err(WinCryptoError("Handshake failed".to_string()).into())
+                Err("Handshake failed".into())
             }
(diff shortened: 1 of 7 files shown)


wincrypto/src/dtls.rs
```diff
@@ -423,15 +413,15 @@ impl Dtls {
             e => {
                 // Failed
                 self.state = EstablishmentState::Failed;
-                Err(WinCryptoError(format!("DTLS handshake failure: {:?}", e)).into())
+                Err(format!("DTLS handshake failure: {:?}", e).into())
```
xnorpx: can't you do {e:?} here?

efer-ms: actually.. I can just convert the hresult to a WinCryptoError.

xnorpx (approved):
lgtm

algesten:
Nice!

PR #708, opened 2025-09-25
Dimpl and RustCrypto

algesten:


algesten:
~This branch is currently setting the default to dimpl, but I don't think that's how we're going to merge it. Instead we probably feature flag dimpl, leave openssl as the default and use it if the feature flag is on. So the order is:~

...

`dimpl` comes as last option.

Cargo.toml
```diff
@@ -35,6 +35,9 @@ sha1 = ["dep:sha1"]
 # openssl and sha1.
 wincrypto = ["dep:str0m-wincrypto"]
 
+# DTLS impl made for str0m (experimental) in pure Rust.
+dimpl = ["dep:dimpl"]
```
xnorpx: Should the rustcrypto primitives only be added when dimpl is enabled?

xnorpx: dimpl = ["dep:dimpl"], "dep:aes", ..... etc

algesten: Yeah, good point. They're not really related, but for simplicity we maybe make them.

Cargo.toml
```diff
@@ -24,17 +24,20 @@ rust-version = "1.71.1"
 # security patching than updating statically linked software.
 
 [features]
-default = ["openssl", "vendored", "sha1"]
+default = ["dimpl", "vendored", "sha1"]
```
xnorpx: vendored is tied to openssl so if dimpl is default then you don't need vendored?

.github/workflows/cargo.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
```
xnorpx, line 17: so if dimpl is default then add a section

```yaml

  test_openssl:
    strategy:
      matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
        rust: [stable, beta, 1.81.0]

    runs-on: ${{ matrix.os }}

    steps:
      - uses: actions/checkout@v2
      - uses: dtolnay/rust-toolchain@master
        id: toolchain
        with:
          toolchain: ${{ matrix.rust }}
      - uses: Swatinem/rust-cache@v1
      - name: Build
        run: cargo +${{steps.toolchain.outputs.name}} build --no-default-features --features openssl
      - name: Test
        run: cargo +${{steps.toolchain.outputs.name}} test --no-default-features --features openssl
```

xnorpx: ok, this should be change to test the dimpl impl instead

algesten: Very true. It should try all the backends.

algesten: Done

ValorZard:
I am very excited to see this PR! Let me know if I can do anything to help! The fact that str0m had such a hard dependency on either wincrypto or openssl was a big blocker in terms of me using it, so I'd love to chip in in any way that I can

xnorpx:
> I am very excited to see this PR! Let me know if I can do anything to help! The fact that str0m had such a hard dependency on either wincrypto or openssl was a big blocker in terms of me using it, so I'd love to chip in in any way that I can

test it? :)

algesten:
I suspect a big drain on performance is how much zeroing dimpl does on the buffers it uses for encrypted data. This line could be the culprit https://github.com/algesten/dimpl/blob/main/src/buffer.rs#L27

PR #710, opened 2025-09-27
github CI: use dtolnay/rust-toolchain

algesten:


.github/workflows/cargo.yml
```diff
@@ -21,19 +22,18 @@ jobs:
...
-      - uses: actions-rs/cargo@v1
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@master
+        id: toolchain
         with:
-          command: test
+          toolchain: ${{ matrix.rust }}
+      - uses: Swatinem/rust-cache@v2
+      - name: Build 1
+        run: cargo +${{steps.toolchain.outputs.name}} build --all-features
```
thomaseizinger, line 34: I think the action sets the version as default so you could technically get rid of this (unless you want to be explicit)

algesten: Thanks!

PR #714, opened 2025-10-25
Efer/apple crypto

efer-ms:
Progress! Still missing DTLS .. but cert, ciphers, hashes are all implemented.. just need DTLS.


algesten:
Oh wow! I had no idea you were looking at this. Very cool!

It struck me when I was integrating dimpl that the surface between str0m and the DTLS layer is a bit untidy because it was quite openssl specific. 

Wonder if we should revisit that layer now that we get more and more deps.

efer-ms:
Well this will likely be stuck for a bit, since it looks like Apple doesn't have a reasonable DTLS impl available after all. So whilst the ciphers, hashes, and cert generation seem to all be possible, DTLS itself is another matter.

Didn't mean to get your hopes up, but it does mean if you want to rethink the interface, you don't need to worry about migrating this one exactly.

xnorpx:
Could the DTLS 1.2 impl be reused from dimpl while all crypto things come from apple?

efer-ms:
> Could the DTLS 1.2 impl be reused from dimpl while all crypto things come from apple?

That seems like the most promising approach.

algesten:
We'd need to make the crypto pluggable in dimpl, but there isn't much because it severely restricts the supported ciphers.

efer-ms:
Ok.. I spent a good amount of time looking at what was there in Network Framework and the older deprecated Security Framework.

- Network Framework just seems to hide too much. You can't replace the actual UDP socket, and you can't capture packets after DTLS but before being sent on the wire.
- Security Framework (ignoring it being Deprecated for years), got further. I managed to get DTLS to establish with RSA certs, but not ECDSA certs, and the SRTP key material was not generated correctly. Looking at the actual handshake though showed it still performing DTLS 1.0 though I set it to min 1.2, and no ability to set the ec curve params not use_srtp extensions (which I suspect has to do with the ECDSA and SRTP failures).

So... I think the best option would be to consume dimpl. I think it still allows people to use system crypto, vs bringing their own. it's unfortunate that Apple didn't continue with the Security Framework and just add the few things it needed to be effective. But I'm sure they had their reasons /shrug

algesten:
> DTLS 1.0 though I set it to min 1.2,

Notice that on the record layer most DTLS impls say 1.0 even for 1.2

https://github.com/algesten/dimpl/blob/main/src/message/record.rs#L80-L83

But elliptic curve crypto in 1.2 do require extra extensions, so it doesn't surprise me that apple doesn't do it even if they do 1.2.

efer-ms:
Perhaps, although I was trying to force 1.2, but who knows. for sure there was no clear way to mess with the extensions. And Security Framework is deprecated, although it showed the most promise. Anyways, once you have dimpl in a reasonable state, I can see if I can get it working with applecrypto.

algesten:
Oh?

efer-ms:
I was just removing it from something that was getting ready for review. I still have my branch that I'll dust off at some point. Once we think we're good using dimpl for the actual protocol parts.

Issue #716, opened 2025-10-27
Runaway loop, out of memory crash in TWCC responder


kvasilye:
Hello,

We use str0m 0.10 in production and are running into a bug on the subscribe side if TWCC is enabled.

The bug appears in the code which builds a TWCC response, specifically when it detects a gap in the sender's sequence numbers - it thinks that the gap is much larger than it is (in the 10 ^ 20 range) and quickly runs out of memory when trying to build the list of missing SEQ's. And then Linux's OOM killer kicks in and it's all over.

Reproducibility is about 3/5, happens seconds to minutes after the connection is established.

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/5e925319-2614-42ea-8431-6b22a596a126" />

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/3e9bd1a1-df4e-49b4-9e20-fd84efd6d194" />

-------

One time when I stopped in the debugger, the `interims.len` was 4222938112 and the value of `todo` loop control variable was `18446709479400701806` so tried to loop basically forever.

Each iteration through the loop pushes one element to the array and decrements the loop variable by 8192. Let's reconstruct the original value of the loop variable. `echo "obase=16;18446709479400701806+8192*4222938112" | bc` outputs: `10000000000027F6E`, looks like some kind of overflow to me.

Since TWCC sequence numbers are 16 bit, the number of missing sequences should never be that large.

algesten:
This sounds like a bug when we extend the sequence numbers.

When str0m receives a sequence number, we always extend it to a u64 for straight away. https://github.com/algesten/str0m/blob/main/src/session.rs#L440-L444

If you can reproduce this easily, could you check what the sequence TWCC sequence numbers are leading up to this bug?

kvasilye:
Yes I agree that it probably has to do with extending from 16 bit to 64 bit - because a delta of two 16-bit numbers shouldn't produce "10 ^ 20" :)

We'll see about logging sequence numbers - what is a good place to add that in str0m? The file you linked above? Where in that code is the "raw" sequence number from the packet?

algesten:
The file i linked is good. Log Ssrc + unextended seq.

kvasilye:
Found it.

Our TWCC sequence numbers start at a random 16 bit value and sometimes that starting value is >= 32768.

[...]

PR #717, opened 2025-10-29
fix: TWCC last sequence number in empty queue should be None

ant1eicher:
TWCC last sequence number in empty queue should be None - prevents underflow.

Addresses #716


algesten:
Excellent! Thanks!

PR #718, opened 2025-10-30
Expose VP8 and h264 (De)Packetizers unversioned

k0nserv:
This adds a new opt-in module for unversioned API surface and exposes the H264 and VP8 depacketizers under it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -935,11 +935,12 @@ pub enum Output {
 }
 
 /// The reason for the next [`Output::Timeout`].
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
 pub enum Reason {
     /// No timeout scheduled.
     ///
     /// The timeout value is in the distant future.
+    #[default]
     NotHappening,
 
     /// The DTLS subsystem.
@@ -1008,12 +1009,6 @@ pub enum Reason {
     ///
     /// Calculations regarding sender bandwidth using incoming TWCC.
     Bwe,
-}
-
-impl Default for Reason {
-    fn default() -> Self {
-        Self::NotHappening
-    }
 }
 
 impl Rtc {


src/packet/h264.rs
```diff
@@ -14,6 +14,11 @@ pub struct H264CodecExtra {
 }
 
 /// Packetizes H264 RTP packets.
+///
+/// ## Unversioned API surface
+///
+/// This struct is not currently versioned according to semver rules.
+/// Breaking changes to this trait may be made in minor or patch releases.
```
thomaseizinger: Nit: Copy paste error from the trait.

thomaseizinger (commented):
If you only need access to the types and not hook into the actual functionality, you could also move them into a different crate and have people depend on that instead.

algesten:
> If you only need access to the types and not hook into the actual functionality, you could also move them into a different crate and have people depend on that instead.

This is a bit of a compromise. The thinking is like this:

* str0m primarly works on "media level". Thus, RTP packetization is a detail necessary for WebRTC, and not public API.
* RTP mode is necessary. This is how you get the lowest possible latency when str0m is used in a relay. However RTP mode does not require us to expose any packetization API.
* Exposing more API surface, means there's more to maintain and it locks us down.

If we split the packetizers out to a separate crate, that doesn't really help in terms of keeping the API surface down. The `unversioned` is maybe better from that perspective.

thomaseizinger:
> If we split the packetizers out to a separate crate, that doesn't really help in terms of keeping the API surface down. The `unversioned` is maybe better from that perspective.

It does allow you to more liberally bump the version of that crate without having to worry about bumping str0m's. You can update that internally for everyone using the `Rtc` API.

The price to pay there is a bit more overhead in terms of managing and releasing crates.

src/packet/h264.rs
k0nserv: Nice catch, thanks!

algesten:
@thomaseizinger Yeah, I have that setup with ureq-proto vs ureq, but it is a bit painful to have many crates. I think on balance we go with this for now.

PR #727, opened 2025-11-09
Replace dimpl/RustCrypto with aws-lc-rs

algesten:
This only kicks in with the feature `dimpl`.


xnorpx:
lgtm, but I would expect it to have been

```rust
        match self {
            SrtpCrypto::OpenSsl(v) => Box::new(v.new_aes_128_cm_sha1_80(key, encrypt)),
            SrtpCrypto::WinCrypto(v) => Box::new(v.new_aes_128_cm_sha1_80(key, encrypt)),
            SrtpCrypto::Dimpl(v) => Box::new(v.new_aes_128_cm_sha1_80(key, encrypt)),
        }
```

But maybe we need to rework these abstractions.

xnorpx (approved):

algesten:
> ```rust
>             SrtpCrypto::Dimpl(v) => Box::new(v.new_aes_128_cm_sha1_80(key, encrypt)),
> ```

Right. So, there isn't a connection between dimpl and the SRTP cryptos. Dimpl is _just_ DTLS implementation. I think that could mean two things:

* `dimpl` as a feature name is bad, because we mean `dimpl` + _some SRTP impl_
* We should make the DTLS and SRTP completely separate flags, you combine anyway you want.

PR #730, opened 2025-11-10
Reallocate sendonly conflicting PTs

algesten:


src/format.rs
```diff
@@ -475,6 +511,43 @@ impl PayloadParams {
...
+                    remote_pt = new_pt;
+                } else {
+                    panic!("Exhausted all PT ranges, inconsistent PayloadParam state");
+                }
+            }
+
+            // Check if RTX PT also conflicts
+            if local_is_controlling {
+                if let Some(rtx) = remote_rtx {
+                    if claimed.is_claimed(rtx) {
```
k0nserv, line 544: Don't we also have a default pt for rtx we should try here? For example, 97 for VP8 RTX

src/format.rs
```diff
@@ -475,6 +511,43 @@ impl PayloadParams {
...
+                        if let Some(new_rtx) = claimed.find_unclaimed(PREFERED_RANGES) {
```
k0nserv, line 545: Since we don't claim until below don't we need to exclude the pt we chose above for the primary PT here?

k0nserv (approved):
Nice! Just a few things on the RTX assignment

src/format.rs
algesten: That would be a second concern about our default PTs:

1. I haven't checked that our default PTs matches that of Chrome etc in a long while. Are they still OK?
2. I never checked the default RTX, maybe we should?

src/format.rs
algesten: Good point

PR #734, opened 2025-11-11
Fix bug in PT remapping to unlocked PTs

algesten:


src/format.rs
```diff
@@ -1277,4 +1305,265 @@ mod test {
...
+            None,
+            FormatParams {
+                packetization_mode: Some(1),
+                profile_level_id: Some(0x64001f), // High profile (not offered by Chrome)
+                ..Default::default()
+            },
+        );
+
+        // Verify H264 High starts at its default PT 114 BEFORE Chrome's offer
+        let h264_high_before = config
```
k0nserv, line 1352: This is misleading because `CodecConfig::add_config` doesn't do anything to prevent Pt collisions. Yes H264 high will be mapped to 114, but at this point so is h264 CB RTX.

k0nserv: It seems weird to start this test with a collision in place

src/format.rs
```diff
@@ -480,6 +481,7 @@ impl PayloadParams {
         remote_pts: &[PayloadParams],
         claimed: &mut [bool; 128],
         local_is_controlling: bool,
+        unlocked: &mut HashSet<Pt>,
```
k0nserv, line 484: Isn't this the same data structure as `claimed`? It seems it could be `&mut [bool; 128]`

Related: We could change `claimed` to be `u128` with some bit fiddling and be bit more space constrained

k0nserv (approved):

src/format.rs
algesten: You're right. It should start with `new_with_defaults()`

src/format.rs
algesten: Yes and no. It's _almost_ the same, to the point that I agree we want to make them the same. The problem is the existing code has that `claim_once` logic which means mixing the locked with the unlocked in this structure will not work without extensive work.

This is why I say I want to rewrite this entire thing: https://discord.com/channels/1436725650302959829/1436725651301466284/1437859673163431946

It's a mess, and it needs cleaning up, but that's a separate task. Not this PR.

Issue #731, opened 2025-11-10
NTP timestamp in SenderReport's


kvasilye:
Hello,

The SenderReport structure (defined in sr.rs) uses a `time::Instant` for the ntp_time.

We have a need in our app to read the actual NTP timestamp that came in on the wire - the original 32 bit "seconds" + 32 bit "fraction".

Is there a way to do this right now? I see there is an `InstantExt` which has a function to convert to `u64` (restoring the original) but that trait is not public.

It also seems to me that `Instant` may not be the best data type for the ntp time because:

- Instant is "opaque" - has no value that can be read from it; deliberately does not have an epoch.
- NTP time by comparison - has a well defined value and there is a well defined epoch.

Unless I'm missing something and this is indeed a gap in the API - would you be open to me submitting a PR that changes the type of `ntp_time` to `u64`?

On the receiving side, it would just contain the exact value received from the socket, without having to do a roundtrip conversion.

On the sending time, the code needs to convert to `Instant::now` to the NTP time anyway, regardless of the data type, and it would just happen sooner (when the SenderReport data structure is constructed) rather than later (when it's marshalled  onto the wire).

kvasilye:
PS - I could also crate a brand new type `NtpTime` instead of using a "raw" `u64` - it might be overkill, but I'll be happy to let you decide if that's appropriate.

algesten:
Hi!

The question is why you need the NTP time? Depending on situation, you can't trust the remote peer has a correct idea of time. The only thing you can trust the NTP time for is relative comparisons between consecutive `SenderReport` – thus it's no better or worse than `Instant`.

I wrote this recently, not sure it applies to your situation: https://str0m.zulipchat.com/#narrow/channel/377845-general/topic/Manually.20supply.20RTCP.20packets.2Ftweak.20outgoing.20packets/near/547340060

> Synchronizing many incoming RTP streams is tricky. You need to decide how to slot multiple ideas of time into a common timeline. 
>
> * You could use incoming `ntp_time` from remote SR
> * You could look at `rtp_time` with regards to the `sample_rate`/`clock frequency` of the media.
> * You could look at timestamp when a packet arrives at your network interface
>
> (and probably other ways too).
> 
[...]

PR #735, opened 2025-11-11
Represent NTP as SystemTime in sender reports and xrr

kvasilye:
Fixes https://github.com/algesten/str0m/issues/731

--------------

This is my first time writing Rust code, so I'm ready for comments on what I did wrong :)

### DIFF
--- a/src/util/time_tricks.rs
+++ b/src/util/time_tricks.rs
@@ -136,7 +136,7 @@ impl SystemTimeExt for SystemTime {
         let secs_epoch = secs_ntp - SECS_1900 as f64;
 
         // Duration not allowed to be negative
-        let secs_dur = Duration::try_from_secs_f64(secs_epoch).unwrap_or_else(|_| Duration::ZERO);
+        let secs_dur = Duration::try_from_secs_f64(secs_epoch).unwrap_or(Duration::ZERO);
 
         // Time in SystemTime
         SystemTime::UNIX_EPOCH + secs_dur


algesten:
@kvasilye I think we can simplify a bit here. Let's start in `time_tricks.rs`

Regarding `InstantExt`:

1. Remove `from_ntp_64` and `as_ntp_64`
2. Introduce `to_system_time(&self) -> SystemTime`

The new `SystemTimeExt`:

1. Make `from_ntp_64` and `as_ntp_64`

In SR and XR:

1. `ntp_time` should remain `SystemTime` (not `Option<SystemTime>`)

If you start with those changes as starting point, I believe the rest should just follow through nicely.

kvasilye:
@algesten 

Re: ntp_time should remain SystemTime

What if we fail to parse an incoming u64 ntp time? How would you like to represent that?

algesten:
> What if we fail to parse an incoming u64 ntp time? How would you like to represent that?

How could it fail? It's a number.

kvasilye:
I mean this:

```rust
    fn from_ntp_64(v: u64) -> Option<SystemTime> {
        let secs_ntp = (v as f64) / F32;
        let secs_epoch = secs_ntp - SECS_1900 as f64;

        // Duration not allowed to be negative
        if secs_epoch < 0.0 {
            return None // <<<<< ------- what to do here?
        };

        Some(SystemTime::UNIX_EPOCH + Duration::from_secs_f64(secs_epoch))
    }
```

The check for "secs_epoch < 0.0" is [from here](https://github.com/algesten/str0m/blob/main/src/util/time_tricks.rs#L121) - how do you want to handle it?

Return `SystemTime::UNIX_EPOCH` ?

algesten:
I think very little change from how it is today, including code doc.

On row 128 here we have all we need already:

https://github.com/algesten/str0m/blob/main/src/util/time_tricks.rs#L108-L128

kvasilye:
Updated, now using `SystemTime` instead of `Option<SystemTime>`.

@algesten can you please review again?

Regarding your other comments:

> Regarding InstantExt:
> 
> Remove from_ntp_64 and as_ntp_64
> Introduce to_system_time(&self) -> SystemTime
> 
> The new SystemTimeExt:
> 
> Make from_ntp_64 and as_ntp_64

Was already done.

src/util/time_tricks.rs
```diff
@@ -118,33 +140,28 @@ impl InstantExt for Instant {
...
-        let secs_ntp = secs_epoch + SECS_1900 as f64;
+        let since_epoch = self.duration_since(SystemTime::UNIX_EPOCH);
+        match since_epoch {
+             Ok(value) => {
+                let secs_epoch = value.as_secs_f64();
+                let secs_ntp = secs_epoch + SECS_1900 as f64;
+                (secs_ntp * F32) as u64
+             },
+             Err(_) => 0u64
+        }
```
algesten: Rather than doing a bunch of the logic in the `Ok` branch, I prefer introducing a local variable. Something like this, or it could be a combinator:

```rs
        let since_epoch = match self.duration_since(SystemTime::UNIX_EPOCH) {
            Ok(v) => v,
            Err(_) => Duration::ZERO,
        };
```

src/util/time_tricks.rs
```diff
@@ -118,33 +140,28 @@ impl InstantExt for Instant {
...
+    }
 
-        (secs_ntp * F32) as u64
+    #[cfg(test)]
+    fn to_instant(&self) -> Instant {
+        let diff = self.duration_since(BEGINNING_OF_TIME.1).unwrap_or_else(|_| Duration::from_secs(0));
+        BEGINNING_OF_TIME.0 + diff
```
algesten: I'm not convinced we need `to_instant()`. Let's see if we can fix the tests to use `SystemTime` directly instead.

kvasilye:
Both items addressed

src/rtp/rtcp/mod.rs
```diff
@@ -599,11 +599,16 @@ mod test {
...
         // Ensure ntp_time is not too far off.
-        let abs = if now > now2 { now - now2 } else { now2 - now };
+        let abs = abs_time_delta(now, now2);
         assert!(abs < Duration::from_millis(1));
     }
 
-    fn sr(ssrc: u32, ntp_time: Instant) -> Rtcp {
+    fn abs_time_delta(st1: SystemTime, st2: SystemTime) -> Duration {
+        let delta = if st1 > st2 {  st1.duration_since(st2) } else { st2.duration_since(st1) };
+        delta.unwrap()
```
k0nserv: With Options it's preferable to use `expect` with the message making it clear why `delta` will not be `None`.

It's not super important in test code, but it's a good habit

src/streams/receive.rs
```diff
@@ -579,12 +579,11 @@ impl StreamRx {
         // (SR) packet from source SSRC_n.  If no SR has been received yet,
         // the field is set to zero.
         report.last_sr_time = {
-            let t = self
-                .sender_info
-                .map(|(_, s)| s.ntp_time)
-                .unwrap_or(already_happened());
+            let t64 = match self.sender_info {
```
k0nserv: This can be `self.sender_info.map_or(0u64, |(_, s)| s.ntp_time.as_ntp_64()`

algesten:
The ci says you need to run `cargo fmt`

src/util/time_tricks.rs
```diff
@@ -118,33 +136,19 @@ impl InstantExt for Instant {
...
     }
 
     fn as_ntp_64(&self) -> u64 {
-        let since_beginning_of_time = self.duration_since(BEGINNING_OF_TIME.0);
-
-        let since_epoch = since_beginning_of_time + epoch_to_beginning();
-        let secs_epoch = since_epoch.as_secs_f64();
-
-        let secs_ntp = secs_epoch + SECS_1900 as f64;
+        let secs_epoch = match self.duration_since(SystemTime::UNIX_EPOCH) {
```
k0nserv: This can also use `map_or` as an alternative

src/rtp/rtcp/mod.rs
kvasilye: @k0nserv can you give me a snippet I can just copy paste? This is my first time working with Rust :)

kvasilye: I think it's like this?

```rust
    fn abs_time_delta(st1: SystemTime, st2: SystemTime) -> Duration {
        let delta = if st1 > st2 {
            st1.duration_since(st2)
        } else {
            st2.duration_since(st1)
        };
        delta.expect("In tests, the time delta should not be None")
    }
```

k0nserv: In this case `delta.expect("delta should be absolute")`.

If you are curious you can read more about how to word expect messages here https://doc.rust-lang.org/1.90.0/std/option/enum.Option.html#method.expect

src/util/time_tricks.rs
kvasilye: Used `if let = Some....`

src/streams/receive.rs
kvasilye: Done

src/rtp/rtcp/mod.rs
kvasilye: Done

algesten:
The next lint is `cargo clippy`

kvasilye:
> The next lint is `cargo clippy`

Fixed. Thank you for guiding me Martin.

kvasilye:
@k0nserv - I've updated the code based on your comments. Could you please take a look again?

k0nserv:
Looks good @kvasilye!

kvasilye:
@k0nserv I think github wants you to click the approve button - otherwise it considers your existing comments are blocking.

k0nserv (approved):

algesten:
Nice! Thanks!

PR #748, opened 2025-11-18
fix: off-by-one nack reports

lherman-cs:
NACK is not generated on a packet that is on a NACK report boundary, Last pid + 17 will be lost.

### DIFF
--- a/src/streams/register_nack.rs
+++ b/src/streams/register_nack.rs
@@ -5,8 +5,6 @@ use crate::rtp_::{Nack, NackEntry, ReportList, SeqNo};
 /// Number of out of order packets we keep track of for reports
 const MAX_MISORDER: u64 = 100;
 
-const U16_MAX: u64 = u16::MAX as u64 + 1_u64;
-
 /// The max number of NACKs we perform for a single packet
 const MAX_NACKS: u8 = 5;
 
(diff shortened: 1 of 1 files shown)


algesten:
@davibe you wrote the original code here. That u16::MAX+1 looks very deliberate.

src/streams/register_nack.rs
```diff
@@ -619,4 +615,35 @@ mod test {
...
+        let reports: Vec<_> = reg
+            .nack_reports()
+            .expect("should generate reports")
+            .flat_map(|nack| nack.reports)
+            .collect();
+
+        assert_eq!(reports.len(), 2, "Should have found two NACK entries");
+        assert_eq!(reports[0].pid, 2, "First missing packet should be 2");
+        assert_eq!(
+            reports[0].blp, 0,
```
Razzwan, line 640: blp must not be 0 here. If I understand it correctly, it must contain all packages between 2 and 19

Razzwan (changes requested):

src/streams/register_nack.rs
algesten: I think this PR is correct

lherman-cs: @Razzwan this is the RFC, https://datatracker.ietf.org/doc/html/rfc6642#section-5.1.

Each report contains 17 sequence numbers (1 defined by PID and other 16 by BLP bits). `The BLP allows for reporting losses of any of the 16 RTP immediately following the RTP packet indicated by the PID.` PID represents the base sequence number for the rest of 16 packets defined in BLP.

PR #753, opened 2025-11-21
Constant time HMAC tag equality

k0nserv:
`subtle` was already in the tree due to `sha1`

### DIFF
--- a/src/crypto/srtp.rs
+++ b/src/crypto/srtp.rs
@@ -1,0 +1,529 @@
+use std::fmt;
+
+use self::aes_128_cm_sha1_80::AesKey;
+
+use super::CryptoProvider;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SrtpProfile {
+    #[cfg(feature = "_internal_test_exports")]
+    PassThrough,
+    Aes128CmSha1_80,
+    AeadAes128Gcm,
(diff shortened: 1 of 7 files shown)


lolgesten (approved):

src/crypto/srtp.rs
```diff
@@ -214,7 +216,8 @@ pub mod aes_128_cm_sha1_80 {
     pub fn rtp_verify(key: &[u8], buf: &[u8], srtp_index: u64, cmp: &[u8]) -> bool {
         let roc = (srtp_index >> 16) as u32;
         let tag = crate::crypto::sha1_hmac(key, &[buf, &roc.to_be_bytes()]);
-        &tag[0..HMAC_TAG_LEN] == cmp
+
+        tag[0..HMAC_TAG_LEN].ct_eq(cmp).into()
```
thomaseizinger, line 220: Perhaps we should introduce: `crate::crypto::ct_eq()`?

thomaseizinger: I guess we are already in the crypto module here so doesn't really matter.

thomaseizinger:
Any other cases we can fix while we are at it?

algesten:
> Any other cases we can fix while we are at it?

I bet there is some equality check somewhere in dimpl.

PR #754, opened 2025-11-21
Relate NACK to RTT

### LINKED ISSUE #744: NACK reporter is ineffective in wireless network environments
From my understanding, the current NACK reporter generates a NACK only after at least 33ms (`NACK_MIN_INTERVAL`) have passed since the last NACK. Each missing packet can be reported up to 5 times (`MAX_NACKS`). The NACK window length is determined by `MAX_MISORDER`. When a new highest sequence number arrives, the sliding window advances and drops any incoming packets with
`sequence_number < new_highest_sequence_number - MAX_MISORDER`.

Unfortunately, the current NACK reporter becomes ineffective under several common wireless network conditions such as LTE and 5G.

## Low Jitter and Low Packet Loss

**Assume:**

* jitter ≈ 30 ms
* packet_loss ≈ 1%

In this scenario, the receiver’s jitter buffer is typically small (around 30–50 ms). Because the NACK reporter waits at least 33 ms before sending a NACK, there is already a minimum delay of 33 ms just to notify the sender about a missing packet. In practice, the retransmission will take 33 ms + 1 RTT before it reaches the SFU.

Given the small jitter buffer on the client side, the likelihood that the retransmitted packet becomes unusable by the time it arrives is high.

**Proposal:**
The NACK reporter should send the first NACK as soon as a sequence gap is detected, or at least use the observed jitter as a baseline before applying throttling.

## High RTT and Low Packet Loss

**Assume:**

* RTT ≈ 200 ms
* packet_loss ≈ 1%
* received sequence: P1, P2, _, P4

When the NACK reporter sees a gap after receiving P4, it generates a NACK for P3 in the next report. It then continues sending NACKs for P3 until the packet is received or until the `MAX_NACKS` limit is reached. Once it hits `MAX_NACKS`, the window advances and the packet is treated as permanently lost—even if the retransmission eventually arrives.

This issue occurs whenever:

```
NACK_MIN_INTERVAL * MAX_NACKS < RTT
```

In these cases, the reporter exhausts all retry attempts before the retransmitted packet has any chance of arriving, since the retransmission itself requires at least one RTT.

**Proposal:**
The NACK reporter should be RTT-aware. Instead of using a static `NACK_MIN_INTERVAL`, it could use a dynamic value such as `RTT / 2`, or otherwise base the interval on the estimated RTT.

---

**Expected Outcome**: Less bandwidth used for retransmission, and and the stream should be smoother. Each NACK will have a higher chance to recover, and more responsive to a legitimate packet loss.

---

**Context**: I've been observing this unrecoverable packet loss issues typically while I'm on LTE. I have an SFU implementation with RTP mode and no jitter buffer. The SFU just acts a dumb pipe. This is the actual obfuscated ping result (definitely not a friendly network, but great for an experiment 🙂):

```
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=1 ttl=47 time=132 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=2 ttl=47 time=226 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=3 ttl=47 time=246 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=4 ttl=47 time=206 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=5 ttl=47 time=99.3 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=6 ttl=47 time=97.9 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=7 ttl=47 time=90.0 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=8 ttl=47 time=103 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=9 ttl=47 time=152 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=10 ttl=47 time=282 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=11 ttl=47 time=178 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=12 ttl=47 time=250 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=13 ttl=47 time=98.4 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=14 ttl=47 time=192 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=15 ttl=47 time=96.2 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=16 ttl=47 time=237 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=17 ttl=47 time=363 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=18 ttl=47 time=502 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=19 ttl=47 time=322 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=20 ttl=47 time=228 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=21 ttl=47 time=251 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=22 ttl=47 time=171 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=23 ttl=47 time=193 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=24 ttl=47 time=155 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=25 ttl=47 time=238 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=26 ttl=47 time=157 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=27 ttl=47 time=181 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=28 ttl=47 time=204 ms
64 bytes from 123-77-192-44.example.net (123.77.192.44): icmp_seq=29 ttl=47 time=227 ms

--- example-host ping statistics ---
29 packets transmitted, 29 received, 0% packet loss, time 28035ms
rtt min/avg/max/mdev = 90.028/202.622/502.172/88.494 ms
``` 

---

Please let me know what you guys think or correct my understanding. I'm happy to make the contribution too!

algesten:
Close #744

### DIFF
--- a/src/streams/mod.rs
+++ b/src/streams/mod.rs
@@ -411,6 +411,7 @@ impl Streams {
         medias: &[Media],
         config: &CodecConfig,
         feedback: &mut VecDeque<Rtcp>,
+        twcc_rtt: Option<Duration>,
     ) {
         self.mids_to_report.clear(); // Clear for checking StreamRx.
         for stream in self.streams_rx.values() {
@@ -428,7 +429,7 @@ impl Streams {
                 stream.create_rr_and_update(now, sender_ssrc, feedback);
             }
 
-            stream.maybe_create_nack(now, sender_ssrc, feedback);
+            stream.maybe_create_nack(now, sender_ssrc, feedback, twcc_rtt);
 
             stream.handle_timeout(now);
         }
--- a/src/streams/mod.rs
+++ b/src/streams/mod.rs
@@ -408,7 +408,6 @@ impl Streams {
         &mut self,
         now: Instant,
         sender_ssrc: Ssrc,
-        do_nack: bool,
         medias: &[Media],
         config: &CodecConfig,
         feedback: &mut VecDeque<Rtcp>,
@@ -429,9 +428,7 @@ impl Streams {
                 stream.create_rr_and_update(now, sender_ssrc, feedback);
             }
 
-            if do_nack {
-                stream.maybe_create_nack(sender_ssrc, feedback);
-            }
+            stream.maybe_create_nack(now, sender_ssrc, feedback);
 
             stream.handle_timeout(now);
         }
@@ -687,6 +684,11 @@ impl Streams {
             self.any_nack_active = Some(self.streams_rx.values().any(|s| s.nack_enabled()));
         }
         self.any_nack_active.unwrap()
+    }
+
+    pub(crate) fn nack_at(&self) -> Option<Instant> {
+        // Return the earliest NACK time across all streams
+        self.streams_rx.values().filter_map(|s| s.nack_at()).min()
     }
 
     fn rx_lookup_at(&self) -> Instant {
--- a/src/streams/mod.rs
+++ b/src/streams/mod.rs
@@ -404,10 +404,12 @@ impl Streams {
         !self.streams_rx.is_empty()
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub(crate) fn handle_timeout(
         &mut self,
         now: Instant,
         sender_ssrc: Ssrc,
+        do_nack: bool,
         medias: &[Media],
         config: &CodecConfig,
         feedback: &mut VecDeque<Rtcp>,
@@ -429,7 +431,9 @@ impl Streams {
                 stream.create_rr_and_update(now, sender_ssrc, feedback);
             }
 
-            stream.maybe_create_nack(now, sender_ssrc, feedback, twcc_rtt);
+            if do_nack {
+                stream.maybe_create_nack(now, sender_ssrc, feedback, twcc_rtt);
+            }
 
             stream.handle_timeout(now);
         }
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -686,6 +686,13 @@ impl StreamRx {
             return None;
         }
 
+        // Don't generate NACKs before scheduled time
+        if let Some(nack_at) = self.nack_at {
+            if now < nack_at {
+                return None;
+            }
+        }
+
         let rtt = twcc_rtt.unwrap_or(NACK_DEFAULT_RTT);
 
         let nacks = self
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -393,6 +393,7 @@ impl StreamRx {
         clock_rate: Frequency,
         is_repair: bool,
         seq_no: SeqNo,
+        twcc_rtt: Option<Duration>,
     ) -> RegisterUpdateReceipt {
         self.last_used = now;
 
@@ -423,12 +424,7 @@ impl StreamRx {
 
         // Update nack_at when gaps are detected (after releasing register borrow)
         if should_update_nack_at {
-            // Convert stats.rtt (f32 ms) to Duration, or use default
-            let rtt = self
-                .stats
-                .rtt
-                .map(|ms| Duration::from_secs_f32(ms / 1000.0))
-                .unwrap_or(NACK_DEFAULT_RTT);
+            let rtt = twcc_rtt.unwrap_or(NACK_DEFAULT_RTT);
             let register = self.register.as_ref().unwrap();
             if let Some(next_time) = register.next_nack_time(now, rtt) {
                 // If there are missing packets, update nack_at
@@ -683,18 +679,14 @@ impl StreamRx {
         now: Instant,
         sender_ssrc: Ssrc,
         feedback: &mut VecDeque<Rtcp>,
+        twcc_rtt: Option<Duration>,
     ) -> Option<()> {
         if !self.nack_enabled() {
             self.nack_at = None;
             return None;
         }
 
-        // Convert stats.rtt (f32 ms) to Duration, or use default
-        let rtt = self
-            .stats
-            .rtt
-            .map(|ms| Duration::from_secs_f32(ms / 1000.0))
-            .unwrap_or(NACK_DEFAULT_RTT);
+        let rtt = twcc_rtt.unwrap_or(NACK_DEFAULT_RTT);
 
         let nacks = self
             .register
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -17,22 +17,8 @@ use super::register::ReceiverRegister;
 use super::StreamPaused;
 use super::{rr_interval, RtpPacket};
 
-/// Default RTT when no RTT measurement is available yet.
-/// Using 33ms to match the original NACK_MIN_INTERVAL for backwards compatibility
-/// with existing systems that expect this timing.
-const NACK_DEFAULT_RTT: Duration = Duration::from_millis(33);
-
-/// Maximum RTT used for NACK timing. We cap the RTT to ensure NACK
-/// retries can happen within a reasonable timeframe, even on high-RTT
-/// networks. 50ms provides a balance between avoiding duplicate NACKs
-/// and ensuring quick recovery under cellular conditions.
-const NACK_MAX_RTT: Duration = Duration::from_millis(50);
-
-/// Compute the RTT to use for NACK timing.
-/// Uses measured RTT if available, defaults to NACK_DEFAULT_RTT, and caps at NACK_MAX_RTT.
-fn nack_rtt(twcc_rtt: Option<Duration>) -> Duration {
-    twcc_rtt.unwrap_or(NACK_DEFAULT_RTT).min(NACK_MAX_RTT)
-}
+/// Default RTT when no RTT measurement is available yet (matching libWebRTC's kDefaultRtt).
+const NACK_DEFAULT_RTT: Duration = Duration::from_millis(100);
 
 /// Incoming encoded stream.
 ///
@@ -679,7 +665,7 @@ impl StreamRx {
             return None;
         }
 
-        let rtt = nack_rtt(twcc_rtt);
+        let rtt = twcc_rtt.unwrap_or(NACK_DEFAULT_RTT);
 
         // nack_report uses RTT-aware timing internally to decide which packets need NACKs
         let nacks = self
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -17,8 +17,22 @@ use super::register::ReceiverRegister;
 use super::StreamPaused;
 use super::{rr_interval, RtpPacket};
 
-/// Default RTT when no RTT measurement is available yet (matching libWebRTC's kDefaultRtt).
-const NACK_DEFAULT_RTT: Duration = Duration::from_millis(100);
+/// Default RTT when no RTT measurement is available yet.
+/// Using 33ms to match the original NACK_MIN_INTERVAL for backwards compatibility
+/// with existing systems that expect this timing.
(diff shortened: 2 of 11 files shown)


src/streams/mod.rs
```diff
@@ -675,6 +672,11 @@ impl Streams {
         self.any_nack_active.unwrap()
     }
 
+    pub(crate) fn nack_at(&self) -> Option<Instant> {
+        // Return the earliest NACK time across all streams
```
Razzwan, line 695: Why earliest and not latest? According to the logic of this code, it should return the time of the last requested NACK, not the first one. Or am I misunderstanding something?

src/streams/register_nack.rs
```diff
@@ -196,6 +245,35 @@ impl NackRegister {
...
+
+        // Iterate over u64 values, then convert back to SeqNo
+        for s_u64 in **start..=**end {
+            let packet = self.packet(s_u64.into());
+            if packet.received || packet.nack_count >= MAX_NACKS {
+                continue;
+            }
+
+            let need_at = if let Some(sent_at) = packet.sent_at {
+                sent_at + rtt
```
Razzwan, line 297: Maybe we should add a little more time here. I mean, is RTT considered as the largest possible round trip time or is it just the last round trip time? If it's the last round trip time, then it would make sense to multiply it by 1.2, for example: `sent_at + Duration::from_secs_f64(rtt.as_secs_f64() * 1.2)`

src/streams/register_nack.rs
```diff
@@ -42,34 +55,52 @@ impl PacketStatus {
...
 
-        self.reg.packet_mut(self.next.into()).nack_count += 1;
+        self.reg.packet_mut(self.next.into()).mark_nacked(self.now);
         self.next += 1;
 
         for (i, s) in (self.next..self.end).take(16).enumerate() {
             let packet = self.reg.packet_mut(s.into());
-            if packet.needs_nack() {
-                self.reg.packet_mut(self.next.into()).nack_count += 1;
+            if packet.needs_nack(self.now, self.rtt) {
```
Razzwan, line 98: Do I correctly understand, that if `packet.needs_nack() == false` then we have to send `Pli` here?

tests/nack.rs
```diff
@@ -290,8 +290,18 @@ pub fn nack_delay() -> Result<(), RtcError> {
 
     let first_nack_tx = nacks_tx.first().expect("nack");
 
+    // First NACK should be sent within ~20-30ms (when gap is detected after next packet arrives)
```
Razzwan, line 308: Shouldn't it be sent immediately? Why ~20-30ms?

alexlapa: I think that waiting a little bit is a good thing cause that should help to mitigate cases when gap is occuring due to packet reordering. Small reordering is a pretty usual thing even when there is no packet loss at all.

algesten: libWebRTC has one more mechanism here. They record continuous "histograms" of how out-of-order incoming packets are. So they have a way to predict whether packets typically are 1-2-3-4 out of order etc. That way they know exactly how long to wait.

src/streams/register_nack.rs
alexlapa: PLIs are meant to be sent when a reference frame can not be assembled and NACKer does not really know if loss breaks the decoder chain or not. That would be the job of depacketizer/decoder.

The only case when NACKer can probably send PLI is when pending NACKs list is too big and NACKer decides to reset itself clearing the list.

src/streams/mod.rs
algesten: We return earliest because the poll loop needs to wake at the soonest deadline. If stream A needs NACK at 100ms and stream B at 200ms, we wake at 100ms to not miss A's deadline.

src/streams/register_nack.rs
algesten: Valid point. libWebRTC uses straight RTT without multiplier - philosophy is slightly early retry (bandwidth cost) beats late retry (latency cost). Could add variance tracking later if needed.

tests/nack.rs
algesten: Also, this is a test, the actual send is immediate. The ~20-30ms is when the gap is *detected* (when subsequent packet arrives), not intentional delay. First NACK fires instantly upon detection, retransmissions wait for RTT.

Razzwan (approved):

lherman-cs:
I've been testing this new patch locally with a simulated network that is close to the lte network here, https://github.com/algesten/str0m/issues/744.

```bash
        sudo tc qdisc add dev lo root netem \
                rate 3mbit \
                delay 45ms 35ms distribution pareto \
                loss 0.15%; \
```


**The test setup:**

```
WHIP Publisher -> SFU -> WHEP Subscriber
```

* WHIP Publisher publishes 1 video without simulcast, and uses a pre-recorded video. https://codepen.io/lherman-cs/pen/xbVYqrJ
* WHEP Subscriber: https://codepen.io/lherman-cs/pen/pvgVZar
* SFU: RTP Mode as a dumb pipe, no buffering.

**Result:**

fwiw, the result is indefinitive, maybe due to too many non-deterministic parts in the tests. I can't really say one is better than the other. But, I noticed that the SFU now reports less NACKs to the sender.

The main metrics I looked at were: Sender's NACK & PacketLost, and Receiver's PacketLost. 

Before:

* Sender's NACK: ~38 NACKs
* Sender's PacketLost: ~2
* Receiver's PacketLost: ~10

After:

* Sender's NACK: ~20NACKs
* Sender's PacketLost: 0
* Receiver's PacketLost: ~11

Note: I tried to stop the stream at a similar duration. But, this could've been done better with a shorter video..

**Before**

sender:

before-sender

receiver:

before-receiver



**After**

sender:

after2-sender

receiver:

after2-receiver

src/streams/receive.rs
```diff
@@ -377,16 +384,42 @@ impl StreamRx {
...
+            // Check if we should update nack_at (only for main register, not RTX)
+            should_update_nack_at = !is_repair && self.nack_enabled();
+        }
+
+        // Update nack_at when gaps are detected (after releasing register borrow)
+        if should_update_nack_at {
+            // Convert stats.rtt (f32 ms) to Duration, or use default
+            let rtt = self
+                .stats
+                .rtt
```
lherman-cs: Since the nack reporter now ties to RTT stats, should we enable stats by default in config?

algesten: @davibe @k0nserv why did we make stats optional in the first place?

If we set the default time in case RTT isn't there to 33ms, will we be back to roughly the old behavior then?

k0nserv: I think it's mostly because there's a cost to having the stats. Ultimately the RTT in the stats come from TWCC so we can still have RTT without stats being enabled, just grab it from the TWCC receive register

lherman-cs:
@algesten  just ran another run with real network this time, and statistic interval set to 200ms. I noticed that the packet lost is higher with this PR as observed in remote-inbound-rtp stats. But, it looks like it recovers the packets in the next receiver reports. The original NACK reporter seems to be more responsive as there are less packet loss reported in the receiver reports.

The CPU usage seems higher? I sometimes observed CPU spikes in some occasions. Looking at the flamegraph, "do_poll_output" is a bit more expensive, and it seems to create more timer wheel churns.

My rtc loop looks more or less like this:

```rust
let rtc_timer = tokio::time::sleep_until(current_deadline);
tokio::pin!(rtc_timer);

loop {
    let Some(new_deadline) = self.core.poll_rtc() else {
        break;
    };
    if new_deadline != current_deadline {
        rtc_timer.as_mut().reset(new_deadline);
        current_deadline = new_deadline;
    }
    
    tokio::select! {...}
}
````

**original nack reporter**
nack-original


**this PR nack reporter**
nack-rtt

algesten:
@lherman-cs thanks for testing!

There's a bug in this PR that would have caused a constant incorrect NACK reporting. Would you mind testing again?

> But, it looks like it recovers the packets in the next receiver reports.

I might be looking at the wrong thing, but to me these graphs look like an improvement. Instead of hovering around a constant 4 packets lost, we are going to 0-1?

Before
image

After
image

> The CPU usage seems higher? I sometimes observed CPU spikes in some occasions. Looking at the flamegraph, "do_poll_output" is a bit more expensive, and it seems to create more timer wheel churns.

Yeah, it's doing a lot more now, so CPU would be higher. If we can get the functionality correct, we can look at how to optimize it.

algesten:
@lherman-cs i also found a bug that meant nack reports were sent at times they shouldn't bee (too soon).

I also fixed the RTT being read from TWCC (like @k0nserv mentioned).

src/streams/receive.rs
algesten: Done

lherman-cs:
> @lherman-cs thanks for testing!
> 
> There's a bug in this PR that would have caused a constant incorrect NACK reporting. Would you mind testing again?
> 
> > But, it looks like it recovers the packets in the next receiver reports.
> 
> I might be looking at the wrong thing, but to me these graphs look like an improvement. Instead of hovering around a constant 4 packets lost, we are going to 0-1?
> 
> Before image
> 
> After image
> 
> > The CPU usage seems higher? I sometimes observed CPU spikes in some occasions. Looking at the flamegraph, "do_poll_output" is a bit more expensive, and it seems to create more timer wheel churns.
> 
> Yeah, it's doing a lot more now, so CPU would be higher. If we can get the functionality correct, we can look at how to optimize it.

Ah, this is my bad. I forgot to mention that the stats from the original probably got messed up at ~1:03:30PM, since I started a profiler at this moment. 

From my perspective, there are 2 ideal behaviors that would lead to improved end-to-end UX (from publisher to subscriber):
1.  Only generate NACK reports when necessary: This will reduce network pressure for the publisher, allowing more bandwidth used for the actual streams, aka higher goodput. This can be observed through a proxy metric from candidate-pair's availableOutgoingBandwidth. 
2. No reported packet lost in remote-inbound-rtp/rtcp receiver report from sfu: I typically treat sustained packet lost and temporary packet lost as equal. Either case, the packet recovery is probably too slow if the packet lost gets reported since each receiver report takes ~1 second interval for video, https://github.com/PulseBeamDev/str0m/blob/9b28349eb828410a40b038888efc7b40e00fcd10//src/streams/mod.rs#L35-L36. 

I considered the original to perform better because there are fewer packet lost reported overall, which implies a higher success rate in recovering the stream.

lherman-cs:
@algesten I still see similar packet lost reports from the latest change here unfortunately. The RTT graph is very noisy. This makes me think that this is related to the rtt calculation. Maybe we need a windowed histogram, and take P50 or even P30?

**original**

nack3-original

**rtt**

nack3-rtt

PR #758, opened 2025-11-24
Update H.264 profile and level matching logic to comply with RFC 6184

### LINKED ISSUE #757: h264 Profile Level Matching
Quick note, I'm not an h264 expert so if I'm wrong on this please let me know.

I'm running into an issue in which an incoming SDP Offer is rejected by str0m when using an h264 profile `level_idc` lower than the one specified in the supported list. For example the default `enable_h264()` method adds `0x42e01f` (`0x1f` = Level 3.1), but if I have an SDP Offer with `profile-level-id=0x42e01e` (`0x1e` = Level 3.0) it is rejected.

Looking at [RFC 6184 Section 8.2.2](https://datatracker.ietf.org/doc/html/rfc6184#section-8.2.2) it appears that a lower level should be accepted as long as the profile matches exactly, but str0m is currently performing an exact match of the entire `profile-level-id`. 

The sections of the RFC I'm specifically referencing are:
> These media format configuration parameters (except for the level part of profile-level-id) MUST be used symmetrically

and 

> In SDP Offer/Answer, an answer to an offer may indicate a level equal to or lower than the level indicated in the offer.

I know this isn't exactly the scenario (since the client offer is the lower one), but the idea is that if your decoder supports level 3.1 it also naturally supports all levels less than 3.1. In this case the client Offer is asking for that lower level, therefor I believe str0m (which is set to support a higher level) should Answer to accept that lower level as well.

I have implemented this on a fork of the main branch and it is working correctly, both allowing me to connect a producer at a lower level, and receive correctly on a browser. Included it a test which covers these variations.

LWCoder:
- Updated profile comparison to ensure exact matches.
- Added checks to reject levels exceeding configured capabilities.
- Introduced unit tests for various H.264 profile and level matching scenarios.

FIXES #757

### DIFF
--- a/src/format/payload_params.rs
+++ b/src/format/payload_params.rs
@@ -370,7 +370,13 @@ impl PayloadParams {
             return None;
         }
 
-        Some(100)
+        // Decrement score based on level difference
+        let level_difference: usize = c0_profile_level
+            .level()
+            .ordinal()
+            .saturating_sub(c1_profile_level.level().ordinal());
+
+        Some(100_usize.saturating_sub(level_difference))
     }
 
     pub(crate) fn update_param(
@@ -538,7 +544,7 @@ mod test {
         struct Case {
             c0: CodecSpec,
             c1: CodecSpec,
-            must_match: bool,
+            expected: Option<usize>,
             msg: &'static str,
         }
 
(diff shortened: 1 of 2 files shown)


algesten:
@LWCoder thanks!

I believe this PR could do with a tweak. If you look at the other score calculations, you see how we return a number between 0 and 100 (I think max is 100). The idea is that a better match is given higher priority.

I agree we should match lower profiles the way you describe, but I think we should calculate some kind of score.

A perfect match is 100, anything which matches a lower to a higher, is not 100.

I think like this: If you declare both 3.1 PTa and 3.0 on PTb, and we get an SDP with 3.0 only, you presumably prefer to get the 3.0 match on PTb. Does that sound right?

LWCoder:
@algesten Great point and I absolutely agree.

Just pushed an update which subtracts the level distance between configured and offered from 100. After your comment I checked the opus score method to see how you were doing that. This should give priority to exact matches first (score 100) and then next to the match with the least number of steps between levels. Using your example "3.1 PTa and 3.0 PTb" if we offer level 2.2 then PTa will score 98 and PTb will score 99.

This ended up being more code than I hoped as Level1B has an oddity that it, capability wise, sits between Level 1 and Level 1.1, but numerically is the lowest. To compensate I had to add an `ordinal()` with the correct capability ordering and then implement PartialOrd and Ord to use that instead of numerical values.

All of my tests are updated to check the correct scoring, and I have edge case tests around the Level1B special case.

algesten (approved):
Looks great! Thanks!

PR #760, opened 2025-11-27
New CryptoProvider API

algesten:
# Description

The starting point for this change is that we want One Unified API for providing the crypto needed for str0m. We want it to be possible to plug in the crypto without making any changes to str0m.

`CryptoProvider` is the new Unified API. It consists of 4 things:

* SrtpProvider – the functions needed for SRTP (AEAD GCM, and ECB for key derivation)
* Sha1 HMAC – for STUN auth.
* Sha256 – for certificate fingerprint.
* DTLSProvider – abstraction over DTLS

The `DTLSProvider` is an additional sub-API for DTLS. It looks exactly like `dimpl` (it's the exact 6 functions you find here: https://docs.rs/dimpl/0.1.5/dimpl/struct.Dtls.html (ignore `new`).

To elaborate, there of course is already an [existing internal API for DTLS](https://github.com/algesten/str0m/blob/461c7315ec9c3143a913fb834763c5ecea8c6ea0/src/crypto/dtls.rs#L189-L216) – with associated types. However I don't like this one. It got this extra `handle_handshake` function that is really an internal detail to the DTLS impl. I want to get rid of this API.

Additionally the new `DTLSProvider` re-exports the `dimpl` types like [`Output`](https://docs.rs/dimpl/0.1.5/dimpl/enum.Output.html) – the thinking is that `dimpl` is a clean Sans-IO abstraction over DTLS. We don't need to double up on these types in str0m.

Thus. str0m always depends on `dimpl`, but compiles it without any crypto backend, just to get the types to re-export.

The backends we already have (`openssl`, `wincrypto`, `aws-lc-rs`) are all changed in this PR to fulfill the `CryptoProvider` API. That means that although `openssl` and `wincrypto` does the DTLS itself, it exposes that via the `dimpl` types/API.

For backends such as `aws-lc-rs` that uses `dimpl` not only for the API, but also for the actual DTLS implementation, the DTLSProvider is a thin shim since even the types are the same.

# The future

Move all crypto out of str0m.

I wanted to move at least `wincrypto` out already now, but realised that we rely on Github CI building/testing the entire windows backend. I couldn't find a good way to keep this CI going, so decided to keep the backend as internal – for now… 

That means although the new `CryptoProvider` API is public, the actual implementations of are still inside str0m.

In a future PR I want to refactor out the integration tests to sit in another crate. That way we can break out the crypto backends to individual crates, such as `str0m-openssl`, `str0m-aws-lc-rs` etc.

### DIFF
--- a/src/crypto/wincrypto/dtls.rs
+++ b/src/crypto/wincrypto/dtls.rs
@@ -2,6 +2,7 @@
 
 use crate::crypto::{CryptoError, DtlsCert, DtlsInstance, DtlsOutput, DtlsProvider};
 use crate::crypto::{KeyingMaterial, SrtpProfile};
+use crate::util::not_happening;
 use std::collections::{HashMap, VecDeque};
 use std::sync::{Arc, LazyLock, Mutex};
 use std::time::Instant;
@@ -54,6 +55,7 @@ impl DtlsProvider for WinCryptoDtlsProvider {
         Ok(Box::new(WinCryptoDtlsInstance {
             dtls,
             pending_outputs: VecDeque::new(),
+            last_timeout: None,
         }))
     }
 }
@@ -65,6 +67,8 @@ impl DtlsProvider for WinCryptoDtlsProvider {
 struct WinCryptoDtlsInstance {
     dtls: str0m_wincrypto::Dtls,
     pending_outputs: VecDeque<PendingOutput>,
+    /// The last time we were given via handle_timeout, used for calculating next timeout.
+    last_timeout: Option<Instant>,
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


Razzwan:
I tried this option today. Any hope that it will work faster than OpenSSL?

algesten:
> I tried this option today. Any hope that it will work faster than OpenSSL?

It's not really that kind of change. The crypto code has been a bit spaghetti since it grown organically. This changes that into a clear `CryptoProvider` struct that can even be implemented outside of str0m.

The `aws-lc-rs` variant should hopefully perform at least as good as `openssl`.

Razzwan:
Why is the feature called '_aws-lc-rs_'?

I'm still learning str0m, but let me know if I can help with this feature somehow.

algesten:
> Why is the feature called '_aws-lc-rs_'?

It's because the crypto provider is this crate: https://crates.io/crates/aws-lc-rs

This mirrors how [rustls](https://crates.io/crates/rustls) does it.

> I'm still learning str0m, but let me know if I can help with this feature somehow.

If you got access to windows, we really want these tests that break to work. In the main-branch they do work, so it should be fairly easy to see where it goes wrong by comparing/println etc.

Razzwan:
> If you got access to windows, we really want these tests that break to work. In the main-branch they do work, so it should be fairly easy to see where it goes wrong by comparing/println etc.

It looks like you've finished developing this feature. My congratulations! The scale of the work is really great!

I have access to all operation systems, but I tried running this encryption `aws-lc-rs` on Linux x64 and it doesn't work even here: the first connection works normally, the second one not happening. No errors. I just see, that second connection not happening in the browser, quite weird, so I didn't get to Windows, sorry

algesten:
> but I tried running this encryption `aws-lc-rs` on Linux x64

Running an example or how did you test it?

xnorpx (approved):
yolo

Razzwan:
@algesten 
I tested it within my SFU version based on str0m. Can try example with that settings

PR #763, opened 2025-11-29
Add apple crypto with dimpl dtls impl

xnorpx:


xnorpx:
Ok, we should try to remove some of the rust deps here in the apple crate, so I will iterate on that.

xnorpx:
Ok I merged it to unblock other changes, we should probably move OpenSSL out and fully move Wincrypto out. We will continue iterate on the apple crypto

PR #769, opened 2025-12-03
Handle SSRC 0 BWE probe

### LINKED ISSUE #547: Implement send and recv support for bitrate_probing on SSRC 0
https://github.com/pion/webrtc/pull/2816

algesten:
Close #547

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Handle incoming SSRC 0 BWE probe #769
   * New CryptoProvider API #760
   * Update H.264 profile and level matching logic #758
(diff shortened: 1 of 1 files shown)


algesten:
> Do we really have to create the `Media` I guess we have invariants for received streams that forces us into that?
> 
> I think it would be cleaner if we could just have orphaned streams that aren't hooked up to any media.

Good point. Let's explore that.

algesten:
> Do we really have to create the `Media` I guess we have invariants for received streams that forces us into that?

It seems we can get away with that. It breaks my mental invariants, but I suppose that's okay 🙄

algesten:
@xnorpx NB

PR #771, opened 2025-12-03
Send SSRC 0 BWE probe

algesten:
- **Send SSRC 0 BWE probe**
- **Use non-zero BWE target before first estimate**


k0nserv (approved):

algesten:
I have thoughts on how to do this with BWE V2. Let's make a new PR when it's time.

PR #774, opened 2025-12-04
Network emulation for tests

algesten:


netem/src/config.rs
```diff
@@ -0,0 +1,464 @@
+use std::time::Duration;
+
+/// Probability in range 0.0..=1.0
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub struct Probability(pub f32);
```
k0nserv: Should this be below the main `Netemconfig´?

netem/src/lib.rs
```diff
@@ -0,0 +1,535 @@
...
+
+    /// The packet data.
+    data: T,
+
+    /// Ever increasing counter to break ties when send_at is the same.
+    packet_index: u64,
+}
+
+/// Sans-IO network emulator.
+pub struct Netem<T> {
```
k0nserv, line 62: Should this struct be first in the file?

netem/src/lib.rs
```diff
@@ -0,0 +1,535 @@
...
+                self.timeout_pending = false;
+            }
+            Input::Packet(now, data) => {
+                self.progress_time(now);
+                self.process_packet(now, data);
+            }
+        }
+    }
+
+    fn progress_time(&mut self, now: Instant) {
```
k0nserv, line 129: This isn't an invariant that warrants panic?

netem/src/lib.rs
```diff
@@ -0,0 +1,535 @@
...
+
+/// Sans-IO network emulator.
+pub struct Netem<T> {
+    config: NetemConfig,
+    rng: Rng,
+    loss_state: LossState,
+
+    /// Priority queue of packets, ordered by send time (earliest first).
+    /// Using Reverse to make BinaryHeap a min-heap.
+    queue: BinaryHeap<Reverse<QueuedPacket<T>>>,
```
k0nserv, line 69: There's one small issue with this approach and reordering. When we reorder we ignore queuing due to link overuse. 

I think this could be solved by inverting `QueuedPacket::send_at` and `QueuedPacket::packet_index` in the sort order and then changing how we set `packet_index` to

```rust
        let packet_index =
            (self.packet_count + 1) * self.config.reorder_gap.map(|g| g as u64).unwrap_or(1);
```

Then we can set `send_at` to the value from the packet we are reordering ahead of. When we want to reorder a packet we set `packet_index` to 

```rust
let expected_index = (self.packet_count + 1) * gap as u64;
let packet_index = expected_index -  (self.reorder_counter - 1 as u64) * gap as u64 - 1;
```

With this and a reorder gap of 3 we'd get the following packet indexes:

```
3(t=0) 6(t=5) 2(t=0) 12(t=15) 15(t=20) 11(t=15)
```

netem/src/config.rs
algesten: Yes

netem/src/lib.rs
algesten: I've been a bit loose on that in the rest of the code. Sometimes opting for warn.

netem/src/lib.rs
algesten: The reason I had to introduce the `packet_index` was two-fold.

1. The `eq` must not overwrite packets. If they appear at the same time (which they do with the openssl backend), they were overwriting.
2. Fiddling with `send_at` times to "space out" packets is not working very well with windows. I don't exactly understand the resolution we are working with, but it seems moving packets with `microsecond` is not sufficient to make an `Instant` sufficiently different. It appears that 1-2ms is the smallest value I can use.

We could potentially drop the `BinaryHeap` in favor of a `VecDeque` because we don't expect incoming packets to be unordered on the input. The only reason not insert a packet last is to deliberately cause "a jumble" (for some reason of the emulation), and that's simple enough to find the insertion point also with the queue.

> When we reorder we ignore queuing due to link overuse.

Could you expand on what you mean with this?

netem/src/lib.rs
algesten: Yes

netem/src/lib.rs
k0nserv: > The reason I had to introduce the packet_index was two-fold.

>    The eq must not overwrite packets. If they appear at the same time (which they do with the openssl backend), they were overwriting.
>    Fiddling with send_at times to "space out" packets is not working very well with windows. I don't exactly understand the resolution we are working with, but it seems moving packets with microsecond is not sufficient to make an Instant sufficiently different. It appears that 1-2ms is the smallest value I can use.

> We could potentially drop the BinaryHeap in favor of a VecDeque because we don't expect incoming packets to be unordered on the input. The only reason not insert a packet last is to deliberately cause "a jumble" (for some reason of the emulation), and 
that's simple enough to find the insertion point also with the queue.

Cool, I think that all makes sense. `BinaryHeap` seems appropriate to me.


> Could you expand on what you mean with this?
The way we move packets to the front for reordering is by setting `send_at` to `now`. However, this causes reordering to ignore the rate limiting.

For example at `t=10` if we queue a packet that should be reordered with the current queue state `10(t=49), 11(t=69)` we'll get the queue state `12(t=10), 10(t=49), 11(t=69)` which means we'll send the packet with index `12` immediately, rather than at time `t=49` (which is more correct).

With my suggestion we'd instead get the initial queue state `9(t=49), 12(t=69)` and then when we queue the reorder packet we'd get the queue state `8(t=49), 9(t=49), 12(t=69)`. We'd correctly delay sending the reordered packet until t=49. We pull out packet 8 first because we invert the ordering. It's pretty minor however

algesten: Got ya. The reordering logic is clashing with the bandwidth limiting logic, undoing the latter.

xnorpx: Reorder is very rare from what we see from telemetry.

xnorpx: If you have loss and jitter (that don't cause reorder) is probably enough as a start.

algesten: Right. So the jitter would need a fix to ensure the order is preserved.

xnorpx: Yeah my mental model of it is it builds up in queue and it release you get trains of packets (without reorder). So first packet jitter large then it decreases etc.

Issue #779, opened 2025-12-08
Q: RFC 8851 for simulcast offer


kvasilye:
Hello,

We use str0m to send simulcast video to an SFU.

The SDP offer represents simulcast layers like this:

```
 a=rid:high send
 a=rid:medium send
 a=rid:low send
```

We'd like the SDP to also contain video stream attributes - bitrate, width, height, frame rate - like this:

```
 a=rid:high send max-width=1280;max-height=720;max-fps=30;max-br=1500000
 a=rid:medium send max-width=640;max-height=360;max-fps=30;max-br=800000
 a=rid:low send max-width=320;max-height=180;max-fps=15;max-br=250000
```

The syntax and the keys above are [defined by RFC 8851](https://datatracker.ietf.org/doc/html/rfc8851#name-arid-restrictions). I believe str0m already supports it when parsing SDP answers, but I'm not sure about the capability to generate SDP offers like this.

Is this possible, and how?

If this is not currently possible, would you be open to us submitting a PR which implements this feature?

PS - "we" = Amazon Interactive Video and we already worked with you [on this pull request](https://github.com/algesten/str0m/pull/735) :)

algesten:
> Is this possible, and how?

It's not possible, no. Despite using simulcast, we don't use these values ourselves.

> If this is not currently possible, would you be open to us submitting a PR which implements this feature?

Yeah, sure, PR welcome! It would be a breaking change for the `Simulcast` which is fine. Since we're talking about a bunch of optional attributes, I suggest turning `Simulcast` struct into a builder pattern (or the `send`/`recv` parts).

For SDP API, I'm a big believer in that str0m needs to be as user friendly as possible, so it's key we nail this API in the most approachable way.

We should probably also indicate in doc that these values are not enforced by str0m. The sender of the simulcast could break them.

We're currently going through a big code churn in main regarding refactoring crypto backends, BWE tests etc. I don't have a timeline for releasing that. My preference is that you fork off 0.12 and make the PR against that. I will release with this as 0.13 (with maybe some cherry-picks from main).

> PS - "we" = Amazon Interactive Video and we already worked with you

It's not the royal "we", but a collective voice, check :)

kvasilye:
Thank for you being open to a PR!

Yes this would be optional because 1) we don't want to burden all applications with having to provide these attributes 2) we don't want to send these attributes by default in cases there are  buggy SFUs which can't handle them.

[...]

PR #780, opened 2025-12-09
RFC 8851 attributes for Simulcast

kvasilye:
Fix https://github.com/algesten/str0m/issues/779

Allow applications to specify additional attributes for simulcast layers.

Those attributes are then propagated into the SDP offer per RFC 8851.

Off by default to preserve compatibility and avoid placing an additional configuration burden on the application.

Default layer configuration:

```
 a=rid:high send
 a=rid:medium send
 a=rid:low send
```

Configuration with attributes:

```
 a=rid:high send max-width=1280;max-height=720;max-fps=30;max-br=1500000
 a=rid:medium send max-width=640;max-height=360;max-fps=30;max-br=800000
```

This required an API change in `SdpApi::add_media` or specificaly in the `Simulcast` structure. The layers are now `SimulcastLayer` structures instead of just `Rid`'s.

A layer can be crated from a Rid name using `SimulcastLayer::new(rid: &str)` which just has the name without any options / attributes. Or you could use `SimulcastLayer::new_with_attributes` passing in a `SimulcastLayerAttributes` structure.

The `SimulcastLayerAttributes` structure looks like this:

```rust
/// Layer attributes, per RFC 8851. There are more attributes in the RFC, but we've chosen these
/// for practical reasons - they are the ones in the Google VLA extension, so an application publishing
/// a simulcast stream has to provide these, to the SDP and also to the VLA.
#[derive(Debug, PartialEq, Eq, Clone)]
pub struct SimulcastLayerAttributes {
    /// Maximum video width, or 0 if none
    pub max_width: u32,
    /// Maximum video height, or 0 if none
    pub max_height: u32,
    /// Maximum bitrate, in bits per second (not kilobits), or 0 if none
    pub max_br: u32,
    /// Maximum frame rate, in frames per second, or 0 if none
    pub max_fps: u32,
}

```

Implementation question:

I've created structures for layer attributes in the `change::sdp` crate and also the `media::event` crate with conversion functions back and forth. This follows the existing pattern with how we treat Rid's vs. RestrictionIds.

If this is not necessary we could define the structure once (in the sdp crate I think) and re-export it from the event crate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Simulcast send attributes RFC 8851 #780
+
 # 0.12.0
 
   * Bump str0m-wincrypto to 0.2.1


algesten:
Hey, I came down with a bad cold, so my head is a mush and the world is spinning. I hope to read this in detail later this week.

kvasilye:
Thank you! Get well!

src/change/sdp.rs
```diff
@@ -910,11 +905,14 @@ fn add_pending_changes(session: &mut Session, pending: Changes) {
...
+        let layers = add_media.simulcast.map(|x| x.send).unwrap_or(vec![]);
 
         for (i, (ssrc, rtx)) in add_media.ssrcs.into_iter().enumerate() {
-            let maybe_rid = rids.get(i).cloned();
-            let midrid = MidRid(add_media.mid, maybe_rid);
+            let maybe_layer = layers.get(i).cloned();
+            let midrid = match maybe_layer {
+                Some(layer) => MidRid(add_media.mid, Some(layer.rid)),
+                None => MidRid(add_media.mid, None),
+            };
```
algesten: This looks like a `maybe_layer.map` instead of the `match`.

src/change/sdp.rs
```diff
@@ -1304,15 +1285,38 @@ impl AsSdpMediaLine for Media {
         }
 
         if let Some(s) = self.simulcast() {
+            fn to_attributes(
+                attributes: Option<SimulcastLayerAttributes>,
+            ) -> Vec<(String, String)> {
+                if let Some(attributes) = attributes {
+                    let mut attrs = vec![];
+                    if attributes.max_width > 0 {
```
algesten: This looks like we're using 0 as a sentinel value instead of `Option` – is that right?

src/change/sdp.rs
kvasilye: Fixing, thanks

src/change/sdp.rs
kvasilye: Yes, I thought that using an `Option` for every single item is overkill

kvasilye: Regarding the builder pattern for `Simulcast` - it would require the application to create a mutable `Simulcast` struct so it can be modified (layers added).

Is this what we want?

Maybe we just use the builder pattern for SimulcastLayer and its attributes?

kvasilye:
@algesten I updated the PR:

- Use the builder pattern for constructing layers
- The builder has functions for the same four attributes as my previous iteration and a function to specify custom attributes
- Use a map<string, string> for storing attributes instead of a custom structure
- We don't have to treat 0 as a None anymore - if a value is not specified, it just won't be in the map

kvasilye:
In the latest commit (as of now) I added a fully fluent interface for building `Simulcast`. To my eyes it maybe looks too verbose, but it's your decision.

If you prefer the partially fluent interface (where `SimulcastLayer` uses a builder but `Simulcast` does not) - I'll just revert that commit.

If you prefer the fully fluent interface (where `Simulcast` itself uses a builder) - I'll keep the commit.

algesten:
Released in 0.13

kvasilye:
Thank you!

PR #787, opened 2025-12-18
openssl: queue data until handshake completes

xnorpx:
With Openssl backend data channel open event is slow.

thread 'data_channel_direct' (42450) panicked at tests/data-channel-direct.rs:119:5:
ChannelOpen should happen within 250ms of ICE completing, but took 3.052487208s
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * wincrypto: queue data until handshake completes #787
   * openssl: queue data until handshake completes #787
 
 # 0.14.0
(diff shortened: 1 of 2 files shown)


xnorpx:
So burning some claude tokens he suggested that we started polling sctp before dtls was connected.

```rust
// Only poll SCTP when DTLS is connected. If we poll SCTP before DTLS is ready,
        // SCTP will generate transmits (like INIT) that can't be sent yet, causing
        // SCTP's internal retransmission timer to start prematurely.
        while self.dtls_connected {
            let Some(e) = self.sctp.poll() else {
                break;
            };
            match e {
                SctpEvent::Transmit { mut packets } => {
...
```

PR #789, opened 2025-12-18
Format wincrypto toml

xnorpx:
Snowflake was failing on master, this is trying to fix it and add taplo to CI


algesten:
I only added taplo.toml to avoid formatting changes in VSCode locally. Do we really need to have a whole new CI step for it?

lolgesten (approved):

PR #790, opened 2025-12-18
openssl: queue app data until handshake completes

algesten:


algesten:
Close in favor of putting all this in #787

PR #792, opened 2025-12-18
Defer panic on missing crypto provider until Rtc init

algesten:


src/lib.rs
```diff
@@ -1073,10 +1073,20 @@ impl Rtc {
     }
 
     pub(crate) fn new_from_config(config: RtcConfig) -> Result<Self, RtcError> {
+        // If crypto_provider is not set in config, check process default
```
algesten: This comment needs fixing.

src/lib.rs
```diff
@@ -1073,10 +1073,20 @@ impl Rtc {
...
+        let crypto_provider = config
+            .crypto_provider
+            .clone()
+            .or_else(|| CryptoProvider::get_default().cloned().map(Arc::new))
+            .or_else(|| Some(Arc::new(from_feature_flags())))
+            // unwrap is correct, the crypto provider is guaranteed to be set by this point.
+            // from_feature_flags already panics if no crypto backend feature flag was enabled.
+            .unwrap();
```
k0nserv: Use expect and put the reasoning from the comment in the string?

algesten: ok

PR #794, opened 2025-12-18
Fix order of DTLS pending packet poll

algesten:


algesten:
Close in favor of #796

PR #797, opened 2025-12-19
feat: allow defining tcptype ICE candidate role

lherman-cs:
In Firefox, TCP ICE candidates will get rejected if they are not defined as "passive" in the answer SDP. 

I'm not sure about the Candidate API or the TcpType enum placement.

For TcpType enum, I see there are 3 good places:
1. io
2. sdp
3. ice

In this PR, I chose `io` because we already have WebRTC common types there as well, e.g. Protocol, MultiplexKind. And, Protocol enum is already used in sdp and ice as well.

For the candidate API, I wonder if it would make sense to turn it into a builder API? It allows people to define optional extensions like this `tcptype` without breaking the existing APIs. `tcptype` can be defined for any candidate type, I just did the "host" candidate type for getting an early feedback here.

### DIFF
--- a/src/ice/candidate.rs
+++ b/src/ice/candidate.rs
@@ -716,7 +716,7 @@ pub struct HasRoute {
 ///
 /// This builder uses the Type State Pattern to enforce correct construction order and
 /// protocol-specific constraints (such as preventing `tcptype` on UDP).
-pub struct CandidateBuilder<P, S> {
+pub struct CandidateBuilder<P, R> {
     foundation: Option<String>,
     component_id: u16,
     prio: Option<u32>,
@@ -726,7 +726,7 @@ pub struct CandidateBuilder<P, S> {
     local_preference: Option<u32>,
 
     protocol_state: P,
-    route_state: S,
+    route_state: R,
 }
 
 // Step 1: Protocol Selection
@@ -769,10 +769,10 @@ impl CandidateBuilder<NoProtocol, NoRoute> {
     }
 }
 
-// Step 2: Kind Selection
-impl<P> CandidateBuilder<P, NoRoute> {
+// Step 2: Route Selection
+impl<P> CandidateBuilder<HasProtocol<P>, NoRoute> {
     /// Configures as a host candidate.
-    pub fn host(self, addr: SocketAddr) -> CandidateBuilder<P, HasRoute> {
+    pub fn host(self, addr: SocketAddr) -> CandidateBuilder<HasProtocol<P>, HasRoute> {
         self.into_has_route(CandidateKind::Host, addr, addr, addr, None)
     }
 
@@ -782,7 +782,7 @@ impl<P> CandidateBuilder<P, NoRoute> {
         self,
         addr: SocketAddr,
         base: SocketAddr,
-    ) -> CandidateBuilder<P, HasRoute> {
+    ) -> CandidateBuilder<HasProtocol<P>, HasRoute> {
         self.into_has_route(
             CandidateKind::ServerReflexive,
             addr,
@@ -794,7 +794,11 @@ impl<P> CandidateBuilder<P, NoRoute> {
 
     /// Configures as a Relayed (TURN) candidate.
     /// Base is set to `addr`, and `local` is the interface address.
-    pub fn relayed(self, addr: SocketAddr, local: SocketAddr) -> CandidateBuilder<P, HasRoute> {
+    pub fn relayed(
+        self,
+        addr: SocketAddr,
+        local: SocketAddr,
+    ) -> CandidateBuilder<HasProtocol<P>, HasRoute> {
         self.into_has_route(
             CandidateKind::Relayed,
             addr,
@@ -811,7 +815,7 @@ impl<P> CandidateBuilder<P, NoRoute> {
         base: SocketAddr,
         local: SocketAddr,
         raddr: Option<SocketAddr>,
-    ) -> CandidateBuilder<P, HasRoute> {
+    ) -> CandidateBuilder<HasProtocol<P>, HasRoute> {
         CandidateBuilder {
             protocol_state: self.protocol_state,
             route_state: HasRoute {
@@ -867,7 +871,7 @@ impl<P> CandidateBuilder<HasProtocol<P>, HasRoute> {
     }
 }
 
-impl CandidateBuilder<HasProtocol<Tcp>, HasRoute> {
+impl<R> CandidateBuilder<HasProtocol<Tcp>, R> {
     /// Configures the TCP type (active, passive, so).
     pub fn tcptype(mut self, t: TcpType) -> Self {
         self.tcptype = Some(t);


src/sdp/parser.rs
```diff
@@ -294,12 +302,12 @@ where
                 port,
                 _,
                 kind,
-                _,     // (" tcptype ", tcptype)
-                raddr, // (" raddr ", addr, " rport ", port)
-                _,     // (" generation ", generation)
-                _,     // (" network-id ", network_id)
-                ufrag, // (" ufrag ", ufrag)
-                _,     // ("network-cost", network_cost)
+                raddr,   // (" raddr ", addr, " rport ", port)
```
lherman-cs, line 305: reodering per spec, https://datatracker.ietf.org/doc/html/rfc6544#section-4.5.

src/sdp/parser.rs
```diff
@@ -1210,7 +1219,7 @@ mod test {
     }
 
     #[test]
-    fn parse_candidate_ufrag() {
+    fn parse_candidates() {
```
lherman-cs, line 1223: renaming the test as I see the existing tests also assert more than just ufrag, so I'm adding tcp assertions here as well.

algesten:
@lherman-cs thanks!

This looks good!

> For the candidate API, I wonder if it would make sense to turn it into a builder API?

Maybe? Could we lock down with typestate variables what options are correct for what type of candidate maybe?

lherman-cs:
@algesten I got some helped with the boilerplate (thanks AI..). It's nowhere near complete, it's a good start. it's matching with the current Candidate API + tcptype extension. We can keep adding and enforcing more rules to be more compliant to the specs in the future.

Should we deprecate the current Candidate API for reducing maintenance burden? Or, keep both?

I personally like the current Candidate API is very begineer friendly compared to the CandidateBuilder. But, the builder pattern is nice for type safety and useful for advance use cases. The builder API looks more like Hyper layers now, where the number of generics is probably going to grow, idk how people feel about this.

algesten:
> Should we deprecate the current Candidate API for reducing maintenance burden? Or, keep both?

Breaking change is ok. We put it in str0m 0.15

> I personally like the current Candidate API is very begineer friendly compared to the CandidateBuilder. But, the builder pattern is nice for type safety and useful for advance use cases. The builder API looks more like Hyper layers now, where the number of generics is probably going to grow, idk how people feel about this.

I think this looks great.

Some feedback on the API, can we change this

```rs
  // A standard UDP Host candidate
  let udp_host = Candidate::builder()
       .udp()
       .host(addr)
       .build()?;
```

To this:

```rs
  // A standard UDP Host candidate
  let udp_host = Candidate::builder()
       .udp(addr)
       .build()?;
```

My thinking is that a `Candidate::builder().udp().build()` would not be valid and we should ensure this builder helps the user to make correct candidates.

lherman-cs:
I think this pattern should be ok because it is not possible to "build" without having an address. The CandidateBuilder "S" generic state has to be `Ready` first.


This will not compile:
```rust
   // A standard UDP Host candidate
  let udp_host = Candidate::builder()
       .udp()
       .build()?;
```


Also, it feels awkward if the config is for anything other than "host". e.g. 

```rust
Candidate::builder()
       .udp(addr)
       .server_reflexive(base)
       .build()?;
```

Personally, I find its `base` relationship to `addr` gets lost, despite server reflexive candidates' `addr` representing the public mirror to the `base`.

src/ice/candidate.rs
```diff
@@ -679,6 +679,175 @@ impl<'de> Deserialize<'de> for Candidate {
...
+        self.into_ready(CandidateKind::Host, addr, Some(addr), Some(addr), None)
+    }
+
+    /// Configures as a Server Reflexive (STUN) candidate.
+    /// Returns error if `addr` and `base` are different IP versions.
+    pub fn server_reflexive(
+        self,
+        addr: SocketAddr,
+        base: SocketAddr,
+    ) -> Result<CandidateBuilder<P, Ready>, IceError> {
```
thomaseizinger: One thing I quite like about e.g. the builders from the `http` crate is that they only fail at the very end and carry the errors from previous steps through.

algesten:
What about acknowledging that for 99% of cases, UDP is the implicit default?

Just random ideas.

```rs
let udp_candidate = Candidate::host(addr).build();

let tcp_candidate = Candidate::host(addr).tcp().build();
```

algesten:
Or reverse

```rs

let udp_candidate = Candidate::host(addr).build();

let tcp_candidate = Candidate::tcp().host(addr).build();
```

src/ice/candidate.rs
lherman-cs: I like that. I'll update the PR with this in mind.

lherman-cs:
> What about acknowledging that for 99% of cases, UDP is the implicit default?
> 
> Just random ideas.
> 
> ```rust
> let udp_candidate = Candidate::host(addr).build();
> 
> let tcp_candidate = Candidate::host(addr).tcp().build();
> ```

I like the idea. But, not sure if there's a good way to achieve this. We're using `P` as a simple state machine. There are only 3 states here: NoProtocol, Udp, Tcp. If we wanted the default to be Udp, we have to merge the `NoProtocol` and `Udp` transitions. 

Before:

`NoProtocol` -> `Udp`
        |-----------> `Tcp`
        
        
After:

There's no `NoProtocol` state anymore.

`Udp` --> `Tcp`.

Now, Udp state can transition to Tcp state. 

It's technically fine now since we don't have any extensions for Udp only. Otherwise, this will create a hole where users can just create UDP-only extensions, then transition to Tcp.


On a separate note, it seems reasonable to keep both Candidate::host, and Candidate::builder APIs after looking at `http` and `tracing-subscriber` crates,. One API provides sane defaults and beginner-friendly features, and the other is customizable.

`http`: [Response::from_parts](https://docs.rs/http/latest/http/response/struct.Response.html#method.from_parts) vs [Response::builder](https://docs.rs/http/latest/http/response/struct.Response.html#method.builder)
`tracing-subscriber`: `tracing_subscriber::fmt::init` vs `tracing_subscriber::fmt()`

src/ice/candidate.rs
lherman-cs: Addressed this in the update.

lherman-cs:
@algesten @thomaseizinger thoughts on the current state?

src/ice/candidate.rs
```diff
@@ -117,6 +122,51 @@ impl fmt::Debug for Candidate {
...
+    ///
+    /// ```
+    /// # use str0m::Candidate;
+    /// # use str0m::error::IceError;
+    /// # use str0m::net::{Protocol, TcpType};
+    /// # use std::net::SocketAddr;
+    ///
+    /// let addr: SocketAddr = "192.168.1.1:12345".parse().unwrap();
+    ///
+    /// // A standard UDP Host candidate
```
thomaseizinger:
```suggestion
    /// // A standard UDP host candidate
```

src/ice/candidate.rs
```diff
@@ -117,6 +122,51 @@ impl fmt::Debug for Candidate {
...
+    /// let udp_host = Candidate::builder()
+    ///     .udp()
+    ///     .host(addr)
+    ///     .build()?;
+    ///
+    /// // A TCP Host candidate with a passive role
```
thomaseizinger:
```suggestion
    /// // A TCP host candidate with a passive role
```

src/ice/candidate.rs
```diff
@@ -117,6 +122,51 @@ impl fmt::Debug for Candidate {
...
+    /// let tcp_host = Candidate::builder()
+    ///     .tcp()
+    ///     .host(addr)
+    ///     .tcptype(TcpType::Passive)
```
thomaseizinger, line 147: Would it be worth reducing the public API surface and changing this to:

```
.passive()
```

`TcpType` could then be a crate-private type.

src/ice/candidate.rs
```diff
@@ -615,18 +684,209 @@ impl<'de> Deserialize<'de> for Candidate {
     }
 }
 
+/// Marker for a builder requiring a protocol selection.
+pub struct NoProtocol;
```
thomaseizinger, line 685: I'd suggest making all of these `#[doc(hidden)]`.

src/ice/candidate.rs
```diff
@@ -615,18 +684,209 @@ impl<'de> Deserialize<'de> for Candidate {
...
+    pub fn ssl_tcp(self) -> CandidateBuilder<Tcp, Init> {
+        self.into_protocol(Some(Protocol::SslTcp))
+    }
+
+    /// Sets the protocol to TLS.
+    pub fn tls(self) -> CandidateBuilder<Tcp, Init> {
+        self.into_protocol(Some(Protocol::Tls))
+    }
+
+    fn into_protocol<NewP>(self, p: Option<Protocol>) -> CandidateBuilder<NewP, Init> {
```
thomaseizinger: Why the `Option`? Is there ever a case of `None`?

thomaseizinger:
> @algesten @thomaseizinger thoughts on the current state?

Sorry, in holiday mode right now with limited availability. Will look at this in the new year :)

lherman-cs:
> > @algesten @thomaseizinger thoughts on the current state?
> 
> Sorry, in holiday mode right now with limited availability. Will look at this in the new year :)

No worries :)

lherman-cs:
^ just a rebase to the current main.

src/ice/candidate.rs
```diff
@@ -456,6 +515,11 @@ impl Candidate {
         self.raddr
     }
 
+    /// Returns the candidate's TCP role. Refer to [`TcpType`] for more details.
+    pub fn tcptype(&self) -> Option<TcpType> {
```
thomaseizinger, line 515: Do we need to expose it publicly?

src/ice/candidate.rs
```diff
@@ -615,18 +684,209 @@ impl<'de> Deserialize<'de> for Candidate {
...
+                    addr.ip(),
+                    base.ip(),
+                )));
+            }
+        }
+
+        Ok(Candidate::new(
+            self.foundation,
+            self.component_id,
+            self.proto.expect("proto missing"),
```
thomaseizinger: You could carry the data in the type state structs to avoid this error.

src/io/mod.rs
```diff
@@ -46,6 +47,68 @@ pub enum Protocol {
...
+impl fmt::Display for TcpType {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let str = match self {
+            Self::Active => "active",
+            Self::Passive => "passive",
+            Self::So => "so",
+        };
+        f.write_str(str)
+    }
+}
```
thomaseizinger, line 86: Perhaps add some tests to round-trip this parsing?

thomaseizinger (commented):
A few comments / ideas, sorry for the delay!

src/ice/candidate.rs
lherman-cs: It's future-proofing this config and closer to the ice candidate spec. I think it's confusing if a new extension to this candidate type is introduced, and it also uses similar values, especially since "passive" is a generic term.

tcptype=passive,`<ext2>`=passive, etc. In this case, it is not possible to define tcptype as passive indpendently. idk, I just like to avoid making assumptions about some syntax sugar for this stuff.

src/ice/candidate.rs
lherman-cs: This is needed in some situations. If the offerer uses "passive" mode, then we need to respond with "active". We need to know the tcptype so we can know whether we need to send the initial TCP handshake.

src/io/mod.rs
lherman-cs: This is already done here, https://github.com/algesten/str0m/pull/797/files#diff-f05b2f17c83a20594a7d20c04b22478b13b0e940424d278fc171e7632998b019R888-R909. Or, maybe I'm misunderstanding. Could you clarify?

src/ice/candidate.rs
lherman-cs: I went ahead and updated the route stuff as well. No expects left :). 

Also renamed Init and Ready to NoRoute and HasRoute to be consistent with the protocol selection.

src/ice/candidate.rs
thomaseizinger: We could also call it `type_passive` or something like that! It may very well not be worth it though.

src/io/mod.rs
thomaseizinger: Some parts are tested there implicitly yeah. I was thinking of a short test that ensures we can round-trip all variants.

thomaseizinger (approved):
LGTM

src/ice/candidate.rs
lherman-cs: yea, I don't think it's worth changing it to `type_passive`

src/ice/candidate.rs
```diff
@@ -688,6 +688,13 @@ impl<'de> Deserialize<'de> for Candidate {
 /// Marker for a builder requiring a protocol selection.
 pub struct NoProtocol;
 
+#[doc(hidden)]
+/// Marker indicating a transport protocol has been assigned to the builder.
+pub struct HasProtocol<P> {
+    _marker: PhantomData<P>,
+    protocol: Protocol,
+}
+
```
algesten, line 693: Is this to make things nicer?

lherman-cs: This is an internal detail, so we can discriminate `Option<Protocol>` -> `Protocol` early for next configs in selecting route and final build

algesten:
@lherman-cs Do you consider this one ready to merge?

lherman-cs:
@algesten yep, this is good to go, thanks!

PR #804, opened 2025-12-25
Reduce allocating in str0m-aws-lc-rs

alexlapa:
Just a minor optimization for `AeadAes128GcmCipher` and `AeadAes256GcmCipher` encrypt/decrypt for `aws-lc` crypto backend. ~20% gain according to my benches.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
   * Bump dimpl 0.2.1 for SChannel compat #801
+  * Reduce allocating in `str0m-aws-lc-rs` encrypt/decrypt #804
 
 0.14.2
 


crypto/aws-lc-rs/src/srtp.rs
```diff
@@ -90,15 +90,15 @@ impl AeadAes128GcmCipher for AwsLcRsAeadAes128GcmCipher {
         let nonce = Nonce::try_assume_unique_for_key(iv)
             .map_err(|e| CryptoError::Other(format!("Invalid nonce: {}", e)))?;
 
-        // Copy input to a mutable buffer
-        let mut buf = input.to_vec();
```
alexlapa, line 94: Just getting rid of this temp vec and performing encrypt/decrypt in place in the provided output buffer.

algesten (approved):
EXCELLENT!

I thought I done this. It's worth checking dimpl as well.

crypto/aws-lc-rs/src/srtp.rs
algesten: Nice!

algesten:
@alexlapa are you on Discord? I tried searching but don't find you. Perfectly fine if you're not. Just checking if I had a way to chat with you.

PR #806, opened 2025-12-27
Change bitrate only if significant changes

Razzwan:
To reduce the number of RTCP requests, we’ll adjust the bitrate only when there’s a significant change (>5%). 

Does this make sense for your use case?

### DIFF
--- a/proto/src/bandwidth.rs
+++ b/proto/src/bandwidth.rs
@@ -82,9 +82,6 @@ impl Bitrate {
         if a == 0.0 {
             return b != 0.0;
         }
-        if b == 0.0 {
-            return a != 0.0;
-        }
         ((a - b).abs() / a) > 0.05
     }
 }
--- a/src/bwe.rs
+++ b/src/bwe.rs
@@ -40,7 +40,7 @@ impl<'a> Bwe<'a> {
     /// # use str0m::{Rtc, bwe::Bitrate};
     /// let mut rtc = Rtc::new();
     ///
-    /// rtc.bwe().set_current_bitrate(Bitrate::kbps(250));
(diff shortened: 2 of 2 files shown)


k0nserv:
This makes more sense in the application layer than within str0m itself I think. It's too specific to each application how exactly this should be handled I think

algesten:
@Razzwan  I'm working daily on bandwidth estimation in [this branch](https://github.com/algesten/str0m/compare/fix/bwe-probe-cluster).

One piece of the puzzle I'm toying with is a smoothing of emitted estimation. See [this commit](https://github.com/algesten/str0m/commit/471a91a49aa7639eccba09c7e8ebf446b9a7d58e).

Whether or not that smoothing makes it to the final version, I don't know. Once I got something to show, I will run it all by @k0nserv – and it's not ready for that yet.

Razzwan:
> This makes more sense in the application layer than within str0m itself I think. It's too specific to each application how exactly this should be handled I think

Yes, I think you're right. I'm not sure this solution is universal.

@algesten What should we use instead of `set_current_bitrate` after your changes? Am I correctly understand that adaptive bitrate will be added to the core of the project? How will switching between different Rids in this case?

algesten:
The branch doesn't really change the fundamentals of BWE. Instead of set_current, it measures the bitrate you're sending. 

It adds probe clusters which is a method to rapidly check whether there is more bw available. It's useful when you
try to rapidly increase.

The problem i'm mainly staring at now is an oscillation that occurs when the desired bitrate is above the link max.

algesten:
I think this works fine in BWE V2 branch.

PR #808, opened 2025-12-31
Dev/crypto bench

xnorpx:
This pr adds crypto benchmark it also add smal cache optimization for Apple crypto

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -15,8 +15,18 @@ env:
 
 jobs:
   benchmark:
-    name: Benchmark
-    runs-on: ubuntu-latest
+    name: Benchmark (${{ matrix.name }})
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: ubuntu-latest
+            name: linux
+          - os: macos-latest
+            name: macos
+          - os: windows-latest
+            name: windows
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/bench.yml
```diff
@@ -0,0 +1,49 @@
+name: Benchmarks
+
+on:
+  schedule:
+    # Run every night at 2:00 AM UTC
+    - cron: '0 2 * * *'
+  workflow_dispatch:
+    # Allow manual triggering
+  pull_request:
```
xnorpx, line 9: remove before push

algesten:
@xnorpx isn't it quite dodgy to run benchmarks on CI machines that we have zero control over? I assume github actions run on massively virtualized environments sharing hardware resources with an unpredictable number of other CI jobs. How could such benchmarks even be comparable to each other?

xnorpx:
> @xnorpx isn't it quite dodgy to run benchmarks on CI machines that we have zero control over? I assume github actions run on massively virtualized environments sharing hardware resources with an unpredictable number of other CI jobs. How could such benchmarks even be comparable to each other?

@algesten yes you are correct, the idea to have predictable runs that you can compare between is not happening here. You need private runner under your desk for that.

However I think this is one of these, it's not perfect but it's good enough. You can look at relative results between cryptos in report running on one machine and it can give you a "hint".

You can then quickly check out the code and run the benchmark locally if you want to verify in a more controlled environment.

So the value here is just to have benchmark that compiles and is easy to run. (The CI job is mostly there to make sure it doesn't break)

algesten:
> However I think this is one of these, it's not perfect but it's good enough. You can look at relative results between cryptos in report running on one machine and it can give you a "hint".

I'm not convinced. It's 3k lines of code/maintenance complexity for something that is very unclear what it shows us. If we have these results somewhere viewable, worst case this will make str0m look bad for someone stumbling on it.

So far I managed performance via doing manual flamegraphs in a controlled environment (my mac). What are your goals with this?

xnorpx:
> > However I think this is one of these, it's not perfect but it's good enough. You can look at relative results between cryptos in report running on one machine and it can give you a "hint".
> 
> I'm not convinced. It's 3k lines of code/maintenance complexity for something that is very unclear what it shows us. If we have these results somewhere viewable, worst case this will make str0m look bad for someone stumbling on it.
> 
> So far I managed performance via doing manual flamegraphs in a controlled environment (my mac). What are your goals with this?

So the goal was to have a quick way for people to be able to see individual crypto performance. (We noticed that apple crypto was not great and I got sad that I didn't catch that earlier)

But I do get your point that people without context can missinterpret the results and it is also a maintaince burder without clear day to day benefit.

I close it but keep the branch for now until we tested out the refactored apple crypto one then I delete the branch.

PR #813, opened 2026-01-03
Update Apple Crypto crate to use CryptoKit

efer-ms:
Moved most implementations to use CryptoKit via the apple-cryptokit-rs crate. This avoid a Global Lock that Common Crypto has that causes huge pauses.

The only CommonCrypto use is now for AES-ECB which isn't supported in CryptoKit. We also now use the CM implementation build on this ECB. This avoids the Global Lock issue since AES-ECB is a single function call with no state between calls, so it should not have the overhead.

I also had AI add a bunch of unit tests using standard test vectors to verify implementation compatibility.

### DIFF
--- a/crypto/apple-crypto/src/srtp.rs
+++ b/crypto/apple-crypto/src/srtp.rs
@@ -226,7 +226,15 @@ impl SrtpProvider for AppleCryptoSrtpProvider {
     }
 }
 
-// CTR Helper Functions
+// CTR implementation.
+//
+// This is not a generic CTR implementation, as it imposes a 2k limit on
+// the input/output, which is more than enough for our SRTP use where each
+// packet is smaller than the MTU.
+//
+// Note: If we need to support larger blocks, we could loop 2k at a time.
+// However, CTR is frowned upon, it's only provided since it is a requirement
+// for WebRTC, but in almost all cases AES-GCM should be used.
 fn aes_ctr_round(
     key: &[u8],
     iv: &[u8; 16],


efer-ms:
@algesten , @xnorpx  .. ok this is ready, as they published a new apple-cryptokit-rs with the changes we needed.

crypto/apple-crypto/src/srtp.rs
```diff
@@ -446,16 +218,58 @@ impl SrtpProvider for AppleCryptoSrtpProvider {
...
+// CTR Helper Functions
+fn aes_ctr_round(
+    key: &[u8],
+    iv: &[u8; 16],
+    input: &[u8],
+    output: &mut [u8],
+) -> Result<(), CryptoError> {
+    // First, we'll make a copy of the IV with a countered as many times as
+    // needed into a new countered_iv.
+    let mut iv = iv.to_vec();
```
xnorpx: how large is this, can we either reuse memory or use smallvec/arrayvec?

crypto/apple-crypto/src/srtp.rs
```diff
@@ -355,26 +105,48 @@ impl AeadAes256GcmCipher for AppleCryptoAeadAes256GcmCipher {
...
+        _ => &aads.concat(),
+    };
 
-        Ok(ct_len)
-    }
+    let output_vec = apple_cryptokit::aes_gcm_decrypt_with_aad(key, iv, input, aad)
+        .map_err(|err| CryptoError::Other(format!("{err:?}")))?;
+    output[..output_vec.len()].copy_from_slice(output_vec.as_slice());
+
+    Ok(output_vec.len())
```
xnorpx, line 149: ok here is the funky api you mentioned

efer-ms: Ah.. you mean the fact the crate returns new vec each time? Yeah, I filed an issue and offered suggestions to the maintainer. I'll let him decide which way to proceed, but I can update that crate as needed so we can provide a `&mut [u8]` to the cipher. And have it return the size.

crypto/apple-crypto/src/srtp.rs
efer-ms: fixed.. it can just copy the iv it's a fixed 16 bytes. I should update wincrypto, since this code was copied from there.

xnorpx:
Ok, sounds like @algesten is on board. Let's merge it and we can followup with the allocation optimization if the crate owner agrees to change the api.

PR #817, opened 2026-01-03
Efer/common crypto compliance tests

efer-ms:
This moves hash/hmac/cipher compliance tests to str0m. Since the pipeline runs tests against all our providers, this ensures all the providers run test cases against a set of known vectors, rather than each implmentation needing to repeat the same code.

We could do something similar for dimpl providers.

Altervatively, we could have the test code in a common place, and all providers implement the individual test cases to call the common code. Passing in the provider to test.


algesten (approved):
Neat!

Thanks for doing this!

PR #820, opened 2026-01-05
Surface Sender Info as Event

k0nserv:


src/streams/receive.rs
```diff
@@ -441,7 +459,7 @@ impl StreamRx {
 
             // If we get an SR before the first packet, we update the potential clock rate.
             if let Some(info) = &mut self.sender_info {
-                info.1.rtp_time = MediaTime::new(info.1.rtp_time.numer(), time.frequency());
+                info.info.rtp_time = MediaTime::new(info.info.rtp_time.numer(), time.frequency());
```
xnorpx: nit: info.info reads funky, maybe sender_info.info or find another name for the inner info.

algesten: For local variables in functions, I think shorter is better. `i` even to get `i.info`

k0nserv: I like `i.info`. The scope is tiny so that seems fine

PR #823, opened 2026-01-10
BWE V2

algesten:


docs/BWE.md
```diff
@@ -0,0 +1,752 @@
...
+- **Capped at 1.5×** when `LossLimitedBweIncreasing` (loss-limited but
+  recovering): The network was recently congested but is recovering. We
+  allow moderate probing to help the estimate grow, but cap it at 1.5×
+  current estimate to avoid re-triggering congestion. This provides a
+  conservative path back to higher rates.
+
+Probes are also capped by application constraints to avoid wasteful
+probing:
+
+- **2× `max_bitrate`**: Probes are capped at twice the application's
```
xnorpx, line 403: I wonder if the multiplier should be exposed if not already. I am thinking of applications that use server to server communication and might set say 500Mbit/s as max_bitrate. The 2x might just be to much?

docs/BWE.md
```diff
@@ -0,0 +1,752 @@
+# Bandwidth Estimation (BWE)
```
xnorpx, line 1: What BWE stats are we emitting and can we emit? If we want to log these every X second can one get an understanding of how it works today or do we only have debug level logging for this?

src/_internal_test_exports/mod.rs
```diff
@@ -43,3 +43,6 @@ impl PayloadParams {
         self.locked
     }
 }
+
+/// UNSTABLE: not public API!
+pub use crate::bwe_::ProbeClusterConfig;
```
xnorpx, line 48: We might want to expose them so people can do experiments and it can be non stable public api.

src/bwe/alr_detector/budget.rs
```diff
@@ -0,0 +1,269 @@
...
+    max_bytes_in_budget: i64,
+    /// Current bytes remaining in budget (can be negative = debt)
+    bytes_remaining: i64,
+    /// Whether underuse can build up credit across intervals
+    can_build_up_underuse: bool,
+}
+
+impl IntervalBudget {
+    /// Window size for budget tracking
+    const WINDOW: Duration = Duration::from_millis(500);
```
xnorpx, line 33: Config?

src/bwe/alr_detector/detector.rs
```diff
@@ -0,0 +1,336 @@
...
+    bandwidth_usage_ratio: f64,
+    /// Budget ratio threshold to enter ALR (0.80 = 80%)
+    start_budget_level_ratio: f64,
+    /// Budget ratio threshold to exit ALR (0.50 = 50%)
+    stop_budget_level_ratio: f64,
+}
+
+impl AlrDetector {
+    /// Create a new ALR detector with libWebRTC default configuration.
+    pub fn new() -> Self {
```
xnorpx, line 58: nit: new without parameters and no default. But this might just be your preference.

src/bwe/alr_detector/detector.rs
```diff
@@ -0,0 +1,336 @@
...
+        Self {
+            // Start with zero bitrate, will be updated via set_estimated_bitrate()
+            // Use can_build_up_underuse=true for ALR detection
+            budget: IntervalBudget::new(Bitrate::ZERO, true),
+            alr_started_at: None,
+            last_send_time: None,
+
+            // libWebRTC defaults from alr_detector.cc:30-49
+            bandwidth_usage_ratio: 0.65,
```
xnorpx, line 67: be able to configure or should they just be const on top?

src/bwe/delay/arrival_group.rs
```diff
@@ -1,19 +1,19 @@
 use std::mem;
 use std::time::{Duration, Instant};
 
-use crate::rtp_::SeqNo;
+use crate::rtp_::TwccSeq;
 
-use super::time::{TimeDelta, Timestamp};
-use super::AckedPacket;
+use super::super::time::{TimeDelta, Timestamp};
```
xnorpx, line 6: just use crate at this point?

src/bwe/delay/control.rs
```diff
@@ -1,18 +1,20 @@
...
-use super::{AckedPacket, BandwidthUsage};
+use super::trendline::TrendlineEstimator;
+use crate::rtp_::Bitrate;
+use crate::util::{already_happened, MovingAverage};
 
 const MAX_RTT_HISTORY_WINDOW: usize = 32;
 const UPDATE_INTERVAL: Duration = Duration::from_millis(25);
 /// The maximum time we keep updating our estimate without receiving a TWCC report.
 const MAX_TWCC_GAP: Duration = Duration::from_millis(500);
+/// RFC 6298: EWMA smoothing factor for RTT (alpha = 1/8)
```
xnorpx: Exponential Window Moving Average (EWMA) ?

I would write it out once

docs/BWE.md
algesten: The point here is that probe clusters are tiny (15ms of data), and to know if you got enough bandwidth, your cluster must overshoot. In webrtc, they have two values:

1. an aggregate value of the current encoder bitrate
2. a max_bitrate

When they probe, they probe at 2x the aggregate. They use `max` as a hard cap. In str0m, the 2x behavior is correct, we should target 2x desired_bitrate, that is analogous to webrtc, but we might want to also introduce a hard cap.

docs/BWE.md
algesten: It's `Event::EgressBitrateEstimate` that happens for a lot of reasons since there are many many parts of what makes up the full BWE system. The rest is in debug logs to some extent.

src/_internal_test_exports/mod.rs
algesten: Maybe. I mean the probes is one thing out of many mechanics. If we want to expose all of the BWE internal gubbins there is more to emit.

src/bwe/delay/rate_control.rs
```diff
@@ -84,21 +90,120 @@ impl RateControl {
...
             State::Hold => {
                 // Do nothing
             }
         }
     }
 
+    fn update_observed_bitrate(&mut self, observed_bitrate: Bitrate) {
+        if self
+            .averaged_observed_bitrate
+            .lower_range(3.0)
```
xnorpx: nit: magic number

src/bwe/alr_detector/budget.rs
algesten: Webrtc BWE has lots of magic numbers that _CAN_ be modified, but he whole thing is also a bit of a "black box" because I don't really know the motivations for the numbers.

We can of course lift out config for all of this, but it's also like, who for? Randomly changing values is probably not going to have the right effect.

I'm in two minds about this. I mostly went for `const XXX`, but then sometimes the vibe coding added config objects, and I was like "oh well, sure, whatevs".

src/bwe/alr_detector/detector.rs
algesten: Sure, can fix.

src/bwe/alr_detector/detector.rs
algesten: Yeah, const on top would be more consistent. Also see other comment.

src/bwe/delay/rate_control.rs
```diff
@@ -84,21 +90,120 @@ impl RateControl {
...
+    ///
+    /// Returns true if:
+    /// 1. Enough time (1 RTT, clamped to 10-200ms) has passed since last change, OR
+    /// 2. Throughput is critically low (< 50% of current estimate)
+    fn time_to_reduce_further(&self, now: Instant, observed_bitrate: Bitrate) -> bool {
+        let Some(last_change) = self.last_estimate_update else {
+            return true; // No previous change, allow decrease
+        };
+
+        // WebRTC uses: clamp(rtt, 10ms, 200ms)
```
xnorpx, line 139: almost wonder if there should be a file called bwe_constatnts and then a long list and with reference to the libwebrtc implementation. It feels like a lot of these things is easy to miss and will be hard to keep in sync (if we want need to).

But maybe just busy work.

src/bwe/delay/arrival_group.rs
algesten: You might have a point :)

src/bwe/delay/control.rs
algesten: True

src/bwe/delay/rate_control.rs
algesten: ok!

src/bwe/delay/rate_control.rs
xnorpx: ok I saw in another config that you were as split as me :)

algesten: I've made a couple of passes making an LLM check the webrtc vs str0m code and spot differences and similarities. Mostly I think it's covered in BWE.md.

src/bwe/delay/rate_control.rs
```diff
@@ -115,52 +220,41 @@ impl RateControl {
...
 
         if self.estimated_bitrate < new_estimate {
             // Avoid increasing the bitrate on overuse
             new_estimate = self.estimated_bitrate;
         }
 
-        self.averaged_observed_bitrate
-            .update(observed_bitrate.as_f64());
-
         #[allow(unused)]
```
xnorpx, line 257: this allow still needed?

algesten: Need to check

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+        };
+
+        if self.bytes_sent == DataSize::ZERO {
+            // No packets sent yet - can send immediately.
+            return started_at;
+        }
+
+        // Packet-level pacing: schedule based on bytes sent at the target bitrate.
+        // next_time = start_time + (bytes_sent / target_bitrate)
+        let send_duration = self.bytes_sent / self.config.target_bitrate;
```
xnorpx, line 190: Can target_bitrate be 0 here?

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+        }
+
+        // Enforce min burst spacing (`min_probe_delta`) between successive padding requests.
+        // (The very first burst is not gated because `last_padding_at` starts at already_happened()).
+        if now < self.last_padding_at + self.config.min_probe_delta() {
+            return None;
+        }
+
+        let next_time = self.next_probe_time();
+        if now < next_time {
```
xnorpx: nit: just inline this here on 239?

xnorpx: if now <self.next_probe_time()

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+            self.update_state(State::ProbingComplete);
+            // Fall through to process_probing_complete logic.
+            self.process_probing_complete(now);
+            return;
+        }
+
+        // Further exponential probing (when estimate increased enough).
+        if self.estimated_bitrate > self.min_bitrate_to_probe_further {
+            let target = self.estimated_bitrate * self.config.further_exponential_probe_scale;
+            self.initiate_probing(now, vec![target], true, self.alr_start_time.is_some());
```
xnorpx: do we need to create a vec here?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+        let current_bwe_limit =
+            self.estimated_bitrate * self.config.allocation_probe_limit_by_current_scale;
+
+        let mut first_probe_rate = self.max_bitrate * self.config.first_allocation_probe_scale;
+        let mut limited_by_current_bwe = current_bwe_limit < first_probe_rate;
+        if limited_by_current_bwe {
+            first_probe_rate = current_bwe_limit;
+        }
+
+        let mut probes = vec![first_probe_rate];
```
xnorpx: use arrayvec?

src/bwe/probe/estimator.rs
```diff
@@ -0,0 +1,602 @@
...
+            false
+        });
+    }
+}
+
+impl ProbeEstimatorState {
+    pub fn new(config: ProbeClusterConfig) -> Self {
+        Self {
+            config,
+            finalize_at: not_happening(),
```
xnorpx, line 206: if you have default this coult just be ..Default

src/bwe/loss_controller.rs
```diff
@@ -566,10 +693,21 @@ impl LossController {
 
     fn get_candidate_bandwidth_upper_bound(&self) -> Bitrate {
         let mut upper_bound = self.max_bitrate;
+
+        // When in ALR and we have a link capacity estimate from probes,
+        // use it as the upper bound. This prevents estimating beyond proven capacity.
+        if self.is_in_alr() {
```
xnorpx, line 699: if let chains one day :)

src/bwe/loss_controller.rs
```diff
@@ -883,33 +1043,31 @@ impl AsValid<Bitrate> for Option<Bitrate> {
...
             inherent_loss_upper_bound_offset: 0.05,
-            inherent_loss_upper_bound_bandwidth_balance: Bitrate::kbps(75),
+            inherent_loss_upper_bound_bandwidth_balance: Bitrate::kbps(100),
             inherent_loss_lower_bound: 1.0e-3,
             newton_iterations: 1,
             newton_step_size: 0.75,
             not_increase_if_inherent_loss_less_than_average_loss: true,
-            delayed_increase_window: Duration::from_millis(1000),
-            bandwidth_rampup_upper_bound_factor: 1_000_000.0,
+            delayed_increase_window: Duration::from_millis(300),
```
xnorpx, line 1069: Did these numbers change in libwebrtc?

docs/BWE.md
xnorpx: Yes, the hard cap is probably the way, now one just have to find/pick one.

src/bwe/probe/cluster.rs
algesten: Good question

src/bwe/probe/cluster.rs
algesten: Sure

src/bwe/probe/control.rs
algesten: Need to check if we can get rid of that.

src/bwe/probe/control.rs
algesten: Yeah, or reusable vec on instance.

src/bwe/probe/estimator.rs
algesten: Yes

src/bwe/loss_controller.rs
algesten: I assume so. Will double check.

proto/src/bandwidth.rs
```diff
@@ -183,6 +183,11 @@ impl DataSize {
         self.0 as usize
     }
 
+    /// The number of bytes as i64.
+    pub fn as_bytes_i64(&self) -> i64 {
+        self.0 as i64
```
k0nserv: Might want to use `try_into` and `expect` here?

src/bwe/alr_detector/budget.rs
```diff
@@ -0,0 +1,269 @@
...
+///
+/// This implements the same logic as libWebRTC's IntervalBudget class
+/// to ensure ALR detection behaves identically when handling bursty
+/// encoder traffic (keyframes, scene changes, temporal layers, etc.)
+#[derive(Debug, Clone)]
+pub struct IntervalBudget {
+    /// Target bitrate for budget calculation
+    target_rate: Bitrate,
+    /// Maximum bytes that can accumulate in the budget
+    max_bytes_in_budget: i64,
```
k0nserv: `DataSize`?

src/bwe/alr_detector/budget.rs
```diff
@@ -0,0 +1,269 @@
...
+    /// Current bytes remaining in budget (can be negative = debt)
+    bytes_remaining: i64,
```
k0nserv: `DataSize`?

src/bwe/alr_detector/detector.rs
```diff
@@ -0,0 +1,336 @@
...
+/// // Call for each media packet (not padding/probes)
+/// alr.on_bytes_sent(DataSize::bytes(1200), now);
+///
+/// // Check ALR state
+/// if let Some(alr_start) = alr.alr_start_time() {
+///     // We've been in ALR since `alr_start`
+/// }
+/// ```
+#[derive(Debug)]
+pub struct AlrDetector {
```
k0nserv, line 21: Is there a hidden state machine here that would benefit from being be surfaced as an `enum`?

src/bwe/alr_detector/detector.rs
```diff
@@ -0,0 +1,336 @@
...
+    /// Should be called for every media packet sent (excluding padding and probes).
+    /// This updates the IntervalBudget and checks for ALR state transitions.
+    pub fn on_bytes_sent(&mut self, bytes: DataSize, now: Instant) {
+        // First call - just record time
+        let Some(last) = self.last_send_time else {
+            self.last_send_time = Some(now);
+            return;
+        };
+
+        // saturating_duration_since handles potential clock skew between packets
```
k0nserv: Clock skew on `Instant`?

src/bwe/delay/control.rs
```diff
@@ -26,10 +28,12 @@ pub struct DelayController {
     /// Last estimate produced, unlike [`next_estimate`] this will always have a value after the
     /// first estimate.
     last_estimate: Option<Bitrate>,
-    /// History of the max RTT derived for each TWCC report.
+    /// Smoothed RTT using EWMA (RFC 6298, alpha = 1/8).
+    /// Replaces the previous mean_max_rtt calculation for more accurate RTT estimation
```
k0nserv: This comment should go I think. It's LLM noise

src/bwe/delay/control.rs
```diff
@@ -66,37 +71,49 @@ impl DelayController {
...
+                // Got a new delay variation, add it to the trendline.
+                //
+                // IMPORTANT: Match WebRTC's TrendlineEstimator time base.
+                // WebRTC calls Detect/UpdateThreshold with `arrival_time_ms` (remote receive time),
+                // not the local "time we processed this feedback". Using the remote receive time
+                // avoids threshold adaptation artifacts when many deltas are processed in one
+                // feedback batch (e.g. TWCC reports).
                 self.trendline_estimator
-                    .add_delay_observation(delay_variation, now);
+                    .add_delay_observation(delay_variation, delay_variation.last_remote_recv_time);
```
k0nserv, line 86: So we are trusting the remote's clock here?

src/bwe/delay/control.rs
```diff
@@ -109,43 +126,87 @@ impl DelayController {
...
+    ///
+    /// This is useful for gating behaviors (like probing) that would otherwise
+    /// re-excite the system while we're already congested.
+    pub fn is_overusing(&self) -> bool {
+        self.trendline_estimator.hypothesis() == BandwidthUsage::Overuse
+    }
+
+    /// Update smoothed RTT using EWMA (RFC 6298, alpha = 1/8).
+    ///
+    /// This matches WebRTC's approach of each controller maintaining its own smoothed RTT
```
k0nserv: Drop this comment?

src/bwe/delay/rate_control.rs
k0nserv: This is because of the macro. If we don't have stats collection enabled

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
+//! Probe cluster data structures for bandwidth estimation.
+//!
+//! Probe clusters are short bursts of packets sent at specific bitrates to test
+//! network capacity. They replace continuous padding with targeted bandwidth probing.
```
k0nserv: Remove the bit about replacing continuous padding?

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+    /// Calculates the next probe time based on total bytes sent and target bitrate.
+    /// This naturally handles variable packet sizes (media packets can be much larger
+    /// than padding packets). Next packet can be sent when:
+    /// `now >= start_time + (bytes_sent / target_bitrate)`
+    pub fn next_probe_time(&self) -> Instant {
+        let Some(started_at) = self.started_at else {
+            return already_happened();
+        };
+
+        if self.bytes_sent == DataSize::ZERO {
```
k0nserv: This is redundant. If `byte_sent` is `DataSize::ZERO` then `send_duration` will be computed as `Duration::ZERO` below.

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+    /// to queue (which may result in multiple RTP padding packets), so we request a burst
+    /// sized approximately as:
+    ///
+    /// `burst_bytes = target_bitrate * min_probe_delta`
+    ///
+    /// This avoids unintentionally capping probe throughput at `MAX_PADDING_PACKET_SIZE /
+    /// min_probe_delta` (e.g. 240 bytes / 2ms = 960 kbit/s), which would be far below the
+    /// intended probe bitrate.
+    ///
+    /// Returns `DataSize::ZERO` if it's not time to send yet, or if we've already created
```
k0nserv: This comment is wrong? We return `None` not `DataSize::ZERO`

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+
+        // Mark that we've created padding for this instant
+        self.last_padding_at = now;
+
+        // It's time to send a new probe burst.
+        let min_delta = self.config.min_probe_delta();
+        let mut burst_bytes = self.config.target_bitrate * min_delta;
+
+        // Ensure we request at least one padding packet worth of bytes even if min_delta is 0.
+        burst_bytes = DataSize::bytes(cmp::max(
```
k0nserv: Is it desirable to use padding packets here? They are quite expensive due to the overhead and their small size. We don't want to try and use RTX packets in the first instance.

src/bwe/probe/cluster.rs
```diff
@@ -0,0 +1,412 @@
...
+            burst_bytes.as_bytes_usize() as u64,
+            MAX_PADDING_PACKET_SIZE.as_bytes_usize() as u64,
+        ));
+
+        // If we've already met the byte target but not the packet count, continue to request at
+        // least one packet per burst.
+        let bytes_needed = self.config.target_bytes().saturating_sub(self.bytes_sent);
+        let request_bytes = if bytes_needed == DataSize::ZERO {
+            MAX_PADDING_PACKET_SIZE
```
k0nserv: In this case shouldn't we use as small packets as possible since we already reached the byte target?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BandwidthLimitedCause {
+    LossLimitedBweIncreasing,
+    LossLimitedBwe,
+    DelayBasedLimited,
+    DelayBasedLimitedDelayIncreased,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum State {
```
k0nserv: Diagram?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+    // Probe cluster config defaults
+    min_probe_packets_sent: usize, // 5
+    min_probe_duration: Duration,  // 15ms
+    min_probe_delta: Duration,     // 2ms
+
+    // Gating / limits
+    loss_limited_probe_scale: f64, // 1.5
+}
+
+impl Default for Config {
```
k0nserv: Move to end?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+
+    // Level-triggered allocation probing tracking.
+    last_allocation_probe_time: Instant,
+
+    // Next time we should call `handle_timeout` for `process()`-driven decisions.
+    next_timeout: Instant,
+}
+
+impl ProbeControl {
+    pub fn new() -> Self {
```
k0nserv: `Default`?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+        // - Every 5 seconds while in ALR
+        // - Only when estimate < max_bitrate (headroom available)
+        if in_alr
+            && self.estimated_bitrate < self.max_bitrate
+            && self.state == State::ProbingComplete
+            && time_since_last_allocation_probe >= MIN_TIME_BETWEEN_ALR_PROBES
+        {
+            self.last_allocation_probe_time = now;
+            self.initiate_allocation_probing(now);
+            return; // Early return
```
k0nserv: Drop this comment?

src/bwe/probe/control.rs
```diff
@@ -0,0 +1,851 @@
...
+    fn initiate_probing(
+        &mut self,
+        now: Instant,
+        bitrates_to_probe: Vec<Bitrate>,
+        mut probe_further: bool,
+        is_alr_probe: bool,
+    ) {
+        // Use max_bitrate if set, otherwise allow unlimited probing (initial probing case).
+        let mut max_probe_bitrate = if self.max_bitrate > Bitrate::ZERO {
+            self.max_bitrate * 2.0
```
k0nserv: Magic number

src/bwe/probe/estimator.rs
```diff
@@ -0,0 +1,602 @@
...
+    pub fn handle_timeout(&mut self, now: Instant) {
+        self.states.retain(|s| {
+            let do_keep = now < s.finalize_at;
+            if do_keep {
+                return true;
+            }
+
+            let result = s.do_calculate_bitrate();
+            if let ProbeResult::Estimate(_) = result {
+                // This is already logged at update() time.
```
k0nserv: I don't see this log.

src/bwe/probe/estimator.rs
```diff
@@ -0,0 +1,602 @@
+use std::collections::VecDeque;
```
k0nserv, line 1: Might be interesting to have macros for logging estimates from here

src/bwe/probe/estimator.rs
```diff
@@ -0,0 +1,602 @@
...
+        // Only packets with a known remote receive time participate in probe estimation.
+        let Some(recv_time) = record.remote_recv_time() else {
+            return false; // lost/unreceived packet -> ignore for probe result
+        };
+
+        let packet_size = DataSize::from(record.size());
+        let send_time = record.local_send_time();
+
+        // Track min/max send time among included packets.
+        match self.first_send_time {
```
k0nserv: For these match statements we can use `get_or_insert`

```rust
let first_send_time = self.first_send_time.get_or_insert(send_time);
*first_send_time = first_send_time.min(send_time);
```

src/bwe/probe/estimator.rs
```diff
@@ -0,0 +1,602 @@
...
+        debug!(%result, "Probe result");
+
+        Some((self.config, bitrate))
+    }
+
+    /// Calculate the estimated bitrate for this probe cluster.
+    fn do_calculate_bitrate(&self) -> ProbeResult {
+        // WebRTC requires at least kMinClusterSize (4) packets
+        if self.packet_count < MIN_CLUSTER_SIZE {
+            return ProbeResult::ClusterTooSmall {
```
k0nserv, line 272: Why would even send a cluster that's too small or is it rather that we might have lost packets or their feedback?

src/bwe/acked_bitrate_estimator.rs
```diff
@@ -3,9 +3,6 @@ use std::time::{Duration, Instant};
 use crate::rtp_::DataSize;
 use crate::Bitrate;
 
-// Ported from libWebRTC's src/modules/congestion_controller/goog_cc/bitrate_estimator.cc at
```
k0nserv, line 6: This is still true, no?

src/bwe/link_capacity_estimator.rs
```diff
@@ -0,0 +1,196 @@
...
+/// can change, and old measurements become less reliable.
+pub struct LinkCapacityEstimator {
+    /// Current estimate of link capacity, if available
+    capacity_estimate: Option<Bitrate>,
+
+    /// Time when the capacity estimate was last updated
+    last_estimate_time: Option<Instant>,
+
+    /// Duration after which capacity estimate expires and resets
+    capacity_estimate_reset_window: Duration,
```
k0nserv: This is only ever set to `DEFAULT_RESET_WINDOW` does it need to be a field?

src/bwe/link_capacity_estimator.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+impl LinkCapacityEstimator {
+    /// Default duration before capacity estimate resets (60 seconds)
+    const DEFAULT_RESET_WINDOW: Duration = Duration::from_secs(60);
+
+    /// Create a new LinkCapacityEstimator with default settings
+    pub fn new() -> Self {
```
k0nserv, line 32: `Default`?

src/bwe/link_capacity_estimator.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub fn update_from_probe(&mut self, probe_estimate: Bitrate, now: Instant) {
+        // Only accept valid probe estimates
+        if !probe_estimate.is_valid() {
+            return;
+        }
+
+        // Update or set the capacity estimate
+        // If we already have an estimate, take the max (capacity shouldn't decrease
+        // from successful probes, only from decay/timeout)
+        self.capacity_estimate = Some(
```
k0nserv: Consider `Option::get_or_insert`

src/pacer/leaky.rs
```diff
@@ -317,16 +98,6 @@ impl Pacer for LeakyBucketPacer {
         let (next_poll_time, queue) = self.next_poll(now)?;
 
         if now < next_poll_time {
-            let total_packet_count: u32 = self
```
k0nserv, line 320: This has been useful, but maybe it should be gated on trace level being enabled to not redundantly compute

k0nserv, line 320: This has been useful, but maybe it should be gated on trace level being enabled to not redundantly compute

src/pacer/leaky.rs
```diff
@@ -437,21 +270,27 @@ impl LeakyBucketPacer {
...
+                let poll_at = if let Some(ref probe) = self.active_probe {
+                    // During probe: use absolute time directly from probe state
+                    // next_time = probe_start + (bytes_sent / target_bitrate)
+                    // This ensures monotonic time advancement even when now == next_probe_time
+                    probe.next_probe_time()
                 } else {
-                    Duration::ZERO
-                };
+                    // Normal pacing: use relative offset based on debt
+                    let drain_debt_time = self.media_debt / self.adjusted_bitrate;
```
k0nserv, line 298: Maybe we should add in the max burst thing here too

src/pacer/leaky.rs
```diff
@@ -510,14 +362,6 @@ impl LeakyBucketPacer {
         if min_rate > self.adjusted_bitrate {
             // Min rate exceeds our pacing rate, increase the rate to force drain the queue.
             self.adjusted_bitrate = min_rate.clamp(Bitrate::ZERO, MAX_BITRATE);
-            trace!(
```
k0nserv, line 513: This is a useful log, why remove it?

k0nserv, line 513: This has been useful, but maybe it should be gated on trace level being enabled to not redundantly compute

k0nserv, line 513: This has been useful, but maybe it should be gated on trace level being enabled to not redundantly compute

src/bwe/probe/cluster.rs
algesten: This is using "Padding" in the same way as Pacer (and webrtc) does (Pacer returns `RequestPadding`). That doesn't literally mean padding, but can be spurious resends. Whether it's one or the other is up to the `StreamTx` depending on whether it got any RTX or not.

src/bwe/probe/cluster.rs
algesten: True

src/bwe/probe/estimator.rs
algesten: The types don't really lend themselves to it.

src/bwe/loss_controller.rs
algesten: @xnorpx yes. This brings the values into line with current webrtc HEAD.

## Loss-Based BWE Config Changes: libwebrtc

| Parameter | Old Value | New Value | Commit | Date | Author |
|-----------|-----------|-----------|--------|------|--------|
| `bandwidth_rampup_upper_bound_factor` | 1000000.0 | **1.5** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `bandwidth_rampup_upper_bound_factor_in_hold` | 1000000.0 | **1.2** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `loss_threshold_of_high_bandwidth_preference` | 0.15 | **0.2** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `inherent_loss_upper_bound_bandwidth_balance` | 75 kbps | **100 kbps** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `instant_upper_bound_bandwidth_balance` | 75 kbps | **100 kbps** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `observation_window_size` | 20 | **15** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `use_in_start_phase` | false | **true** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `lower_bound_by_acked_rate_factor` | 0.0 | **1.0** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `hold_duration_factor` | 0.0 | **2.0** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `use_byte_loss_rate` | false | **true** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `bound_best_candidate` | false | **true** | `068e6ba696` | 2025-05-20 | Diep Bui |
| `not_use_acked_rate_in_alr` | false | **true** | `9f9b8e0b88` | 2023-10-16 | Diep Bui |

src/bwe/delay/control.rs
algesten:
```
result.receive_time = current_offset_ + delta_since_base;
```

Yeah. This is the same logic as webrtc. `transport_feedback_adapter.cc:232-234`

src/bwe/acked_bitrate_estimator.rs
algesten: All of these references moved into BWE.md since it wasn't consistent across all the ports.

src/pacer/leaky.rs
algesten: Raised in https://github.com/algesten/str0m/issues/827

PR #828, opened 2026-01-12
New ICE Candidate builder API

algesten:
Close #797

PR #829, opened 2026-01-12
New ICE Candidate builder API

lherman-cs:
superseded https://github.com/algesten/str0m/pull/828

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Handle tcptype ICE candidates #797
+  * New API for building ICE Candidates (breaking) #797
   * AV1 Packetizer/Depacketizer #819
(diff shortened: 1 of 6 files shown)


algesten:
@lherman-cs I've forced push away that commit from main.

algesten:
If you rebase, you should get the entire credit.

lherman-cs:
@algesten just rebased.

algesten:
@lherman-cs I tried to explain the requirements here: https://github.com/algesten/str0m/wiki/PR-requirements

Does that make sense?

Again I'm so sorry.

lherman-cs:
@algesten I think these must be enforceable with some automation, otherwise it's easy to forget. I think overall the requirements make sense, but this doesn't include minor tweaks on the PR, though.

PR #831, opened 2026-01-13
 Cap amount SCTP subsystem accepts to buffer

algesten:
Close #825 (picked test case from that PR)

### DIFF
--- a/src/sctp/mod.rs
+++ b/src/sctp/mod.rs
@@ -360,12 +360,24 @@ impl RtcSctp {
         rec.state == StreamEntryState::Open
     }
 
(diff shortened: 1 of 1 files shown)


src/channel.rs
```diff
@@ -51,6 +51,10 @@ impl<'a> Channel<'a> {
     }
 
     /// Write data to the remote peer and indicate whether it's text or binary.
+    ///
+    /// Returns the number of bytes written. This can be 0 if the system cannot handle any
+    /// more data at this point (due to buffers and send windows).
+    #[must_use = "The amount that was handled by the write()"]
```
lolgesten: We should accept all-or-nothing.

PR #832, opened 2026-01-15
assume ice-lite alive until it has a chance to fail

quatrix:
Hi,

When enabling ice-lite, I've observed that candidates get disconnected immediately before they had a chance to send a keep alive packet.

This proposed change assumes the connection is alive if it's ice-lite and we haven't seen any packets yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix incorrectly purged ice-lite pairs #832
   * Provide start Instant in constructor (breaking) #843
   * Fix bug in dimpl resend logic (dimple 0.2.4) #842
   * Fix bug in dimpl HelloVerifyRequests (dimpl 0.2.3) #841
--- a/tests/ice-lite-connectivity.rs
+++ b/tests/ice-lite-connectivity.rs
@@ -1,5 +1,5 @@
 use std::net::Ipv4Addr;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use netem::NetemConfig;
 use str0m::{Event, IceConnectionState, RtcConfig, RtcError};
@@ -15,7 +15,7 @@ pub fn ice_lite_premature_disconnect() -> Result<(), RtcError> {
 
     let mut l = TestRtc::new(Peer::Left);
 
-    let rtc = RtcConfig::new().set_ice_lite(true).build();
+    let rtc = RtcConfig::new().set_ice_lite(true).build(Instant::now());
     let mut r = TestRtc::new_with_rtc(info_span!("R"), rtc);
 
     r.set_netem(NetemConfig::new().latency(Duration::from_millis(5)));


algesten:
Hi @quatrix and welcome to str0m!

Is it possible to simulate this problem with integration tests? I'm asking because I use ice-lite with str0m at work for production systems, and I don't think we've seen this issue.

alexlapa:
I've definitely had connectivity issues that were fixed by disabling ICE lite. Not sure what exactly went wrong though.

quatrix:
Hi @algesten 

> Is it possible to simulate this problem with integration tests? I'm asking because I use ice-lite with str0m at work for production systems, and I don't think we've seen this issue.

Sure! added an integration test, the disconnect happens when adding a very small (>2ms) network latency to the client. (It also seems to be ice-lite specific, with the added latency but without enabling ice-lite, there's no disconnection).

zRedShift:
Hi @algesten,

The bug occurs through this chain:
1. [`last_now`](../blob/5b8f348/src/lib.rs#L1147) is initialized to `already_happened()` (~1 hour in the past)
2. [`accept_offer()`](../blob/5b8f348/src/change/sdp.rs#L99) calls `init_dtls()`, which calls [`self.dtls.handle_timeout(self.last_now)`](../blob/5b8f348/src/lib.rs#L1351)
3. dimpl arms its flight timeout to `now + 1s` - but since `now` is ~1 hour ago, the timeout is still ~59 minutes in the past
4. First `poll_output()` returns this past timeout
5. The timeout future (in our case) returns immediately and we run `handle_timeout` immediately after session creation - before any STUN binding requests can arrive
6. [ICE-lite pruning](../blob/5b8f348/src/ice/agent.rs#L1117-L1119) calls `has_recent_remote_binding_request` which sees [`remote_binding_request_time = None`](../blob/5b8f348/src/ice/pair.rs#L201-L202) and returns `false`, treating "hasn't started yet" as "failed", killing all candidate pairs

You might not hit this if your timeout handling has a minimum granularity or floor that effectively gives STUN a few hundred milliseconds to arrive before the first `handle_timeout` runs.

algesten:
Excellent analysis. Let's fix this! I posted yesterday on our discord this idea: https://gist.github.com/algesten/417fdae74e0b6b05680e0ba8e880205f

It a design that would force the user to provide a timeout before doing any mutation. I.e. it wouldn't be possible to `accept_offer` before giving a timeout.

alexlapa:
> I've definitely had connectivity issues that were fixed by disabling ICE lite. Not sure what exactly went wrong though.

Yeah so I've rechecked what issues I had with ice-lite and its not related to this PR. It was the fact that firefox does not really like ice-lite cause it can only do [Aggressive Nomination](https://datatracker.ietf.org/doc/html/rfc5245#section-8.1.1.2) which [might break when the remote is ice lite](https://bugzilla.mozilla.org/show_bug.cgi?id=1034964)

PR #834, opened 2026-01-17
Fix nightly test for Intel macOS

xnorpx:
Add DYLD_LIBRARY_PATH=/usr/lib/swift for macOS Intel (macos-15-intel) runner.

The apple-crypto feature uses Swift via apple-cryptokit-rs, which requires the Swift runtime libraries. On Intel macOS runners, the dynamic linker cannot find libswiftCore.dylib at runtime without this path being set. This is only needed for Intel x86_64, not ARM64 macOS.


algesten (approved):

xnorpx:
Nightly is passing now: https://github.com/algesten/str0m/actions/runs/21096590001

PR #841, opened 2026-01-25
Stable RNG for tests with dimpl (dimpl 0.2.3)

algesten:


crypto/aws-lc-rs/src/dtls.rs
```diff
@@ -31,7 +31,13 @@ impl DtlsProvider for AwsLcRsDtlsProvider {
         };
 
         // Create a default dimpl Config with AWS-LC-RS crypto provider
-        let config = dimpl::Config::builder()
+        let mut builder = dimpl::Config::builder();
+        if self.is_test() {
+            // We need the DTLS impl to be deterministic for the BWE tests.
+            builder = builder.dangerously_set_rng_seed(42);
```
xnorpx, line 37: Would it be useful to run test that generates a random seed but logs it and run those in nightly to not always test with the same seed?

algesten: As long as we don't run the BWE tests.

PR #845, opened 2026-01-27
Constant stream timeout check

lherman-cs:
I noticed a 15%-25% CPU cost (highly dependent on the load) when the pacer's rate is reduced, even with many packets already in the queue. Since untimed packets are contiguous, we can check the last packet.

image

### DIFF
--- a/src/streams/send_queue.rs
+++ b/src/streams/send_queue.rs
@@ -43,7 +43,12 @@ impl SendQueue {
     }
 
     pub fn need_timeout(&self) -> bool {
-        self.queue.iter().any(|p| p.timestamp == not_happening())
+        // Packets are timestamped contiguously from the front,
+        // so checking the tail summarizes the entire queue's state.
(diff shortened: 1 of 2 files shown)


src/streams/send_queue.rs
```diff
@@ -274,6 +279,55 @@ mod test {
         assert!(queue.pop(Instant::now()).is_some());
     }
 
+    #[test]
+    fn untimed_packets_are_contiguous_at_tail() {
```
lherman-cs, line 283: just in case the handle timeout doesn't maintain the same invariant, this test will fail

algesten:
Nice optimization!

lherman-cs:
> * Add a CHANGELOG.md entry.
> 
>     * Reword the commit to use an uppercase C first.
> 
> 
> Then we merge!

done

algesten:
Thanks!

PR #848, opened 2026-01-28
Fix ICE disconnections with Chrome v144+ when not using ICE Lite

### LINKED ISSUE #846: Disconnections when not using ICE Lite with Chrome v144
We started noticing disconnects when multiple users are connected using the latest version (144) of Chrome (on both Linux and Windows) as well as Edge v144 on Windows. I can not reproduce this with Chrome/Edge v143 and earlier. The latest version of Vivaldi is also unaffected, but it has a current base of Chromium 142.

We noticed this on an implementation using str0m 0.13.0, but it also occurs with str0m 0.16.0, including the chat example.

Reproduction with the chat example:

1. Checkout v0.16 and start the chat example.
2. Use a modern chrome window (I'm on v144.0.7559.96 on Debian Sid, but have reproduced it on Windows with both Chrome and Edge v144)
3. Start opening tabs of the chat page, connecting and enabling camera on each tab
4. With in a minute or so, with 3+ connections, connections will start dropping. It seems to repro faster with more tabs - with 5 or more it tends to reproduce in under 1m.
5. Repeat the test with set_ice_lite(true) on the Rtc builder, and notice that the issue doesn't occur. https://github.com/algesten/str0m/blob/07c2beaebdadb229aa8c1b8a94e1e13e6608a570/examples/chat.rs#L88

This happens both with a local test, and with our service hosted on public IPs. I don't have a great setup to reproduce this, but user reports indicate that it also happens when the connections are coming from different remote users and not just from the same machine. For now, we've just enabled ICE Lite on our systems to work around this.

I'm going to continue testing to see if I can find any relevant logging or other data. Please let me know if there is any specific test I can perform.

algesten:
## Summary

- Fix for ICE disconnections occurring with Chrome v144+ when NOT using ICE Lite mode
- Modifies `is_still_possible()` to consider incoming binding requests as evidence of connectivity
- Adds integration test to verify stable connections

## Problem

Chrome v144+ appears to have changed its STUN response behavior, causing str0m to prematurely prune candidate pairs when not using ICE Lite mode. The symptom was connections dropping within ~10 seconds when 3+ connections were active.

The root cause was that `is_still_possible()` only considered str0m's outgoing binding requests, but didn't factor in incoming binding requests from Chrome. When Chrome stopped responding to str0m's binding requests (but still sent its own), the pair was incorrectly pruned.

## Fix

The fix modifies `is_still_possible()` to also check `remote_binding_request_time`. If we've received a binding request from the remote peer within the last 15 seconds (using the same `RECENT_BINDING_REQUEST` constant as ICE Lite mode), the pair is considered still viable.

This addresses the asymmetric behavior where the remote peer may still be actively communicating but not responding to our specific binding requests.

## Testing

- Reproduced the issue using Playwright with Chromium v145 (closest available to Chrome v144 on arm64)
- Without fix: Connections dropped within ~10 seconds with 5 concurrent connections
- With fix: All connections remained stable for the full 2-minute test
- Also verified that enabling ICE Lite (the existing workaround) continues to work

Fixes #846

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix ICE disconnections with Chrome v144+ when not using ICE Lite #846
   * Downgrade noisy BWE logs #847
   * Fix H265 PACI depacketizer panic on short payload #849
 
--- a/src/ice/pair.rs
+++ b/src/ice/pair.rs
@@ -375,6 +375,17 @@ impl CandidatePair {
     ///
     /// Returns `false` if the candidate has failed.
     pub fn is_still_possible(&self, now: Instant, timing_config: &StunTiming) -> bool {
+        // If we've recently received a binding request from the remote peer, the pair
+        // is still viable even if our own binding requests aren't getting responses.
+        // This handles the case where the remote peer is still sending us requests
+        // (indicating connectivity) but may not be responding to our requests due to
+        // asymmetric behavior in some browser versions (e.g., Chrome v144+).
+        if let Some(t) = self.remote_binding_request_time {
+            if now - t < RECENT_BINDING_REQUEST {
+                return true;
+            }
+        }
+
         let attempts = self.binding_attempts.len();
         let unanswered = self.unanswered().map(|b| b.0).unwrap_or(0);
 
--- a/tests/ice-remote-binding-keepalive.rs
+++ b/tests/ice-remote-binding-keepalive.rs
@@ -1,0 +1,82 @@
+use std::net::Ipv4Addr;
+use std::time::Duration;
+
+use netem::NetemConfig;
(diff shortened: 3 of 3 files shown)


bekriebel:
Confirmed that this fixes #846 

Thanks!

thomaseizinger:
Interesting, I think this may also be a good fix for Firezone. The underlying source for this can be packet loss. If one side floods the UDP socket with millions of packets (not uncommon in high-throughput scenarios with an aggressive congestion controller), the chance of packet loss repeately affecting the binding requests is quite high simply because they are such a small part of the traffic.

algesten:
True that. Seems like such an obvious fix. In hindsight.

PR #850, opened 2026-01-31
[H265] Add MTU validation and error handling.

sr1990:
- Add MIN_MTU and MIN_FU_PAYLOAD constants with detailed documentation
  * MIN_FU_PAYLOAD = 1 byte (ensures FU packets carry meaningful payload)
  * MIN_MTU = 4 bytes (NALU header 2 + FU header 1 + min payload 1)
- Add MtuErrorKind enum with Zero, TooLarge, TooSmall variants
- Validate MTU bounds in packetize():
  * Reject MTU == 0 (invalid configuration)
  * Reject MTU > MAX_PACKET_SIZE (prevent buffer overflow)
  * Reject MTU < MIN_MTU (insufficient for fragmentation)

### DIFF
--- a/src/packet/error.rs
+++ b/src/packet/error.rs
@@ -11,19 +11,10 @@ pub enum PacketError {
     ErrH265CorruptedPacket,
     ErrInvalidH265PacketType,
     ErrH265PACIPHESTooLong,
-    InvalidMtu(MtuErrorKind),
     StapASizeLargerThanBuffer(usize, usize),
     NaluTypeIsNotHandled(u8),
     ErrVP9CorruptedPacket,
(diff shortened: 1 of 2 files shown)


algesten:
Hi @sr1990 thanks for looking at this!

To me this doesn't look like a runtime error. The MTU is calculated as a constant in do_payload, and it can't change runtime. Having the check is fine, but I think it should rather be an assert/panic than something that bubbles to the user.

What do you think?

sr1990:
Hi @algesten , makes sense. I have replaced MTU error handling with warnings. Please check.

algesten:
Thanks!

PR #851, opened 2026-02-01
More integration tests

algesten:


tests/dtls-security.rs
```diff
@@ -0,0 +1,195 @@
...
+use common::{init_crypto_default, init_log, progress, Peer, TestRtc};
+
+/// Test certificate fingerprint format and uniqueness.
+#[test]
+fn dtls_certificate_fingerprint_format() -> Result<(), RtcError> {
+    init_log();
+    init_crypto_default();
+
+    // Create two instances and verify they have different certificates
+    let rtc1 = RtcConfig::new().build(Instant::now());
```
xnorpx: Should all dtls-security use the R L peers so we can test with the difference crypto impl?

algesten: Sorted

PR #854, opened 2026-02-02
Fix two AV1 packetizer bugs causing panics on boundary-sized OBUs

farit2000:
## Summary

Two bugs in the AV1 packetizer that cause panics when packetizing real-world AV1 frames (payloads >= 128 bytes):

- **`leb_128_size()` off-by-one**: `while value > 0x80` should be `while value >= 0x80`. Values at LEB128 encoding boundaries (128, 16384, 2097152, ...) were allocated one byte too few, causing slice panics during packetization.

- **Continuation fragment `payload_size`**: For non-last OBUs spanning multiple RTP packets, `obu.payload.len() - obu_offset` used the OBU-level offset (including header bytes) instead of `payload_offset` (adjusted for header). This produced incorrect slice sizes and data corruption for OBUs with extension headers.

## Reproduction

Any AV1 video stream with frames >= 128 bytes (i.e., every real video call) triggers the `leb_128_size` bug. The `payload_offset` bug manifests when OBUs with extension headers are fragmented across multiple RTP packets.

The panic looks like:
```
panicked at str0m/src/packet/av1.rs:183:24: range end index 1000 out of range for slice of length 999
```

## Test plan

- [x] Added `leb128_size_boundary_at_128` — verifies correct byte count at 127/128 boundary
- [x] Added `leb128_size_boundary_at_16384` — verifies correct byte count at 16383/16384 boundary
- [x] Added `leb128_size_boundary_at_2097152` — verifies correct byte count at 2097151/2097152 boundary
- [x] Added `packetize_obu_with_128_byte_payload_no_panic` — verifies no panic for 128-byte OBU
- [x] Added `packetize_128_byte_obu_fragmented_correctly` — round-trip packetize/depacketize with MTU fragmentation, verifies payload integrity
- [x] Added `packetize_obu_with_extension_fragmented_round_trip` — round-trip for OBU with extension header, verifies payload integrity after fragmentation
- [x] All 31 tests pass (25 existing + 6 new)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 # 0.16.1
 
+  * Fix two AV1 packetizer panics #854
   * Fix ICE disconnections with Chrome v144+ when not using ICE Lite #846
   * Downgrade noisy BWE logs #847
   * Fix H265 PACI depacketizer panic on short payload #849


algesten:
Hi @farit2000, you can tell Claude to fix the CI errors ;)

algesten:
Thanks!

PR #858, opened 2026-02-05
Expose IceAgent::invalidate_candidate via DirectApi

alexlapa:
Expose `IceAgent::invalidate_candidate` via `DirectApi` so it would be possible to remove some local/remote candidate if you know something that `IceAgent` does not :)


src/change/direct.rs
```diff
@@ -5,6 +5,7 @@ use crate::rtp_::MidRid;
 use crate::rtp_::{Mid, Rid, Ssrc};
 use crate::sctp::ChannelConfig;
 use crate::streams::{StreamRx, StreamTx, DEFAULT_RTX_CACHE_DURATION, DEFAULT_RTX_RATIO_CAP};
+use crate::Candidate;
 use crate::IceCreds;
 use crate::Rtc;
 use crate::RtcError;
```
alexlapa, line 11: `imports_granularity = "Crate/Module"` (requires nightly)?

PR #859, opened 2026-02-07
[H265] Use is_irap() for keyframe detection.

sr1990:
Replace is_idr_picture() with is_irap() to detect all H.265 random access points (BLA, IDR, CRA) instead of just IDR frames.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * H265 to use is_irap() for keyframe detection #859
   * Fix probe estimator panic when probes accumulate #857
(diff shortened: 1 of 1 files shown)


algesten (approved):
I have no idea what you're doing, but I guess you do :)

sr1990:
> I have no idea what you're doing, but I guess you do :)

I’m still getting familiar with the project and RTC specs. Two additional HEVC-related improvements we need are:
- Improving the DONL logic and initializing the DONL-based packetizer and depacketizer using SDP.
- Refactoring PACI/TACI-related code and checking spec compliance.
 
Will look into these soon.

PR #864, opened 2026-02-10
Add AbsoluteCaptureTime extension and related functionality

lmiguelgato:
### Summary

  Implements the `abs-capture-time` RTP header extension as specified in the [WebRTC specification](https://webrtc.github.io/webrtc-org/experiments/rtp-hdrext/abs-capture-time/). This extension enables accurate audio/video synchronization in scenarios where media passes through RTCP-terminating mixers or other intermediary systems.

  ### Motivation

  The `abs-capture-time` extension addresses a critical limitation in multi-hop WebRTC scenarios. When RTP packets traverse mixers or SFUs that terminate RTCP (Sender Reports), receivers lose the ability to correlate timestamps from the original capture system, breaking A/V sync. This extension embeds the original capture timestamp in each RTP packet, preserving synchronization information end-to-end.

  **Key use cases:**
  - Multi-point conferencing through mixing servers
  - Cloud recording services that need accurate A/V sync
  - Cascaded SFU architectures
  - Any scenario with RTCP-terminating intermediaries

  ### Implementation Details

  #### Extension Format

  - **URI**: `http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time`
  - **Short format** (8 bytes): 64-bit NTP timestamp (UQ32.32 fixed-point)
  - **Extended format** (16 bytes): Timestamp + 64-bit signed clock offset

  #### Changes

  1. **New Extension Variant** (`src/rtp/ext.rs:27`)
     - Added `Extension::AbsoluteCaptureTime` enum variant

  2. **Extension Values** (`src/rtp/ext.rs:866-869`)
     - `abs_capture_time: Option<SystemTime>` - Absolute NTP timestamp of original frame capture
     - `abs_capture_clock_offset: Option<i64>` - Optional sender's clock offset estimate (raw NTP Q32.32 format)

  3. **Serialization/Parsing** (`src/rtp/ext.rs`)
     - Converts between `SystemTime` and 64-bit NTP format (UQ32.32) using `as_ntp_64()`/`from_ntp_64()`
     - Handles variable-length format (8 or 16 bytes)
     - Uses two-byte header form only when extension ID > 14 (16-byte payload fits in one-byte form)
     - Follows the exact pattern of `SenderInfo` in `src/rtp/rtcp/sr.rs`

  4. **Media Type Support**
     - Enabled for both audio and video streams
     - Follows same patterns as existing `abs-send-time` extension

  #### Design Decision: SystemTime vs Instant

  This implementation uses `SystemTime` for `abs_capture_time` rather than `Instant`:

  - **Why SystemTime**: The `abs-capture-time` extension represents an absolute NTP wall-clock timestamp that must be preserved across machines and network hops. `Instant` is a monotonic clock relative to an arbitrary local epoch and cannot represent absolute timestamps from remote systems.

  - **Precedent**: This follows the exact pattern of `SenderInfo.ntp_time` in `src/rtp/rtcp/sr.rs`, which also uses `SystemTime` for NTP timestamps.

  - **Simplification**: Using `SystemTime` enables direct NTP conversions (`as_ntp_64()`/`from_ntp_64()`) without complex epoch calculations.

  - **Note**: This differs from `abs-send-time`, which uses `Instant` because it's a 24-bit truncated value used only for relative timing/bandwidth estimation, not as an absolute timestamp.

  ### Testing

  - 3 new unit tests covering all format variations:
    - `abs_capture_time_short_form` - 8-byte format (timestamp only)
    - `abs_capture_time_extended_form` - 16-byte format with clock offset
    - `abs_capture_time_two_byte_form` - Two-byte header form (extension ID > 14)
  - 2 new integration tests (`tests/abs-capture-time.rs`):
    - `abs_capture_time_negotiation` - Verifies SDP negotiation and packet delivery
    - `abs_capture_time_sdp_roundtrip` - Verifies extension URI in SDP
  - All existing tests pass (527 library tests)
  - Timestamp accuracy verified within 1ms through roundtrip tests
  - No breaking changes to existing APIs

  ### Event Size Impact

  The addition of two `Option` fields to `ExtensionValues` increased the `Event` enum size by 2 bytes (470 → 472 bytes). Updated the size limit to 490 bytes to provide reasonable headroom for future additions.

  ### Usage Example

  ```rust
  use str0m::rtp::{Extension, ExtensionMap, ExtensionValues};
  use std::time::SystemTime;

  // Configure extension mapping (typically done during SDP negotiation)
  let mut exts = ExtensionMap::standard();
  exts.set(9, Extension::AbsoluteCaptureTime);

  // Set capture time when sending RTP packets
  let mut ext_vals = ExtensionValues::default();
  ext_vals.abs_capture_time = Some(SystemTime::now());

  // For mixers/SFUs: optionally include clock offset (raw NTP Q32.32 format)
  ext_vals.abs_capture_clock_offset = Some(estimated_offset_ntp);

  // Extension is automatically serialized into RTP packets
  // and parsed on the receiving end
```

  ### Bandwidth Considerations

  Per the specification, implementations should not send this extension with every packet to save bandwidth. Senders should include it:
  - Periodically (e.g., every second)
  - When the capture system changes
  - At the start of a stream

  This implementation provides the mechanism; rate limiting policy is left to the application layer.

  ### Compatibility

  - No breaking changes - purely additive
  - Applications not using this extension are unaffected
  - Follows existing extension architecture patterns
  - Compatible with Chrome, Firefox, and other WebRTC implementations

  ### References

  - [abs-capture-time specification](https://webrtc.github.io/webrtc-org/experiments/rtp-hdrext/abs-capture-time/)
  - [RFC 8285 - RTP Header Extensions](https://tools.ietf.org/html/rfc8285)
  - Related extension: `abs-send-time` (already implemented in str0m)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Add `abs-capture-time` RTP header extension #864
+
 # 0.16.2
 
   * Fix hot timeout loop with null pacer #863

### CODE REVIEW

**Comment by xnorpx** [comment]
@lmiguelgato also add an integration test (in the test folder)

File: src/lib.rs [hunk 1]
```diff
@@ -2035,7 +2035,8 @@ mod test {
     #[test]
     fn event_is_reasonably_sized() {
         let n = std::mem::size_of::<Event>();
-        assert!(n < 470); // Increased from 450 to accommodate H.265 profile/tier/level fields
+        eprintln!("Event size: {} bytes", n);
```
algesten: Nit: This `eprintln!` looks like a leftover debug statement. It would add noise to test output; please remove it.

src/rtp/ext.rs
```diff
@@ -140,6 +143,11 @@ pub trait ExtensionSerializer: Debug + Send + Sync + 'static {
 impl Extension {
     fn requires_two_byte_form(&self, ev: &ExtensionValues) -> bool {
         match self {
+            Extension::AbsoluteCaptureTime => {
+                // Extended format (16 bytes) exceeds the one-byte form maximum of 16 bytes.
+                // Actually, the max for one-byte is 16, so 16 bytes should fit, but let's be safe.
+                ev.abs_capture_clock_offset.is_some()
```
algesten: The comment contradicts itself ("16 bytes exceeds the one-byte form maximum of 16 bytes" → "Actually, the max for one-byte is 16, so 16 bytes should fit"). Per RFC 8285 §4.2, the one-byte form L field is 4 bits encoding `length - 1`, so it supports payloads of 1–16 bytes. The 16-byte extended format fits.

This means `requires_two_byte_form` returning `true` when `abs_capture_clock_offset.is_some()` is unnecessarily forcing two-byte headers, wasting 1 extra byte per extension element. The extended format (16 bytes) should work fine with one-byte form.

Consider either:
- Removing this override entirely (one-byte form handles 16 bytes), or
- If there's a specific interop reason to prefer two-byte form, documenting that reason instead of the current confused comment.

src/rtp/ext.rs
```diff
@@ -745,6 +784,40 @@ impl Extension {
...
             }
+            // 8 or 16
+            AbsoluteCaptureTime => {
+                // 64-bit NTP timestamp (UQ32.32 format).
+                // Optional 64-bit signed clock offset.
+                if buf.len() < 8 {
+                    return None;
+                }
+
+                // Parse the NTP timestamp
```
algesten: Nit: This can be simplified to:
```rust
let ntp_64 = u64::from_be_bytes(buf[..8].try_into().unwrap());
```
Same for the `i64` offset parsing below (`buf[8..16].try_into().unwrap()`). The explicit byte listing is more verbose without being clearer.

src/rtp/ext.rs
```diff
@@ -863,6 +936,10 @@ pub struct ExtensionValues {
     #[doc(hidden)]
     pub abs_send_time: Option<Instant>,
     #[doc(hidden)]
+    pub abs_capture_time: Option<Instant>,
```
algesten: The `abs_capture_clock_offset` field stores a raw `i64` but the spec defines it as a signed 64-bit NTP fixed-point number (seconds in upper 32 bits, fractional seconds in lower 32 bits). Without documentation, users may assume this is nanoseconds, milliseconds, or some other common unit.

Please add a doc comment clarifying the format, e.g.:
```rust
/// Clock offset in NTP fixed-point format (signed UQ32.32):
/// upper 32 bits are whole seconds, lower 32 bits are fractional seconds.
pub abs_capture_clock_offset: Option<i64>,
```

Alternatively, consider providing a helper to convert to/from `Duration` for ergonomics.

src/rtp/ext.rs
```diff
@@ -745,6 +784,40 @@ impl Extension {
...
+                ]);
+
+                // Convert NTP timestamp to SystemTime, then to Instant
+                let system_time = SystemTime::from_ntp_64(ntp_64);
+
+                // Convert SystemTime to Instant using our time conversion utilities
+                let duration_since_epoch = system_time
+                    .duration_since(SystemTime::UNIX_EPOCH)
+                    .unwrap_or(Duration::ZERO);
+                let instant = already_happened() + duration_since_epoch - epoch_to_beginning();
```
algesten: This Instant conversion logic is correct but there's a subtle difference from the write path worth noting. The write path uses `InstantExt::to_system_time()` which handles the case where `self < BEGINNING_OF_TIME.0` with a warning. The parse path here doesn't guard against that (e.g., if `duration_since_epoch < epoch_to_beginning()`, the subtraction would panic due to underflow on `Instant`).

In practice this would only happen with a bogus NTP timestamp (e.g., year < ~2020), but for robustness you may want a `checked_sub` or a saturating approach, similar to how `unwrap_or(Duration::ZERO)` is already used for the `SystemTime` conversion.

algesten:
## Fundamental type issue: `Instant` vs `SystemTime` for `abs_capture_time`

The most significant concern with this PR is that `abs_capture_time` is typed as `Option<Instant>`, but it should be `Option<SystemTime>`.

### Why `Instant` is wrong here

`Instant` is a monotonic clock relative to an arbitrary local epoch (typically process start or system boot). It cannot represent an absolute wall-clock time that originated on a remote system. The `abs-capture-time` extension exists specifically to carry the **original capture timestamp** across RTCP-terminating intermediaries — it's an absolute NTP wall-clock time by definition.

With `Instant`:
- The value is meaningless across machines (different epochs)
- Round-tripping through `Instant → SystemTime → NTP → SystemTime → Instant` introduces unnecessary conversion error and complexity (as seen in the parse path)
- A mixer/SFU forwarding this value would need to reconstruct it relative to its own `Instant` epoch, defeating the purpose
- The field cannot be meaningfully set by an application that receives a capture timestamp from an external source (e.g., a hardware encoder providing a wall-clock time)

### Why `SystemTime` is correct

`SystemTime` maps directly to NTP timestamps. The conversion is straightforward (`SystemTimeExt::as_ntp_64` / `SystemTime::from_ntp_64`), already used elsewhere in str0m (SR reports, XR blocks). There's no need for the `already_happened() + duration - epoch_to_beginning()` gymnastics in the parse path.

### Why `abs_send_time` using `Instant` is different

The existing `abs_send_time` uses `Instant` but that's a fundamentally different use case:
- It's a 24-bit truncated value that wraps every ~64 seconds
- It's used for **relative** timing (bandwidth estimation, REMB)
- It's rebased on receipt relative to local time anyway (`update_absolute_send_time`)
- It's not meant to be forwarded as an absolute timestamp

`abs-capture-time` has none of these properties — it's a full 64-bit NTP timestamp meant to be preserved end-to-end.

### Suggested change

```rust
pub abs_capture_time: Option<SystemTime>,
```

This simplifies both the write and parse paths and makes the API correct by construction.

algesten:
To clarify the `SystemTime` point above — `SenderInfo` in `src/rtp/rtcp/sr.rs` is the exact precedent to follow. It has the same wire format (64-bit NTP timestamp) and uses `SystemTime` directly:

```rust
// Field (sr.rs:29)
pub ntp_time: SystemTime,

// Write (sr.rs:75-76)
let mt = self.ntp_time.as_ntp_64();
buf[4..12].copy_from_slice(&mt.to_be_bytes());

// Parse (sr.rs:123-126)
let ntp_time = u64::from_be_bytes([buf[4], buf[5], buf[6], buf[7], buf[8], buf[9], buf[10], buf[11]]);
let ntp_time = SystemTime::from_ntp_64(ntp_time);
```

The `abs_capture_time` field should follow this same pattern — `Option<SystemTime>` with `as_ntp_64()` / `from_ntp_64()`. This would eliminate the `already_happened() + duration_since_epoch - epoch_to_beginning()` conversion and the potential `Instant` underflow issue.

algesten:
If you want to discuss stuff, you can reach me on Discord also.

src/rtp/ext.rs
lmiguelgato: `abs_capture_clock_offset` is a signed offset (can be negative if the sender's clock is behind the capture system's clock), so we cannot use `Duration` here.
But yes, I will add the clarifying comments, thanks.

src/rtp/ext.rs
lmiguelgato: Saturating approach that clamps to the minimum valid value (`already_happened()`) when the timestamp would underflow. Added test case `abs_capture_time_bogus_timestamp_no_panic`.

src/lib.rs
lmiguelgato: Removed.

src/rtp/ext.rs
lmiguelgato: Removed.

src/rtp/ext.rs
lmiguelgato: Done.

lmiguelgato:
> @lmiguelgato also add an integration test (in the test folder)

Added test case `abs_capture_time_negotiation` with `MediaKind::Audio`, where it's verified that the SDP string of the offer contains `abs-capture-time` extension, and that packets are flowing and received correctly.

Also added test case `abs_capture_time_sdp_roundtrip` with `MediaKind::Video`, where it's checked that both offer and answer contain `abs-capture-time`.

lmiguelgato:
> To clarify the `SystemTime` point above — `SenderInfo` in `src/rtp/rtcp/sr.rs` is the exact precedent to follow. It has the same wire format (64-bit NTP timestamp) and uses `SystemTime` directly:
> 
> ```rust
> // Field (sr.rs:29)
> pub ntp_time: SystemTime,
> 
> // Write (sr.rs:75-76)
> let mt = self.ntp_time.as_ntp_64();
> buf[4..12].copy_from_slice(&mt.to_be_bytes());
> 
> // Parse (sr.rs:123-126)
> let ntp_time = u64::from_be_bytes([buf[4], buf[5], buf[6], buf[7], buf[8], buf[9], buf[10], buf[11]]);
> let ntp_time = SystemTime::from_ntp_64(ntp_time);
> ```
> 
> The `abs_capture_time` field should follow this same pattern — `Option<SystemTime>` with `as_ntp_64()` / `from_ntp_64()`. This would eliminate the `already_happened() + duration_since_epoch - epoch_to_beginning()` conversion and the potential `Instant` underflow issue.

This makes total sense. Thanks for bringing it up. I have changed it to `SystemTime`, and now it's also not needed to do the underflow check to avoid panic with bogus timestamps. Thanks.

lmiguelgato:
Rebased and squashed my changes after addressing the code review comments. Updated PR description.

xnorpx:
change lgtm

xnorpx:
@algesten are you good with the changes?

xnorpx:
@algesten  changelog entry?

algesten:
Thanks!

PR #865, opened 2026-02-13
fix(ice): filter out discarded candidates

thomaseizinger:
When a local or remote candidate gets invalidated, str0m internally sets a `discarded` flag instead of removing the `Candidate`. This allows us to retain the indices into the `local_candidates` and `remote_candidates` lists.

Unfortunately, with the current getters of `local_candidates` and `remote_candidates`, this implementation detail leaks through because those discarded candidates still get returned there.

In order to provide a consistent view of the world to the user, we change the API to return an `Iterator` instead and filter the candidates by the `discarded` flag.


src/change/sdp.rs
thomaseizinger, line 1: I believe this may actually fix a bug here where we previously also included discarded candidates in the offer.

PR #866, opened 2026-02-14
Avoid sending VLA in every packet

Ramil980:
Hi @algesten,

At the moment, VLA is added to every packet in a frame. However, based on the [libwebrtc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/rtp_rtcp/source/rtp_sender_video.cc;drc=c3a984cfaa29743e47010e7ac2db2ef45eab831a;l=499) implementation, it should only be included in the first packet of the frame. This PR fixes that behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Avoid sending VLA on every packet #866
   * Add `abs-capture-time` RTP header extension #864
 
 # 0.16.2


xnorpx:
@algesten  I think it looks resonable. But this would not work if we had if a user brought their own RTP header extension?

algesten:
> @algesten I think it looks resonable. But this would not work if we had if a user brought their own RTP header extension?

What do you mean?

xnorpx:
> > @algesten I think it looks resonable. But this would not work if we had if a user brought their own RTP header extension?
> 
> What do you mean?

```rust
  let mut pkt_ext_vals = ext_vals.clone();
            if !first {
                pkt_ext_vals.user_values.remove::<VideoLayersAllocation>();
            }
```

I read it as it will remove VideoLayesAllocation if it's not the first package.

If I create my own say DD header extension which should also only be on the first packet (similar to VLA)

I do not contribute this back to str0m, how can I avoid to get it attached to all packets in the frame?

algesten:
@xnorpx the change only affected the Media level API, where it attaches the VLA while packetizing. With RTP-mode you can send anything you want on any packet.

xnorpx:
@algesten yes, but I am talking about other header extensions on Media level API.

But we can just contribute those back and add in similar way to VLA (this is not a big blocker just pointing it out)

Ramil980:
Thanks for merging!

@algesten @xnorpx, I can pull the list of extensions from libwebrtc that are not send in all packets and fix the behaviour here. For example, from a quick check, Absolute Capture Time is only sent on the first packet, Video Orientation is only sent on the last packet, and so on

PR #868, opened 2026-02-15
VP9: switch packetizer to non-flexible mode (F=0)

farit2000:
We've been running str0m as a WebRTC SFU in rtp_mode and hit an issue: Safari drops all VP9 inter-frames when the packetizer uses flexible mode (F=1). After debugging, we found that Safari's VP9 decoder expects non-flexible mode with `tl0picidx` for temporal reference tracking.

Switching to F=0 (non-flexible mode) fixes Safari and is also what other SFUs like LiveKit and mediasoup use in practice.

### What changed

- Packetizer now emits F=0 with layer indices (L=1) and `tl0picidx` on every packet
- Keyframe first-packets include scalability structure (V=1) with a minimal GOF
- Added `detect_vp9_keyframe_bitstream()` to correctly set the P bit from raw VP9 data
- `tl0picidx` increments on every frame (all TID=0) and wraps at 255→0, per the spec
- `VP9HEADER_SIZE` bumped to 8 to account for the larger header on keyframes

The depacketizer already handles both modes, so no changes needed there.

### Testing

- Updated existing packetizer tests for the new header format
- Added tests for keyframe detection (all 4 VP9 profiles), fragmented keyframes with round-trip verification, and tl0picidx wrapping
- Full test suite passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * VP9 switch packetizer to non-flexible mode (F=0)
   * Fix H265 SDP negotiation level handling #872
   * Update dimpl to 0.3.0 #874
   * Avoid sending VLA on every packet #866


algesten:
@farit2000 Great find! A little cargo fmt and we merge!

algesten:
@farit2000 thinking about this. Safari limitations might guide our default to `F=0`, but should we really throw flexible mode out? Seems like a config option.

@davibe thoughts?

farit2000:
> @farit2000 thinking about this. Safari limitations might guide our default to `F=0`, but should we really throw flexible mode out? Seems like a config option.
> 
> @davibe thoughts?

Makes sense — we could keep both and add a config flag. In our SFU we always use F=0 because Chrome/Safari both handle it well, but having flexible mode as an option is reasonable. I can refactor this so Vp9Packetizer has a mode field (Flexible / NonFlexible) if you'd like.

algesten:
@farit2000 perfect! Let's keep `F=1` around, but switch the default to `F=0`

farit2000:
> @farit2000 perfect! Let's keep `F=1` around, but switch the default to `F=0`

Done - kept both modes with F=0 as default. Added Vp9PacketizerMode enum (Flexible / NonFlexible) and wired it through 
RtcConfig:
  RtcConfig::new()
      .set_vp9_packetizer_mode(Vp9PacketizerMode::Flexible)

farit2000:
@algesten Rebased on main, CI is green. Both modes are implemented with F=0 as default per your suggestion. Ready for review when you have a chance!

PR #870, opened 2026-02-15
feat(streams): use bytes::Bytes for RTP packet payload

farit2000:
We use str0m as an SFU and forward each incoming packet to all other participants. Cloning Vec<u8> per subscriber gets expensive in larger rooms, so this switches RtpPacket.payload to bytes::Bytes — cloning is just a refcount bump.

write_rtp() takes `impl Into<Bytes>` now so passing Vec<u8> still works as before.

This is a breaking change for anyone constructing RtpPacket directly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1773,6 +1773,7 @@ version = "0.16.2"
 dependencies = [
  "_str0m_test",
  "arrayvec",
+ "bytes",
  "combine",
  "crc",
  "dimpl",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,7 @@ sctp-proto = "0.7.0"
 combine = "4.6.6"
 subtle = "2.0.0"
 arrayvec = "0.7.6"
+bytes = "1"
 
 # STUN
 # Pin crc to avoid 3.4 which requires rustc 1.83
--- a/src/_internal_test_exports/fuzz.rs
+++ b/src/_internal_test_exports/fuzz.rs
@@ -107,7 +107,7 @@ pub fn depack(data: &[u8]) -> Option<()> {
             };
             let len = rng.usize(1200)?;
             let data = rng.slice(len)?.to_vec();
-            depack.push(meta, data);
+            depack.push(meta, data.into());
         } else {
             depack.pop();
         }
--- a/src/packet/buffer_rx.rs
+++ b/src/packet/buffer_rx.rs
@@ -3,6 +3,8 @@ use std::fmt;
 use std::ops::{Range, RangeInclusive};
 use std::time::Instant;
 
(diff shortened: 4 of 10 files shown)


xnorpx:
This one will probably be rejected (I know @algesten is not a fan of Bytes crate and it has been discussed before).

Short term solution, use allocator crate such as MiMalloc or alternatives.

I think we discussed to change the API to allow to pass in vectors to allow for external pooling and reuse.

algesten:
Yeah, sorry. It's one of my vetoed types in public API. I want simple Rust built-in buffers. We do have it in tree already, but I want to get rid of it.

There could be arguments for `Arc<[u8]>` or similar, but let's discuss before making.

k0nserv:
@algesten If you outline the issues you have with Bytes, maybe there's a subset of it we could implement for str0m. `Arc<[u8]>` is an improvement on `Vec<u8>` in some cases, but at the very least it's missing a cheap way to sub-slice.

algesten:
> @algesten If you outline the issues you have with Bytes, maybe there's a subset of it we could implement for str0m. `Arc<[u8]>` is an improvement on `Vec<u8>` in some cases, but at the very least it's missing a cheap way to sub-slice.

The biggest problem is that it's not in `std` and that std APIs operate on `&[u8]` (and to some extent `Vec<u8>`). Like `io::Read` and `io::Write`. I think libraries should mainly use `std` types where there is a standard type for it. 

On top of that, `Bytes` is quite "infectious". If you depend on one library using it and others not, conversions are a must. Like we have to handle it in `sctp-proto` and going between `std` types and `Bytes` typically means expensive memcopies.

alexlapa:
Aha, so Bytes come up again :)

Some previous discussion from discord:

> Another low hanging fruit in terms of reducing allocations might be accepting payload as Bytes instead of Vec<u8> in StreamTx::write_rtp

> Maybe `enum RtpPayload { Arc(Arc<[u8]>), Vec(Vec<u8>) }`, might have to recheck but i think the only thing required is for it to be AsRef<[u8]>. It should not need any mutabilty, encryption is not in-place. This way you might choose what to use depending on how many remote receivers there are.

> Box<[u8]> should be btter then Vec for transmits/RtpPacket.payload. More correct since its not expected to be modified. Also size_of::<usize>() smaller (does not really matter but still :)).

@k0nserv

> Arc<[u8]> is an improvement on Vec<u8> in some cases, but at the very least it's missing a cheap way to sub-slice.

Well that would be pretty easy to implement yout own slices on top of Arc<[u8]>, but Im not sure sub-slicing is really used anywhere right now for it to be a concern? Well at least in RTP-mode that Im using. Maybe cheap subslicing might fit nicely into how packetizers work.

PR #871, opened 2026-02-15
fix(session): use VecDeque for pending RTP packets in rtp_mode

farit2000:
In rtp_mode, pending_packet is an Option<RtpPacket>. If you call handle_input(Receive) twice before poll_output(), the second packet overwrites the first and it's silently lost.

Switched to VecDeque<RtpPacket> so all packets are queued and drained one by one via pop_front() in poll_output(). VecDeque is already imported in session.rs for other fields.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -89,8 +89,10 @@ pub(crate) struct Session {
     // temporary buffer when getting the next (unencrypted) RTP packet from Media line.
     poll_packet_buf: Vec<u8>,
 
-    // Next packet for RtpPacket event.
-    pending_packet: Option<RtpPacket>,
+    // Queue of packets for RtpPacket events (rtp_mode).
+    // Must be a VecDeque because multiple handle_input(Receive) calls can happen
+    // before the next poll_output(), and we must not drop any packets.
+    pending_packets: VecDeque<RtpPacket>,
 
     // Whether we sent a single outgoing RTP packet.
     packet_first_sent: bool,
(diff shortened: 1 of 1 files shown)


algesten:
> If you call handle_input(Receive) twice before poll_output()

This is incorrect API usage. Every "mutation" must be followed by poll-to-timeout before you do another mutation.

That's the only way str0m doesn't need buffers everywhere internally (like you notice here).

This might need to be more obviously documented. I have also tried (and failed so far) to enforce the poll-to-timeout both in types and code. The problem comes with stuff like "does adding an ICE candidate mean we did a mutation"?

lherman-cs:
@algesten, curious about this as well. Doesn't "mutation" here also include stream's write_rtp? Since write_rtp already uses a sending buffer to allow multiple write_rtp calls before polling rtc for outputs, it seems appropriate to have a buffer for ingress too? 

For my use case, the receiver loop gets batches of UDP packets. Ideally, I want to feed a batch of these packets to str0m to transfer ownership, then poll in a hot loop.

algesten:
@lherman-cs I recently tried to go "The Other Way" of trying to enforce the Sans-IO contract more strictly using API https://github.com/algesten/str0m/pull/840

I gave up because the API got too unwieldy.

I think overall there are two categories of mutations. Stuff you really do need to handle straight away (SDP changes), and stuff we could allow more buffering in str0m side like RTP and/or datachannel writing.

Consistency is what we should aim for.

Issue #860, opened 2026-02-08
H265 SDP negotiation: level-id=180 from latest Chrome


showier-drastic:
This is what I get from Chrome on macOS `Version 144.0.7559.133 (Official Build) (arm64)`

```
a=rtpmap:49 H265/90000r
a=rtcp-fb:49 goog-rembr
a=rtcp-fb:49 transport-ccr
a=rtcp-fb:49 ccm firr
a=rtcp-fb:49 nackr
a=rtcp-fb:49 nack plir
a=fmtp:49 level-id=180;profile-id=1;tier-flag=0;tx-mode=SRSTr
a=rtpmap:50 rtx/90000r
a=fmtp:50 apt=49r
a=rtpmap:51 H265/90000r
a=rtcp-fb:51 goog-rembr
a=rtcp-fb:51 transport-ccr
a=rtcp-fb:51 ccm firr
a=rtcp-fb:51 nackr
a=rtcp-fb:51 nack plir
a=fmtp:51 level-id=180;profile-id=2;tier-flag=0;tx-mode=SRSTr
a=rtpmap:52 rtx/90000r
a=fmtp:52 apt=51r
```

In the current code, H265 level is 156: https://github.com/algesten/str0m/blob/5ff9b8f4adbfecaecdfc28cbb66b44aeb476ef51/src/format/codec_config.rs#L276

Therefore, negotiation fails. If I change the level-id to 180 manually in `codec_config.rs`, negotiation works.

algesten:
Hi @showier-drastic! I haven't been deeply involved in the H265 stuff that's been happening. However I know @sr1990 has worked hard on these things. This PR which is very new, is working on the matching https://github.com/algesten/str0m/pull/853

It's still only in the main branch. Could you test against latest main to see if this is still an issue?

showier-drastic:
@algesten Thanks for the quick reply! Actually I'm testing with the main branch, so even with this PR it doesn't work.

showier-drastic:
Actually it sounds a little bit strange to me:

https://github.com/algesten/str0m/pull/853/changes#diff-3fb102dce4e1ebfbfcf2c341145d7bb2ae21ae9faed10376e74d6771d8872a57R438

```
    /// The level check `c1_level <= c0_level` enforces that the remote encoder
    /// must not exceed our decoder's configured level.
```

It seems that the code above assumes that the remote is the encoder and str0m is the decoder.

[...]

PR #872, opened 2026-02-17
[H265] Fix SDP negotiation: remove level rejection, narrow to min(loc…

sr1990:
…al,remote), bump default to Level 6.0

This addresses issue https://github.com/algesten/str0m/issues/860.


algesten:
Looks good! Thanks!

PR #873, opened 2026-02-18
Fix infinite loop when ErrStreamAlreadyExist occurs for in-band data channels

timwu20:
### Summary
- Fix three error paths in `RtcSctp::do_poll()` where `continue` skips close handling, causing entries to stay in `AwaitOpen` state indefinitely
- Affected entries are now immediately transitioned to `Closed` and emit `SctpEvent::Close`

### Problem
When an in-band (DCEP) data channel fails to open with `ErrStreamAlreadyExist`, the stream entry has `do_close = true` set but the `continue` statement skips the close handling at the bottom of the loop. Since `state` remains `AwaitOpen`, subsequent polls retry the open — creating an infinite loop of warnings and eventually causing a panic (`writing dcep open: ErrPayloadDataStateNotExist`).

This is triggered when both peers try to open data channels simultaneously and the automatic stream ID allocation conflicts with a stream already opened by the remote peer.

### Fix
In `do_poll()`, three error paths inside the `if want_open` block used `continue` after setting `do_close = true`, which skipped the close-handling code further down the loop body:

1. `configure_reliability` failure
2. `ErrStreamAlreadyExist` for in-band channels
3. Generic `open_stream` errors

All three now transition the entry to `Closed` and return `SctpEvent::Close` directly instead of `continue`-ing.

### Testing
- Added regression test `err_stream_already_exist_in_band_returns_close` that verifies `do_poll()` returns `SctpEvent::Close` (confirmed test fails without the fix)


algesten:
Hi there! Thanks for this. Can you add this line at the top of the CHANGELOG.md?

```
* Fix infinite loop on in-band data channel stream conflict #873
```

algesten:
Probably just the changelog conflicting. Try a rebase.

timwu20:
> Probably just the changelog conflicting. Try a rebase.

I just fixed it.

algesten:
Thanks!

PR #874, opened 2026-02-20
Update dimpl

xnorpx:
## Update dimpl to 0.3.0

Updates the dimpl DTLS library from 0.2.4 to 0.3.0, which introduces DTLS 1.3 and auto-detect support.

The following is omitted from this pr to keep the size down, so follow-up pr's will be:

- Expand test to exercise DTLS 1.3
- Expand nightly test to exercise DTLS.13
- OpenSSL crypto backend DTLS 1.3 support with dimpl (Until OpenSSL releases DTLS 1.3 support)
- Wincrypto crypto backend DTLS 1.3 support with dimpl

### Changes

**New `DtlsVersion` config option**
- Added `DtlsVersion` enum (`Dtls12`, `Dtls13`, `Auto`) to `RtcConfig`
- Defaults to `Dtls12` (no behavior change)
- Configurable via `RtcConfig::set_dtls_version()`
- Marked `#[non_exhaustive]` for future extensibility

**Updated `DtlsProvider` trait**
- `new_dtls()` now takes `now: Instant` and `dtls_version: DtlsVersion` parameters
- `Instant` is always passed from the caller (no `Instant::now()` inside str0m)

**Backend support matrix**

| Backend | DTLS 1.2 | DTLS 1.3 | Auto |
|---------|----------|----------|------|
| aws-lc-rs | Yes | Yes | Yes |
| rust-crypto | Yes | Yes | Yes |
| apple-crypto | Yes | Yes | Yes |
| openssl | Yes | Error | Error |
| wincrypto | Yes | Error | Error |

**apple-crypto DTLS 1.3 support**
- Implemented HKDF provider (extract/expand/expand-label) using CommonCrypto HMAC
- Added DTLS 1.3 cipher suites (`TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384`)
- AES-ECB for record number encryption via existing CommonCrypto FFI
- Updated dimpl provider to new API (`Buf`/`TmpBuf` re-exports, `Dtls12CipherSuite` rename, `SupportedDtls12CipherSuite` trait rename, removed `is_compatible`)
- Added `unsafe impl Send/Sync` for hash contexts (CommonCrypto `*mut c_void`)

**dimpl API migration**
- `Dtls::new()` → `Dtls::new_12()` / `new_13()` / `new_auto()` with `Instant` param
- `SrtpProfile` variants renamed: `Aes128CmSha1_80` → `AES128_CM_SHA1_80`, etc.
- `dimpl::buffer::Buf` → `dimpl::crypto::Buf` (buffer module now private)

Nightly test passed: https://github.com/algesten/str0m/actions/runs/22261201569


src/packet/vp9.rs
```diff
@@ -876,6 +876,7 @@ mod test {
             ),
             (
                 "ParseRefIdx",
+                #[allow(clippy::erasing_op, clippy::identity_op)]
```
xnorpx: revert, not relevant for this pr

src/rtp/rtcp/twcc.rs
```diff
@@ -2115,8 +2115,7 @@ mod test {
         }
 
         now = now + Duration::from_millis(5);
-        #[allow(unused_must_use)]
```
xnorpx, line 2118: not related remove this

algesten:
OpenSSL Cargo.lock version downgrade

  The crypto/openssl/Cargo.lock changed from version = 4 to version = 3. Lockfile v4 was introduced in Rust 1.78 (behind nightly
  flag) and v3 is stable. This is fine but it means the openssl lockfile was regenerated, which also updated many transitive
  dependency versions and switched dimpl from crates.io to a git revision:

```
  -version = "0.2.3"
  -source = "registry+https://github.com/rust-lang/crates.io-index"
  +version = "0.2.7"
  +source = "git+https://github.com/algesten/dimpl?rev=9896731f..."
```

  Concern: The OpenSSL sub-crate's dimpl is pinned to a git revision (0.2.7) rather than the crates.io 0.3.0 used everywhere else.
  This appears to be because the OpenSSL sub-crate has its own Cargo.toml which may not have been updated, or because dimpl 0.3.0
  isn't compatible with the reduced dependency set in the OpenSSL backend. This inconsistency should be clarified - is this
  intentional? Should the OpenSSL Cargo.lock even be checked in?

xnorpx:
> OpenSSL Cargo.lock version downgrade
> 
> The crypto/openssl/Cargo.lock changed from version = 4 to version = 3. Lockfile v4 was introduced in Rust 1.78 (behind nightly flag) and v3 is stable. This is fine but it means the openssl lockfile was regenerated, which also updated many transitive dependency versions and switched dimpl from crates.io to a git revision:
> 
> ```
>   -version = "0.2.3"
>   -source = "registry+https://github.com/rust-lang/crates.io-index"
>   +version = "0.2.7"
>   +source = "git+https://github.com/algesten/dimpl?rev=9896731f..."
> ```
> 
> Concern: The OpenSSL sub-crate's dimpl is pinned to a git revision (0.2.7) rather than the crates.io 0.3.0 used everywhere else. This appears to be because the OpenSSL sub-crate has its own Cargo.toml which may not have been updated, or because dimpl 0.3.0 isn't compatible with the reduced dependency set in the OpenSSL backend. This inconsistency should be clarified - is this intentional? Should the OpenSSL Cargo.lock even be checked in?

ugh slop will fix

xnorpx:
@algesten ok I tried to update all lockfiles, this would be simpler if we had a workspace. But workspaces has other issues so like cross platform builds which would be a pain with applecrypto and wincrypto

crypto/rust-crypto/src/dtls.rs
```diff
@@ -41,7 +47,17 @@ impl DtlsProvider for RustCryptoDtlsProvider {
...
+        let dtls = match dtls_version {
+            DtlsVersion::Dtls12 => dimpl::Dtls::new_12(config, dimpl_cert, now),
+            DtlsVersion::Dtls13 => dimpl::Dtls::new_13(config, dimpl_cert, now),
+            DtlsVersion::Auto => dimpl::Dtls::new_auto(config, dimpl_cert, now),
+            _ => {
+                return Err(CryptoError::Other(format!(
+                    "Unsupported DTLS version: {dtls_version}"
+                )))
+            }
+        };
```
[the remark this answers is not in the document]
xnorpx: No we want it

Cargo.toml
```diff
@@ -57,7 +57,10 @@ crc = ">=3.0, <3.4"
 serde = { version = "1.0.152", features = ["derive"] }
 
 # DTLS made for str0m (required dependency)
-dimpl = { version = "0.2.4", default-features = false }
+dimpl = { version = "0.3.0", default-features = false }
+
+# Pin time to avoid time-core 0.1.8+ which requires edition 2024 (rustc 1.85+)
+time = ">=0.3, <0.3.37"
```
thomaseizinger, line 63: This breaks updating str0m for me:

```
thomas@fedora ~/s/g/f/f/rust (bump-str0m)> cargo update -p str0m
    Updating git repository `https://github.com/algesten/str0m`
    Updating crates.io index
error: failed to select a version for `time`.
    ... required by package `str0m v0.16.2 (https://github.com/algesten/str0m?branch=main#487afbe2)`
    ... which satisfies dependency `str0m = "^0.16.2"` of package `connlib-model v0.1.0 (/home/thomas/src/github.com/firezone/firezone.bump-str0m/rust/libs/connlib/model)`
    ... which satisfies path dependency `connlib-model` (locked to 0.1.0) of package `client-ffi v0.1.0 (/home/thomas/src/github.com/firezone/firezone.bump-str0m/rust/client-ffi)`
versions that meet the requirements `>=0.3, <0.3.37` are: 0.3.36, 0.3.35, 0.3.34, 0.3.33, 0.3.32, 0.3.31, 0.3.30, 0.3.29, 0.3.28, 0.3.27, 0.3.26, 0.3.25, 0.3.24, 0.3.23, 0.3.22, 0.3.21, 0.3.20, 0.3.19, 0.3.18, 0.3.17, 0.3.16, 0.3.15, 0.3.14, 0.3.13, 0.3.12, 0.3.11, 0.3.10, 0.3.9, 0.3.7, 0.3.6, 0.3.5, 0.3.4, 0.3.3, 0.3.2, 0.3.1, 0.3.0

all possible versions conflict with previously selected packages.

  previously selected package `time v0.3.46`
    ... which satisfies dependency `time = "^0.3.46"` of package `client-shared v0.1.0 (/home/thomas/src/github.com/firezone/firezone.bump-str0m/rust/libs/client-shared)`
    ... which satisfies path dependency `client-shared` (locked to 0.1.0) of package `client-ffi v0.1.0 (/home/thomas/src/github.com/firezone/firezone.bump-str0m/rust/client-ffi)`

failed to select a version for `time` which could resolve this conflict
```

To stay on a certain MSRV, it is generally better to use the lockfile and use `--precise` to select a particular version instead of restricting it in `Cargo.toml` because it breaks things for people that are already on the newer version.

algesten:
@thomaseizinger is there a reason you need that new version of `time` crate?

PR #875, opened 2026-02-23
fix(ice): compute ICE timeout for successful pair

thomaseizinger:
Currently, `IceAgent::ice_timeout` computes the ICE timeout for a pair that was never successful, i.e. has 0 successful attempts. This is not particularly useful. Instead, we most likely want to know, how long after the last successful binding request a pair is declared failed.

To do that, we need to effectively skip the first two `stun_resend_delay`s as those will never apply:

- If we do not have a previous attempt, `stun_resend_delay` is never invoked, thus there will never be a time where we check it for 0 unanswered requests.
- If we only have 1 previous attempt but are currently in a successful state, the first request will only timeout after MAX_RTO.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
   * Add `abs-capture-time` RTP header extension #864
+  * Adjust `IceAgent::ice_timeout` to return timeout of successful pair #875
 
 # 0.16.2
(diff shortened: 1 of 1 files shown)


algesten (approved):

algesten:
LGTM. Changelog!

PR #877, opened 2026-02-28
DTLS 1.3 for openssl crypto

xnorpx:


xnorpx:
Let's onboard tests first

crypto/openssl/src/dimpl_provider/cipher_suite.rs
```diff
@@ -0,0 +1,1110 @@
...
+            openssl::cipher::Cipher::aes_256_gcm()
+        }
+    }
+}
+
+impl Drop for AesGcm {
+    fn drop(&mut self) {
+        for b in self.key.iter_mut() {
+            // SAFETY: Volatile write prevents the compiler from eliding this zeroing.
+            unsafe { std::ptr::write_volatile(b, 0) };
```
xnorpx, line 145: Hmm not sure about this one @algesten  thoughts?

algesten:
@xnorpx the weird thing about this PR is that it's not clear which DTLS 1.2 implementation you end up with.

If you pick DtlsVersion::Dtls12, you get OpenSSL's native DTLS 1.2. If you pick DtlsVersion::Auto and the peer only supports 1.2, you get dimpl's DTLS 1.2 implementation (with OpenSSL crypto underneath). Two completely different code paths for the same protocol version.

xnorpx:
> @xnorpx the weird thing about this PR is that it's not clear which DTLS 1.2 implementation you end up with.
> 
> If you pick DtlsVersion::Dtls12, you get OpenSSL's native DTLS 1.2. If you pick DtlsVersion::Auto and the peer only supports 1.2, you get dimpl's DTLS 1.2 implementation (with OpenSSL crypto underneath). Two completely different code paths for the same protocol version.

Yes I agree this is funky. I discussed this with Eduardo and I think it makes sense to just scrap dtls 1.2 here and in Wincrypto so we just use one dtls implementation.

We can revisit dtls 1.3 in OpenSSL the day it get's released but yes I think we should remove it from OpenSSL crypto and Wincrypto for now.

Will do this update

algesten:
> We can revisit dtls 1.3 in OpenSSL the day it get's released but yes I think we should remove it from OpenSSL crypto and Wincrypto for now.
> 
> Will do this update

Sounds good!

crypto/openssl/src/dimpl_provider/cipher_suite.rs
xnorpx: I think it's a decent tradeoff

algesten:
@xnorpx I think we should remove OpenSSL DTLS 1.2 and only rely on dimpl. One single code path for DTLS 1.2.

xnorpx:
> @xnorpx I think we should remove OpenSSL DTLS 1.2 and only rely on dimpl. One single code path for DTLS 1.2.

@algesten my bad fixing

src/lib.rs
```diff
@@ -1700,20 +1700,18 @@ impl Rtc {
...
     /// # use std::time::Instant;
-    /// let mut rtc = Rtc::new(Instant::now());
+    /// // let mut rtc = Rtc::new(Instant::now());
     ///
-    /// let output = rtc.poll_output().unwrap();
+    /// // let output = rtc.poll_output().unwrap();
     ///
     /// // Reason updates every time we get an Output::Timeout
-    /// assert!(matches!(output, Output::Timeout(_)));
+    /// // assert!(matches!(output, Output::Timeout(_)));
```
xnorpx: revert this once separate fix is done

xnorpx: #898

xnorpx:
Let's fix #898 first

efer-ms:
> > @xnorpx the weird thing about this PR is that it's not clear which DTLS 1.2 implementation you end up with.
> > If you pick DtlsVersion::Dtls12, you get OpenSSL's native DTLS 1.2. If you pick DtlsVersion::Auto and the peer only supports 1.2, you get dimpl's DTLS 1.2 implementation (with OpenSSL crypto underneath). Two completely different code paths for the same protocol version.
> 
> Yes I agree this is funky. I discussed this with Eduardo and I think it makes sense to just scrap dtls 1.2 here and in Wincrypto so we just use one dtls implementation.
> 
> We can revisit dtls 1.3 in OpenSSL the day it get's released but yes I think we should remove it from OpenSSL crypto and Wincrypto for now.
> 
> Will do this update

I don't think we're in alignment exactly. I think long term it makes sense to me to use dimpl for everything. But right now we are not approved for that, I'd imagine others may be in the same boat. Using platform native protocols where available is desired, scrapping them for a local impl may require approvals.

So my opinion is that if DTLS 1.2-only is requested, we should use the platform APIs (OpenSSL [Linux], SChannel [Windows]). We can make "Auto/1.3" use dimpl.

Another option would be to put the non-dimpl versions also behind a feature flag. If we roll the str0m version to indicate an incompatiblity. Like I don't want someone caught off guard by this change as it has compliance ramifications that should get looked at.

xnorpx:
Reverting back to draft for now

algesten:
@efer-ms @xnorpx I don't care if we keep the OpenSSL 1.2 or not, but feature gate it.

We should remember that DTLS 1.3 is still unusual in WebRTC, the defaults everywhere should remain 1.2 across the board.

PR #879, opened 2026-03-01
Update dimpl

xnorpx:
### What

Bumps the dimpl DTLS library from 0.3.0 to 0.4.0 across all crates (str0m, proto, apple-crypto, aws-lc-rs, rust-crypto) and adds DTLS 1.3 handshake test coverage.

### Key Changes

**dtls.rs** — `handle_input` now distinguishes `dimpl::Error::HandshakePending` from other errors, mapping it to `WouldBlock` so the SCTP transmit layer can push packets back and retry rather than treating it as a fatal error.

**lib.rs** — Adds a catch-all arm for `DtlsOutput` polling (required because `Output` is now `#[non_exhaustive]` in dimpl 0.4.0). Also imports `DtlsError` directly.

**srtp.rs, dtls.rs** — Add `_ => panic!()` wildcard arms for `SrtpProfile` matches (also now `#[non_exhaustive]`).

**handshake-direct.rs** — Replaces the single `handshake_direct_api_two_threads` test with 7 parameterized tests covering the DTLS version matrix:

| Client \ Server | Dtls12 | Dtls13 | Auto |
|---|---|---|---|
| **Dtls12** | `12_to_12` | — | `12_to_auto` |
| **Dtls13** | — | `13_to_13` | `13_to_auto` |
| **Auto** | `auto_to_12` | `auto_to_13` | `auto_to_auto` |

Also adds: per-peer crypto provider selection via `L_CRYPTO`/`R_CRYPTO` env vars, optional PCAP capture (`SAVE_PCAP`), `rtc_built` timing, and fault-tolerant packet capture (packets returned even on test failure).

### Gaps
**wincrypto and openssl have zero DTLS 1.3/Auto coverage** — All 5 tests involving `Auto` or `Dtls13` are silently skipped (`return Ok(())`) for these providers. That's 5 of 7 tests skipped with no `#[ignore]` annotation or CI indicator.

**Known dimpl limitation: Auto client → DTLS-1.2-only server** — Explicitly skipped with the comment that dimpl advertises X25519 in the hybrid ClientHello but can't process X25519 in `ServerKeyExchange`. This is a dimpl bug, not a str0m concern, but it means `Auto` client negotiation isn't truly auto when the server is 1.2-only.

### DIFF
--- a/src/dtls.rs
+++ b/src/dtls.rs
@@ -137,9 +137,13 @@ impl Dtls {
 
     /// Send application data over DTLS.
     pub fn handle_input(&mut self, data: &[u8]) -> Result<(), DtlsError> {
-        self.instance
-            .send_application_data(data)
-            .map_err(|e| DtlsError::CryptoError(CryptoError::Other(format!("DTLS error: {}", e))))
+        self.instance.send_application_data(data).map_err(|e| {
+            if matches!(e, dimpl::Error::HandshakePending) {
+                DtlsError::Io(io::Error::new(io::ErrorKind::WouldBlock, e))
+            } else {
+                DtlsError::CryptoError(CryptoError::Other(format!("DTLS error: {}", e)))
+            }
+        })
     }
 
     /// Handle a timeout event.
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -661,7 +661,7 @@ use crypto::Fingerprint;
 
 mod dtls;
 use crate::crypto::dtls::DtlsOutput;
-use crate::crypto::{from_feature_flags, CryptoProvider};
+use crate::crypto::{from_feature_flags, CryptoProvider, DtlsError};
 use crate::dtls::is_would_block;
 use dtls::Dtls;
 
@@ -1549,6 +1549,11 @@ impl Rtc {
                 DtlsOutput::Timeout(t) => {
                     self.next_dtls_timeout = Some(t);
                     break;
+                }
+                other => {
+                    return Err(RtcError::Dtls(DtlsError::Io(std::io::Error::other(
+                        format!("Unexpected DTLS output: {other:?}"),
+                    ))));
                 }
             }
         }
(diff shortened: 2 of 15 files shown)


src/lib.rs
```diff
@@ -661,6 +661,7 @@ use crypto::Fingerprint;
 
 mod dtls;
 use crate::crypto::dtls::DtlsOutput;
+use crate::crypto::DtlsError;
```
xnorpx: merge import?

src/dtls.rs
```diff
@@ -137,9 +137,13 @@ impl Dtls {
 
     /// Send application data over DTLS.
     pub fn handle_input(&mut self, data: &[u8]) -> Result<(), DtlsError> {
```
xnorpx, line 139: @algesten this part is the only production code change.  So this should be fairly small to review and safe.

src/dtls.rs
```diff
@@ -137,9 +137,13 @@ impl Dtls {
...
-        self.instance
-            .send_application_data(data)
-            .map_err(|e| DtlsError::CryptoError(CryptoError::Other(format!("DTLS error: {}", e))))
+        self.instance.send_application_data(data).map_err(|e| {
+            if matches!(e, dimpl::Error::HandshakePending) {
+                DtlsError::Io(io::Error::new(io::ErrorKind::WouldBlock, e))
```
algesten, line 142: How does this help? Don't we need to buffer somewhere?

xnorpx: Yes, caller is putting them back

https://github.com/algesten/str0m/blob/96a88053623fda8df2243d94d27f8cad0c0ea58e/src/lib.rs#L1564C25-L1571C26

xnorpx: maybe it's cleaner to have a if else handshake pending in caller code rather than doing this.

algesten: Oh well.

PR #881, opened 2026-03-02
chore(ice): don't log idempotent updates

thomaseizinger:
When updating the timing config of an ICE agent to the exact values that are already being used, we can save ourselves logging a statement on DEBUG. This makes DEBUG logs just a tad less noisy.

### DIFF
--- a/src/ice/agent.rs
+++ b/src/ice/agent.rs
@@ -4,7 +4,6 @@ use std::net::SocketAddr;
 use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
-use std::{fmt, mem};
 
 use serde::{Deserialize, Serialize};
 


src/ice/agent.rs
```diff
@@ -388,7 +391,10 @@ impl IceAgent {
     ///
     /// Defaults to 3000ms.
     pub fn set_max_stun_rto(&mut self, timeout: Duration) {
-        self.timing_config.max_rto = timeout;
+        if mem::replace(&mut self.timing_config.max_rto, timeout) == timeout {
+            trace!(?timeout, "max_rto unchanged");
+            return;
+        }
```
algesten, line 397: Why is this using `mem::replace`? `Duration` is a `Copy` type, and what we're effectively doing is an early return (because we avoid the `bust_candidate_pair_timeout_caches` call).

Consider:

```rs
if self.timing_config_max_rto == timeout {
    trace!(?timeout, "max_rto unchanged");
    return;
}
… // as before
```

thomaseizinger: I used `mem::replace` to only reference the state variable once so that there is less risk of a bug where we check one field but overwrite another. Don't feel strongly though, just figured this was neat!

algesten: Let's simplify to an if-early return.

algesten (approved):

xnorpx:
@thomaseizinger fix lint and this can be merged

src/ice/agent.rs
```diff
@@ -1,9 +1,9 @@
 use std::collections::{HashSet, VecDeque};
-use std::fmt;
 use std::net::SocketAddr;
 use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
+use std::{fmt, mem};
```
thomaseizinger:
```suggestion
```

src/ice/agent.rs
```diff
@@ -1,9 +1,9 @@
 use std::collections::{HashSet, VecDeque};
```
thomaseizinger, line 1:
```suggestion
use std::collections::{HashSet, VecDeque};
use std::fmt;
```

PR #882, opened 2026-03-04
improvement: Use BCryptHash in SHA256 impl to avoid creating a hash instance.

efer-ms:
We don't need to create a hash object for each sha256 call, and can instead call BCryptHash with the psuedo-algorithm


xnorpx:
lgtm, merging this.

fyi @algesten

PR #885, opened 2026-03-04
[H265] Wire up sprop-max-don-diff SDP parameter to enable DONL end-to…

sanr90:
…-end.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * sprop-max-don-diff SDP parameter #885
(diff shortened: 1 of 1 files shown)


algesten:
One minor nit:

No validation of RFC 7798 value range

The field is u16 (0–65535), but RFC 7798 §7.1 defines the range as 0–32767. Values above 32767 will parse and be accepted silently. The test even notes this: "RFC 7798: 0..32767" but doesn't enforce it. Consider clamping or rejecting out-of-range values in FormatParam::parse.

algesten:
Thanks!

PR #886, opened 2026-03-04
Feature: android-crypto

efer-ms:
Implement an Android crypto features, which similar to apple-crypto utilizes the platform APIs to perform crypto.

This is an initial version, there are opportunities to better optimize the interaction with Java APIs.

I leaned heaviliy on Claude for this PR. We need to review all the calls into Java and ensure they are optimal for our use.

Note: We need to investigate how we can run the tests in CI on an android emulator. Running them using cargo-ndk works though if an emulator is available via adb.


efer-ms:
I'm gonna close this for now. I think we need to think a bit more over how to reduce the JNI overhead. Likely we'll need to provide a Class that initializes the provider. Maybe even calling that class sets it up as the default provider or something.

That class can then provide a more optimal API surface to reduce the JNI pain, as in, provide methods that take primitives and map to the java?.security APIs. Possibly have some pinned arrays for transferring data across the boundary, etc.

PR #888, opened 2026-03-06
sdp: Fix m-line ordering when merging offers

### LINKED ISSUE #887: SDP m-lines included out of order when using `merge()`
I've been trying to debug intermittent negotiation issues when joining a room with existing participants and media and trying to offer new media very quickly. When doing this, I'm hitting a "glare" issue where both my client and SFU send offers at the same time.

I already have code in place to handle catching this scenario and rolling back one of the offers, but in doing so I found that it was very easy to get the offers from str0m into a bad state where chrome would no longer accept them.

This appears to be caused by the use of the sdp [`merge()`](https://github.com/algesten/str0m/blob/85c4eba0718505c01e0a6fbfa6324255d06df85a/src/change/sdp.rs#L489) method. When this is used, the original offer is appended to the new offer. Also, since the MID line index is getting incremented for all SDP actions, it's possible to skip indexes when non-media lines (such as transport direction change) are included. When this happens, Chrome will reject the offer with an "m-lines out of order" error. This happens even if you roll back the original offer on the Chrome side, since it still seems to keep track of the m-line ordering in this case.

I found that by only indexing the media lines and by prepending instead of appending the original offer, I can prevent this issue from happening.

I have a PR and updated tests incoming shortly.

bekriebel:
When an offer is sent by a client and the SFU at the same time, we may have a conflict that requires one side to roll back their changes and renegotiate after accepting the incoming offer.

However, at least with Chrome, the client may still expect the m-lines to match their ordering even after a rollback.

To accomodate this, only increase the index on actions that produce an m-line to ensure an expected increment and, when merging, ensure that the original changes are prepended to the new offer instead of appending them and changing the m-line order.

Fixes #887

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   * Add `abs-capture-time` RTP header extension #864
   * Adjust `IceAgent::ice_timeout` to return timeout of successful pair #875
   * Fix infinite loop on in-band data channel stream conflict #873
+  * Fix m-line ordering when using sdp `merge()` #887
 
 # 0.16.2
 
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -488,7 +488,15 @@ impl<'a> SdpApi<'a> {
     /// ```
     pub fn merge(&mut self, mut pending_offer: SdpPendingOffer) {
         pending_offer.retain_relevant(self.rtc);
-        self.changes.extend(pending_offer.changes.drain(..));
+
+        // Prepend the original pending changes before the current SdpApi's own changes.
+        //
+        // AddMedia / AddApp entries from the pending offer carry already-allocated MIDs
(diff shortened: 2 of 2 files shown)


algesten:
@bekriebel looks good, but I think that WIP commit is not supposed to be there right?

bekriebel:
Oh, whoops. That was supposed to be a different test branch. I'll clean that up today. Sorry about that.

bekriebel:
@algesten Fixed, and rebased on the current main. I actually ended up resolving my issue in a different way in our codebase, so this is no longer blocking me. However, I do think it is still the correct way to handle these values.

I moved that temp commit into it's own PR, #893. This was something else that I found that was also a red herring for my issue, but figured I'd open it in case you want to match the RFC.

PR #819, opened 2026-01-05
AV1 support

Ramil980:
This PR introduces AV1 RTP packetizer and depacketizer support to str0m and enables AV1 in SDP by default

The packetizer is based on the libwebrtc implementation while the depacketizer is adapted from the pion (Go) implementation.

Unit tests for both the packetizer and depacketizer were adapted from libwebrtc, and AV1 support has been verified end-to-end using the chat example.

This is my first time writing code in Rust, so I’d really appreciate any feedback or suggestions 🙂

PR #889, opened 2026-03-07
Avoid vector overhead in av1 packetizer/depacketizer

Ramil980:
Hi, follow up for https://github.com/algesten/str0m/pull/819

Some optimizations by reusing Vec buffers instead of allocating new vectors each time

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Reduce allocations in AV1 packetizer/depacketizer #889
   * Workaround DCEP panics due to reuse race condition #883
   * sprop-max-don-diff SDP parameter #885
   * Reduce number of RTP header extensions sent on every packet #880


src/packet/av1.rs
```diff
@@ -583,45 +587,52 @@ pub fn leb_128_size(mut value: usize) -> usize {
...
         if let Some(obu_type) = obu.obu_type() {
             if obu_type.include_in_packetization() {
-                parsed_obus.push(obu);
+                obu_idx += 1;
             }
         }
     }
-
-    Ok(parsed_obus)
+    info!("[AV1 Depacketizer] Parsed {} OBUs from the packet", obu_idx);
```
algesten: This is probably a bit noisy. Remove?

Ramil980: Oops, added it for a quick testing, forgot to remove. Done, thanks!

PR #890, opened 2026-03-07
DTLS 1.3 more updates to apple crypto

xnorpx:
**Added X25519 key exchange** (kx_group.rs) — Implemented X25519 (Curve25519) key agreement using Apple CryptoKit alongside the existing P-256/P-384 ECDH groups. X25519 is first in the preference list as it's the most commonly negotiated key exchange in modern WebRTC. Includes a zero shared-secret check per RFC 7748 §6.1 to reject non-contributory low-order points.

**Added DTLS 1.3 cipher suites** (cipher_suite.rs) — Registered `TLS_AES_128_GCM_SHA256` and `TLS_AES_256_GCM_SHA384` as DTLS 1.3 cipher suites, including AES-ECB-based sequence number encryption. Added a doc comment explaining why ChaCha20-Poly1305 is not supported (Apple CryptoKit doesn't expose the raw ChaCha20 stream cipher needed for DTLS 1.3 record sequence number encryption).

**Updated provider docs** (lib.rs) — Added DTLS 1.3 and Auto to the provider documentation.

### Security Fixes

**Fixed certificate validity dates** (dtls.rs) — The certificate validity was hardcoded to `2024-01-01` through `2025-12-31`, meaning all generated certificates are expired as of 2026. Replaced with dynamic dates computed from `time::OffsetDateTime::now_utc()` with a 1-year validity window. Corrected the ASN.1 time encoding to use UTCTime (tag `0x17`, `YYMMDDHHMMSSZ`) for dates before 2050 per RFC 5280 §4.1.2.5, instead of the incorrectly-used GeneralizedTime.

**Removed bogus `SecAccessControl`** (dtls.rs) — Removed `SecAccessControl::create_with_flags(0)` which passed `0` as the protection level (not a valid `SecAttrAccessible` constant). Ephemeral DTLS keys don't need access control.

**Replaced `assert!` with error returns in crypto paths** (srtp.rs) — Three `assert!` calls in SRTP decrypt/encrypt paths (`input.len() >= TAG_LEN` and `aad.len() >= 12`) were panicking on malformed input from remote peers — a denial-of-service vector. Replaced with `Err(CryptoError::...)` returns.

**Fixed AES-CTR buffer overflow** (srtp.rs) — The `aes_ctr_round` function used fixed 2048-byte stack buffers but had no input length validation. Added an explicit bounds check returning an error if input exceeds 2048 bytes.

### Bug Fixes

**Fixed DER length encoding** (dtls.rs) — `encode_length` had a confusing fall-through structure where the short-form case was a no-op body relying on a shared `out.push(len as u8)` at the end. Restructured so each branch is self-contained and added a `debug_assert!` for lengths > 65535.

**Fixed CTR counter loop off-by-one** (srtp.rs) — Changed `while offset <= input.len()` to `while offset < input.len()` to avoid generating one extra unused IV block. Removed dead `_count` variable.

**Fixed `OutputBuffer` threshold** (cipher_suite.rs) — Changed `size < STACK_BUFFER_SIZE` to `size <= STACK_BUFFER_SIZE` so buffers of exactly 1024 bytes use the stack instead of heap-allocating.

**Fixed error message** (common_crypto.rs) — Changed "AES-256-ECB encryption failed" to include the actual key length since the function handles both AES-128 and AES-256.

**Replaced `.unwrap()` with `.expect()`** (srtp.rs) — The `srtp_aes_128_ecb_round` and `srtp_aes_256_ecb_round` wrappers now use descriptive `.expect()` messages instead of bare `.unwrap()`.

### Code Quality

**Unified test utilities** (mod.rs, hash.rs, hmac.rs) — Consolidated duplicate `hex_to_vec` and `slice_to_hex` implementations from hash.rs and hmac.rs tests into a shared `test_utils` module in `dimpl_provider/mod.rs`.

**Added HKDF test vectors** (hkdf.rs) — Added RFC 5869 test cases 1-3 for HKDF-Extract/Expand with SHA-256. Also replaced heap-allocated zero-salt with a stack array and moved the HKDF-Expand `input` vector allocation outside the loop to avoid repeated allocations. Used `u16::try_from` instead of a manual bounds check for the HKDF label output length.

### DIFF
--- a/crypto/apple-crypto/src/common_crypto.rs
+++ b/crypto/apple-crypto/src/common_crypto.rs
@@ -73,7 +73,8 @@ pub fn aes_ecb_round(key: &[u8], input: &[u8], output: &mut [u8]) -> Result<(), 
     match status {
         kCCSuccess => Ok(()),
         status => Err(CryptoError::Other(format!(
-            "AES-256-ECB encryption failed. Status: {status}"
+            "AES-ECB encryption failed (key_len={}). Status: {status}",
+            key.len()
         ))),
     }
 }
--- a/crypto/apple-crypto/src/dimpl_provider/cipher_suite.rs
+++ b/crypto/apple-crypto/src/dimpl_provider/cipher_suite.rs
@@ -202,6 +202,10 @@ static TLS13_AES_128_GCM_SHA256: Tls13Aes128GcmSha256 = Tls13Aes128GcmSha256;
 static TLS13_AES_256_GCM_SHA384: Tls13Aes256GcmSha384 = Tls13Aes256GcmSha384;
 
 /// All supported DTLS 1.3 cipher suites.
+///
+/// Note: ChaCha20-Poly1305 is not supported because Apple CryptoKit does not
+/// expose the raw ChaCha20 stream cipher needed for DTLS 1.3 record sequence
+/// number encryption (RFC 9001 Section 5.4.4).
 pub(super) static ALL_DTLS13_CIPHER_SUITES: &[&dyn SupportedDtls13CipherSuite] =
     &[&TLS13_AES_128_GCM_SHA256, &TLS13_AES_256_GCM_SHA384];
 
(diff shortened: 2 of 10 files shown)


xnorpx:
@efer-ms can you take a first pass on this on Monday?

PR #893, opened 2026-03-07
sdp: Always use ActPass when sending an offer

bekriebel:
When I was investigating a different issue, I came across this discrepancy. I don't have a real-world issue that this is affecting, but I wanted to open the PR in case you want to align with the RFC.

The text from the RFC:

> [5.5. ](https://datatracker.ietf.org/doc/html/rfc8842#section-5.5)[Modifying the Session](https://datatracker.ietf.org/doc/html/rfc8842#name-modifying-the-session)
> When an offerer sends a subsequent offer, if the offerer wants to establish a new DTLS association, the offerer MUST insert an SDP "setup" attribute [[RFC4145](https://datatracker.ietf.org/doc/html/rfc4145)] with an "actpass" attribute value, as well as or more SDP "fingerprint" attributes according to the procedures in [[RFC8122](https://datatracker.ietf.org/doc/html/rfc8122)]. In addition, the offerer MUST insert in the offer an SDP "tls-id" attribute with a new unique attribute value.
> 
> When an offerer sends a subsequent offer and does not want to establish a new DTLS association, if a previously established DTLS association exists, the offerer MUST insert in the offer an SDP "setup" attribute with an "actpass" attribute value, and one or more SDP "fingerprint" attributes with attribute values that do not change the previously sent fingerprint set. In addition, the offerer MUST insert an SDP "tls-id" attribute with the previously assigned attribute value in the offer.
> 
> NOTE: When a new DTLS association is being established, each endpoint needs to be prepared to receive data on both the new and old DTLS associations as long as both are alive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Always use ActPass when sending an SDP offer (RFC 8842 Section 5.2/5.5) #893
   * Fix frame drops in Firefox due missing RTX SSRC #940
   * Drop stale depacketizer state on stream pause and restore paused timestamp repair #929
 
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1551,15 +1551,26 @@ impl<'a, 'b> AsSdpParams<'a, 'b> {
             )
         };
 
+        // RFC 8842 Section 5.2/5.5: Offerers MUST always use a=setup:actpass.
+        // Answerers use the negotiated role (active/passive).
+        // We distinguish offer vs answer by the presence of `pending` (Some = offer).
+        let setup = if pending.is_some() {
+            // This is an offer — always actpass per RFC 8842
+            Setup::ActPass
+        } else {
+            // This is an answer — use the negotiated DTLS role
+            match rtc.dtls.is_active() {
+                Some(true) => Setup::Active,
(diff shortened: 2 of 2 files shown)


algesten:
Thanks for digging into this!     
                                                                                                                        
One thing worth noting for future-you (or future-me): even with this change we're not fully RFC 8842 compliant, because str0m doesn't emit the a=tls-id attribute. In 8842, tls-id (plus unchanged fingerprints) is what signals "reuse the existing DTLS association"

algesten:
Because you open from org `Route-8` I'm not allowed to add additional commits. Please rebase off latest main and add a changelog entry.

bekriebel:
> Because you open from org `Route-8` I'm not allowed to add additional commits. Please rebase off latest main and add a changelog entry.

Done!

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Always use ActPass when sending an SDP offer (RFC 8842 Section 5.2/5.5)
```
algesten, line 3: Add the PR number at the end #893 to follow precedence below.

algesten (changes requested):

algesten (approved):

algesten:
Thanks!

PR #894, opened 2026-03-08
DTLS 1.3 support with WinCrypto

xnorpx:


crypto/wincrypto/src/sys/dtls.rs
efer-ms, line 1: this can't be removed.. we need to keep using schannel for DTLS 1.2

efer-ms: We can either make:
- Use SChannel for 1.2 based on config (code has both dimpl for auto/1.3 and schannel for 1.2
- Make a feature flag for this crate that uses dimpl or schannel
- Create a different crate str0m-windows-crypto, and have it use dimpl exclusively, and we deprecate the current crate.

PR #898, opened 2026-03-08
Remove openssl condition

xnorpx:
Noticed when removing the Openssl dtls 1.2 impl that this doc test started failing.

So removed all openssl configs as we should work with defaults and changed the test to be predicable

This change should be done before Openssl is switching to dimpl


xnorpx:
@algesten this is test case fix for #899

Issue #895, opened 2026-03-08
Fuzz Fail


xnorpx:
@Ramil980 can you fix this one

thread '<unnamed>' (5842) panicked at /home/runner/work/str0m/str0m/src/packet/mod.rs:220:31:
attempt to multiply with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
==5842== ERROR: libFuzzer: deadly signal

```
8b3853b9c src/packet/mod.rs       (Ramil980          2026-03-07 19:02:27 +0400 219)     fn get_bytes(&mut self, size: usize) -> Option<&[u8]> {
47b196513 src/packet/mod.rs       (Ramil980          2026-01-11 14:09:23 +0400 220)         if self.remaining() < (size * 8) {
47b196513 src/packet/mod.rs       (Ramil980          2026-01-11 14:09:23 +0400 221)             return None;
```

https://github.com/algesten/str0m/actions/runs/22810386032/job/66166151285

Ramil980:
Will fix today

xnorpx:
This is merged in #900

PR #900, opened 2026-03-09
Fix the fuzz crash due to multiplication overflow

Ramil980:
Fix for the https://github.com/algesten/str0m/issues/895

Apparently Claude decided the overflow check was optional in the latest [PR](https://github.com/algesten/str0m/pull/889/changes#diff-a148bdc2d256ce7e39e50b7177cf124fa3cbd8021559b05dc2765347e85967f1L606)… and I missed it

@xnorpx, please review

### DIFF
--- a/src/packet/mod.rs
+++ b/src/packet/mod.rs
@@ -156,36 +156,36 @@ pub use error::PacketError;
 
 /// Helper to replace Bytes. Provides get_u8 and get_u16 over some buffer of bytes.
 pub(crate) trait BitRead {
-    fn remaining(&self) -> usize;
+    fn remaining_bits(&self) -> usize;
     fn remaining_bytes(&self) -> usize;
-    fn get_offset(&self) -> usize;
+    fn byte_offset(&self) -> usize;
     fn get_u8(&mut self) -> Option<u8>;
     fn get_u16(&mut self) -> Option<u16>;
     fn get_remaining(&mut self) -> &[u8];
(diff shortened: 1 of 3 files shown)


src/packet/mod.rs
```diff
@@ -217,11 +217,12 @@ impl BitRead for (&[u8], usize) {
     }
 
     fn get_bytes(&mut self, size: usize) -> Option<&[u8]> {
```
xnorpx: maybe size_bytes here

src/packet/mod.rs
```diff
@@ -244,11 +245,12 @@ impl BitRead for (&[u8], usize) {
     }
 
     fn consume(&mut self, bytes: usize) {
```
xnorpx: use same names here as in get?

Ramil980: Consume doesn’t return anything, so get might not be the best fit. Some of the other method names also feel a bit off. let me see what alternatives I can come up with.

Ramil980: Renamed some of them and applied the get_ prefix only to the ones that ‘read the byte(s), advance the cursor, and return the result

xnorpx: sorry ran out of time today, will take a look tomorrow

algesten:
@Ramil980 is this ready?

Ramil980:
@algesten, functionality‑wise it’s ready. @xnorpx left a few minor nit comments, and he’ll review the final version today

PR #901, opened 2026-03-10
No panic, just error and return None

xnorpx:
```
thread 'REDACTED' (19) panicked at /opt/cargo/registry/src/REDACTED/str0m-0.16.2/src/session.rs:793:14:
index is media
stack backtrace:
   0: __rustc::rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::option::expect_failed
   3: <str0m::session::Session>::poll_packet
   4: <core::option::Option<str0m::io::DatagramSend>>::or_else::<<str0m::Rtc>::do_poll_output::{closure#2}>
   5: <str0m::Rtc>::do_poll_output
   6: <str0m::Rtc>::poll_output
   7: <REDACTED>::poll_event
   8: <REDACTED>::poll_and_handle_rtp_connection_events
   9: <REDACTED>::handle_ingress_packet
  10: <REDACTED>::handle_rtc_packet_from_endpoint
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Which is most likely an artifact of pacer not empty and we remove the mid before poll output.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -790,7 +790,7 @@ impl Session {
         // The cluster_id is captured by the pacer at poll time, before register_send() might clear it
         let (midrid, cluster_id) = self.pacer.poll_queue()?;
         let Some(media) = self.medias.iter().find(|m| m.mid() == midrid.mid()) else {
-            error!("Pacer pointed to mid {} which has no media", midrid.mid());
+            trace!("Pacer pointed to mid {} which has no media", midrid.mid());
             return None;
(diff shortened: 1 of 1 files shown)


algesten:
@xnorpx your analysis of this problem seems plausible, which means the condition is not an error-level event, but expected, so either trace or debug. I made trace for now.

algesten (approved):

PR #903, opened 2026-03-10
Dev/wincryptodtl13 feat

xnorpx:
### Problem
`str0m-wincrypto` was locked to DTLS 1.2 only, because it relied on Windows SChannel which doesn't expose DTLS 1.3. This limited interop testing and blocked DTLS 1.3 usage on Windows with Wincrypto provider.

### Solution: Two DTLS paths via feature flags

**Root Cargo.toml:**
- `wincrypto` (existing) — now implies `str0m-wincrypto/dtls12` → SChannel DTLS 1.2
- `wincrypto_dimpl` (new) — enables `str0m-wincrypto` *without* `dtls12` → dimpl DTLS 1.2/1.3/Auto

**`str0m-wincrypto/Cargo.toml`:**
- New `dtls12` feature (default) gates SChannel dependencies (`Win32_Security_Authentication_Identity`, `Win32_Security_Credentials`, `Win32_System_Rpc`, `tracing`)
- New dependency: `dimpl = "0.4.2"` (always present, used for DTLS when `dtls12` is off)

### New modules (~2,800 lines)

| Module | Lines | Purpose |
|---|---|---|
| `dimpl_provider/sign.rs` | 802 | ECDSA P-256/P-384 signing & verification via CNG |
| `dimpl_provider/kx_group.rs` | 679 | ECDH (P-256/P-384) and X25519 key exchange via CNG |
| `dimpl_dtls.rs` | 446 | dimpl DTLS adapter: certificate generation (self-signed X.509 DER), DTLS instance wiring, keying material extraction |
| `dimpl_provider/cipher_suite.rs` | 432 | AES-128-GCM and AES-256-GCM cipher suites via CNG |
| `schannel_dtls.rs` | 240 | Extracted SChannel DTLS code (was in `dtls.rs`) |
| `dimpl_provider/hkdf.rs` | 146 | HKDF-Extract/Expand via CNG |
| `dimpl_provider/hash.rs` | 114 | SHA-256/SHA-384 incremental hashing via CNG |
| `dimpl_provider/hmac.rs` | 74 | HMAC-SHA-256/SHA-384 via CNG |
| `dimpl_provider/tls12.rs` | 69 | TLS 1.2 PRF (P_SHA-256) via CNG |
| `dimpl_provider/mod.rs` | 59 | Provider assembly + `BCryptGenRandom` secure random |

### Refactored modules

- **`dtls.rs`** — Slimmed from 249 → 24 lines. The `WinCryptoDtlsProvider` now delegates to either `schannel_dtls` or `dimpl_dtls` via `cfg(feature = "dtls12")`.
- **`sys/provider.rs`** — Deleted (407 lines). The monolithic provider was broken up: `default_provider()` moved to lib.rs, individual provider structs moved to their own modules (`srtp.rs`, `sha1.rs`, `sha256.rs`, `dtls.rs`).
- **`sys/srtp.rs`** — Added bounds check to prevent buffer overflow when input exceeds `MAX_BUFFER_SIZE`.

### Test infrastructure

- **common.rs** — `wincrypto_dimpl` recognized as a crypto provider name, routed to the same `str0m_wincrypto::default_provider()`.
- **handshake-direct.rs** — DTLS 1.3/Auto tests are correctly skipped only when using the SChannel path (`wincrypto`), not the dimpl path (`wincrypto_dimpl`).
- **nightly.yml** — New `windows-crypto-dimpl-test` job: 3x3 matrix (`rust-crypto`, `aws`, `wincrypto_dimpl`) testing cross-provider DTLS interop on Windows.

### Key design decisions

1. **Feature-gated, not runtime-switched** — compile-time `cfg` keeps the binary lean; no SChannel deps when using dimpl and vice versa.
2. **All crypto stays Windows-native** — the dimpl path uses CNG (`BCrypt*` APIs) for every primitive, not ring/aws-lc-rs. Only the DTLS state machine comes from dimpl.
3. **Backward compatible** — `dtls12` is the default feature, so existing `--features wincrypto` users get the same SChannel behavior as before.


xnorpx:
@efer-ms  to make a first pass

.github/workflows/nightly.yml
```diff
@@ -42,6 +42,35 @@ jobs:
           R_CRYPTO: ${{ matrix.r_crypto }}
         run: cargo +${{steps.toolchain.outputs.name}} test --release --no-default-features --features rust-crypto,aws-lc-rs,wincrypto --lib --bins --tests
 
+  windows-crypto-dimpl-test:
```
xnorpx, line 45: why can't this be part of the regular matrix?

Cargo.toml
```diff
@@ -72,7 +73,7 @@ str0m-apple-crypto = { version = "0.1.4", path = "crypto/apple-crypto", optional
 
 [target.'cfg(windows)'.dependencies]
 # Windows Crypto (CNG + SChannel)
```
efer-ms, line 75: probably should update the SChannel comment? like it's CNG.. but it's only SChannel with the feature.

Cargo.toml
```diff
@@ -28,7 +28,8 @@ default = ["aws-lc-rs", "examples"]
 aws-lc-rs = ["dep:str0m-aws-lc-rs"]
 rust-crypto = ["dep:str0m-rust-crypto"]
 openssl = ["dep:str0m-openssl"]
-wincrypto = ["dep:str0m-wincrypto"]
+wincrypto = ["dep:str0m-wincrypto", "str0m-wincrypto/dtls12"]
+wincrypto_dimpl = ["dep:str0m-wincrypto"]
```
efer-ms, line 32: can we invert this? can we make dimpl the default, and dtls12 the special-case. Like eventually we want dimpl to be the only one... also dimpl is an implementation detail.

I'm willing to bet we can count the users of wincrypto on one finger, I'm pretty sure they can just update the dep.

efer-ms: like I'd rather have "wincrypto" and "wincrypto_dtls12" or "wincrypto_schannel".

src/crypto/srtp.rs
```diff
@@ -65,12 +65,12 @@ impl SrtpCrypto {
...
     }
 
-    #[cfg(all(feature = "wincrypto", target_os = "windows"))]
+    #[cfg(all(any(feature = "wincrypto", feature = "wincrypto_dimpl"), target_os = "windows"))]
     pub fn new_wincrypto() -> SrtpCrypto {
         Self::WinCrypto(super::wincrypto::WinCryptoSrtpCryptoImpl)
     }
 
-    #[cfg(not(all(feature = "wincrypto", target_os = "windows")))]
+    #[cfg(not(all(any(feature = "wincrypto", feature = "wincrypto_dimpl"), target_os = "windows")))]
```
efer-ms, line 73: of course this kinda of thing makes me wonder why we don't have a feature for dtls1_3?

like if dtls1_3 is enabled... then it enables dimpl in wincrypto

crypto/wincrypto/Cargo.toml
```diff
@@ -10,12 +10,27 @@ edition = "2024"
...
+  "windows/Win32_Security_Authentication_Identity",
+  "windows/Win32_Security_Credentials",
+  "windows/Win32_System_Rpc",
+]
+
 [dependencies]
-tracing = "0.1.37"
-str0m-proto = { version = "0.1.1", path = "../../proto" }
+str0m-proto = { version = "0.1.2", path = "../../proto" }
+dimpl = { version = "0.4.2", default-features = false }
```
efer-ms, line 26: shouldn't this be optional too? or maybe in a depedencies section tied to not being dtls12?

crypto/wincrypto/src/schannel_dtls.rs
```diff
@@ -0,0 +1,240 @@
...
+        certificate: der_bytes,
+        private_key: vec![], // Private key is managed internally by Windows
+    })
+}
+
+pub(crate) fn new_dtls(
+    cert: &DtlsCert,
+    dtls_version: DtlsVersion,
+) -> Result<Box<dyn DtlsInstance>, CryptoError> {
+    if dtls_version != DtlsVersion::Dtls12 {
```
efer-ms, line 42: why woulnd't Auto work? it auto selects 1.2.

efer-ms: like it just feels unneccessarily fragile.

most code should use Auto.. cause they shouldn't care which DTLS is used... only the toml files which control which feature is enabled need to care that only 1.2 is supported.

crypto/wincrypto/src/schannel_dtls.rs
```diff
@@ -0,0 +1,240 @@
...
+        queued_app_data: VecDeque::new(),
+        last_timeout: None,
+    }))
+}
+
+// ============================================================================
+// DTLS Instance
+// ============================================================================
+
+struct SchannelDtlsInstance {
```
efer-ms, line 77: drop the `Instance`.. it's cleaner.

crypto/wincrypto/src/dtls.rs
```diff
@@ -1,244 +1,43 @@
...
-        let cert = Certificate::new_self_signed(true, "CN=WebRTC").ok()?;
-        let cert = Arc::new(cert);
-
-        // Get the DER bytes from the certificate
-        let der_bytes = cert.get_der_bytes().ok()?;
-
-        // Cache the certificate for later use in new_dtls, keyed by DER bytes
-        if let Ok(mut cache) = CERT_CACHE.lock() {
-            cache.insert(der_bytes.clone(), Arc::clone(&cert));
+        #[cfg(feature = "dtls12")]
```
efer-ms, line 17: instead of this file existing.. just include schannel_dtls, or dimpl_dtls. you don't need this file at all.

you just have two different WinCryptoDtlsProviders .. and you include the one based on feature. You can make the two mods have different versions if you want and just reexport the one that's included under the WinCryptoDtlsProvider name

crypto/wincrypto/src/dimpl_dtls.rs
```diff
@@ -0,0 +1,446 @@
...
+    let mut tbs = Vec::new();
+
+    // Version: v3
+    let version = encode_explicit_tag(0, &encode_integer(&[2]));
+    tbs.extend_from_slice(&version);
+
+    // Serial number (random)
+    let mut serial = [0u8; 16];
+    use windows::Win32::Security::Cryptography::BCRYPT_USE_SYSTEM_PREFERRED_RNG;
+    use windows::Win32::Security::Cryptography::BCryptGenRandom;
```
efer-ms, line 99: why are these here instead of at the top with other uses?

crypto/wincrypto/src/dimpl_dtls.rs
```diff
@@ -0,0 +1,446 @@
...
+    cert_content.extend_from_slice(&signature_bits);
+
+    Ok(encode_sequence(&cert_content))
+}
+
+fn build_tbs_certificate(
+    public_key_bytes: &[u8],
+    now: OffsetDateTime,
+) -> Result<Vec<u8>, CryptoError> {
+    let mut tbs = Vec::new();
```
efer-ms, line 90: tbs is kinda a weird abbreviation. I know it stands for ToBeSigned ... but like here we're building the cert structure. maybe the variable tbs just needs to be like cert_buffer? or something like that?

crypto/wincrypto/src/dimpl_dtls.rs
```diff
@@ -0,0 +1,446 @@
...
+
+    let mut cert_content = Vec::new();
+    cert_content.extend_from_slice(&tbs);
+    cert_content.extend_from_slice(&sig_algorithm);
+    cert_content.extend_from_slice(&signature_bits);
+
+    Ok(encode_sequence(&cert_content))
+}
+
+fn build_tbs_certificate(
```
efer-ms, line 86: this whole function seems generic. The only wincrypto piece is using the RNG for serial number. Do none of the other dimpl providers do the same thing?

Like can we add this stuff to dimpl?

crypto/wincrypto/src/dimpl_dtls.rs
```diff
@@ -0,0 +1,446 @@
...
+    dtls: Dtls,
+}
+
+impl std::fmt::Debug for DimplDtlsInstance {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("DimplDtlsInstance").finish()
+    }
+}
+
+impl DtlsInstance for DimplDtlsInstance {
```
efer-ms, line 383: I also guess this is likely generic too. Shouldn't str0m have a DimplDtls struct that any crypto impl can use if thye use dimpl

crypto/wincrypto/src/dimpl_provider/tls12.rs
```diff
@@ -0,0 +1,69 @@
...
+            .map(|h| h.to_vec())
+            .map_err(|e| format!("{e}")),
+        HashAlgorithm::SHA384 => super::hmac::win_hmac_sha384(key, data)
+            .map(|h| h.to_vec())
+            .map_err(|e| format!("{e}")),
+        _ => Err(format!("Unsupported hash: {hash:?}")),
+    }
+}
+
+pub(super) static PRF_PROVIDER: WinCngPrfProvider = WinCngPrfProvider;
```
efer-ms, line 69: copy test vector tests from apple-crypto ...

again.. we could just have a dimpl mod under str0m so this stuff is more easily shared.

crypto/wincrypto/src/dimpl_provider/hmac.rs
```diff
@@ -0,0 +1,74 @@
...
+            None,
+            Some(key),
+            0,
+        ))?;
+
+        WinCryptoError::from_ntstatus(BCryptHashData(*hash_handle, data, 0))?;
+        WinCryptoError::from_ntstatus(BCryptFinishHash(*hash_handle, &mut hash, 0))?;
+    }
+    Ok(hash)
+}
```
efer-ms, line 74: copy test vectors from apple-crypto

crypto/wincrypto/src/dimpl_provider/hash.rs
```diff
@@ -0,0 +1,114 @@
...
+    fn update(&mut self, data: &[u8]) {
+        self.data.extend_from_slice(data);
+    }
+
+    fn clone_and_finalize(&self, out: &mut Buf) {
+        let digest = self.finalize_snapshot();
+        out.clear();
+        out.extend_from_slice(&digest);
+    }
+}
```
efer-ms, line 114: copy test vectors from apple-crypto

crypto/wincrypto/src/dimpl_provider/hash.rs
```diff
@@ -0,0 +1,114 @@
...
+#[derive(Clone, Copy)]
+enum HashKind {
+    Sha256,
+    Sha384,
+}
+
+/// Incremental hash context using Windows CNG.
+///
+/// CNG hash handles are NOT clonable, so to implement `clone_and_finalize`
+/// we buffer all input and re-hash from scratch each time.
```
efer-ms, line 40: No. Please use...

https://learn.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptduplicatehash

efer-ms: so basically.. create the hash once store it in the context... clone_and_finalize.. can call duplicatehash and finish the dupliocate

crypto/wincrypto/src/dimpl_provider/hkdf.rs
```diff
@@ -0,0 +1,146 @@
...
+        HashAlgorithm::SHA384 => {
+            let result =
+                super::hmac::win_hmac_sha384(key, data).map_err(|e| format!("HMAC-SHA384: {e}"))?;
+            Ok(result.to_vec())
+        }
+        _ => Err(format!("Unsupported hash for HKDF: {hash:?}")),
+    }
+}
+
+impl HkdfProvider for WinCngHkdfProvider {
```
efer-ms, line 25: this is identical to the one in apple-crypto. can we have default implementations for the trait maybe? like? the only thing that changes is the hmac function? actually we have an hmac provider.. can't we just use that? why do we repeat the hmac code in this file?

PR #904, opened 2026-03-12
Refactor: Move generic code out of apple-crypto so other dimpl based crates can share the common code

efer-ms:
When we were moving wincrypto to utilize dimpl, we saw lots of generic code that could be shared with Apple-crypto. I anticipate it will be the same with other dimpl cryptos.

### DIFF
--- a/proto/src/crypto/dimpl.rs
+++ b/proto/src/crypto/dimpl.rs
@@ -201,9 +201,11 @@ macro_rules! impl_hmac_providers {
         #[cfg(test)]
         mod dimpl_common_hmac_tests {
             use super::*;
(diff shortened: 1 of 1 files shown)


algesten:
@efer-ms I realize now that we can make the prf and hkdf stuff impls once for all crypto providers in dimpl, and only have the `CryptoProvider` provide the hmac functions.

I prepared a PR for that here: https://github.com/algesten/dimpl/pull/94 - this also copies across the tests you made for this in the macro.

The other part, making certificate generation without rcgen, is something that was also discussed on Discord. I am for that, and think we should make dimpl create certs without that dep, just relying on what we already have. I think this is a another PR in dimpl.

PR #907, opened 2026-03-19
Fix jitter calculation in ReceiverReport

alexlapa:
I've been implementing mean opinion score for SFU connections and it seems that there is something off with jitter reported in str0m's ReceiverReport cause I'm runing everything on localhost and observing some huge values of `RTCRemoteInboundRtpStreamStats.jitter` reported by Chrome senders (which is based on ReceiverReport sent by str0m).

According to [this test](https://github.com/algesten/str0m/blob/fc14f4597d358f8862326b1eb1927399b9133f7b/src/streams/register.rs#L278-L307) it seems that jitter written to ReceiverReport is not a number of RTP timestamp "ticks" as its suppposed to be but just a raw number of microseconds. So I've added [some tests from libwebrtc](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/rtp_rtcp/source/receive_statistics_unittest.cc;drc=81f4e7805bdc25255a364a8ff7a3ac2fad1dcd5e) and yeah the numbers were "a bit" off: `expected: 29, actual: 3632`, `expected: 300, actual: 6250`.

I've also ran few "golden" tests comparing libwebrtc to str0m (after this fix) and numbers are still off but not that much so its probably ok. Alternatively I can fully allign str0m implementation with libwebrtc, there are few things that work in a different way. But, honestly, it should not make a huge difference.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
 # Unreleased
 
+  * Fix jitter calculation in ReceiverReports #907
+
+# 0.17.0
+
   * Update to dimpl 0.4.3 #902
   * Update ~dimpl to 0.4.2~ (see 902) and sctp-proto to 0.8.1 #892
+  * Update ~dimpl to 0.3.0~ (see 902) #874
   * MSRV 1.85, Edition 2024 and bump deps #891
   * Reduce allocations in AV1 packetizer/depacketizer #889
   * Workaround DCEP panics due to reuse race condition #883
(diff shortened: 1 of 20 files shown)


alexlapa:
Ah, so I've initially mixed up SenderReports and ReceiverReports in changelog and PR description. Updated.

src/streams/register.rs
algesten: Is this comment wrong?

Should it be "Estimated jitter in microseconds"?

algesten, line 255: Claude informs me we have one too many zeroes here.

algesten:
Looks good, thanks! Two small things.

src/streams/register.rs
alexlapa: Fixed

alexlapa: Yea I wanted to fix this in a separate PR cause it goes a bit deeper, not only there is too much zeroes but it also can be negative, so `ReceptionReport.packets_lost` should alos be changed to `i32`. And I'l also probably add some unit tests from libwebrtc.

> cumulative number of packets lost: 24 bits
      The total number of RTP data packets from source SSRC_n that have
      been lost since the beginning of reception.  This number is
      defined to be the number of packets expected less the number of
      packets actually received, where the number of packets received
      includes any which are late or duplicates.  Thus, packets that
      arrive late are not counted as lost, and the loss **may be negative**
      if there are duplicates.  The number of packets expected is
      defined to be the extended last sequence number received, as
      defined next, less the initial sequence number received.  This may
      be calculated as shown in [Appendix A.3](https://www.rfc-editor.org/rfc/rfc3550#appendix-A.3).

alexlapa:
@algesten , updated comment, will fix loss calculation in separate PR.

PR #908, opened 2026-03-19
fix(ice): preserve nomination on redundant pair replacement

zRedShift:
When a redundant pair replacement happens while nomination is in `Attempt`, the replacement pair loses nomination because `copy_nominated_and_success_state` skips `Attempt`.

`Attempt` cannot be copied directly because the in-flight transaction belongs to the old pair. Falling back to `None` loses the nomination state entirely. Resetting to `Nominated` is the smallest safe fallback: the next binding cycle creates a fresh nominated binding attempt, which can then promote the replacement pair back to `Success`.

For a controlled agent, losing nomination during pair replacement can move an otherwise connected session back to `Checking` while candidate pairs are still considered possible. Since the checks themselves may continue succeeding, the agent can remain in `Checking` until retransmits expire instead of recovering nomination promptly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   * Fix infinite loop on in-band data channel stream conflict #873
   * Add `Hash`, `PartialOrd` and `Ord` traits to `TransId` #905
   * Fix m-line ordering when using sdp `merge()` #887
+  * Preserve nomination on redundant pair replacement #908
 
 # 0.16.2
 


src/ice/agent.rs
```diff
@@ -2407,4 +2407,76 @@ mod test {
 
         buf
     }
+
+    /// Regression test: when a pair with nomination in `Attempt` state is replaced by a
+    /// redundant candidate, the replacement pair must preserve the nomination so the next
+    /// binding cycle can re-establish it. Before this fix, `Attempt` was dropped, causing
+    /// nomination loss → Connected→Checking linger → connection slot exhaustion.
+    #[test]
+    fn pair_replacement_preserves_nomination_in_attempt_state() {
```
thomaseizinger: Can we write this test such that only two agents interact with each other? I'd like to see in what circumstances this problem can get triggered just be sending back and forth STUN bindings and the timings of adding new candidates.

zRedShift: Added a two-agent regression that drives the real STUN exchange and injects the signalled candidate while the controlled side is in `Attempt`; it asserts we stay connected and don’t fall back to `Checking`. Kept the lower-level pair-state test alongside it.

thomaseizinger: Thank you! It is verbose but I guess that is expected when we need to create such a specific state. Thank you for adding that.

From my perspective, we don't need the low-level test then but I'll let @algesten be the judge of that.

thomaseizinger: Just confirming: The high-level test fails without your patch?

algesten: @thomaseizinger I defer to you. You're much more in-tune with this code than me at this point :)

zRedShift: > Just confirming: The high-level test fails without your patch?

Yes, both tests fail without the patch.
I don't mind removing the low-level test and the cfg(test) function it requires - just LMK.

thomaseizinger: Yeah cool. Lets remove the low-level test then please!

zRedShift: Done.

src/ice/mod.rs
```diff
@@ -702,6 +702,158 @@ mod test {
...
+        a1.handle_timeout(now);
+        drain_manual_test_events(&mut a2, &mut a1);
+        advance_manual_test_time(&mut a1);
+        advance_manual_test_time(&mut a2);
+
+        let now = a2.time;
+        a2.handle_timeout(now);
+        drain_manual_test_events(&mut a1, &mut a2);
+        advance_manual_test_time(&mut a1);
+        advance_manual_test_time(&mut a2);
```
thomaseizinger: This pattern is repeated a lot throughout the test. Does it make sense to introduce something like `progress_without_network`?

src/ice/mod.rs
```diff
@@ -702,6 +702,158 @@ mod test {
...
+
+        let initial_request = a1
+            .poll_transmit()
+            .expect("controlling agent should emit an initial binding request");
+        {
+            let message = StunMessage::parse(&initial_request.contents).unwrap();
+            assert!(message.is_binding_request());
+            assert!(!message.use_candidate());
+        }
```
thomaseizinger: This could be made a bit more concise if we add a `poll_next_stun_message` function to `TestAgent` that parses the next transmit as a `StunMessage`.

src/ice/mod.rs
```diff
@@ -702,6 +702,158 @@ mod test {
...
+
+            assert!(message.is_binding_request());
+            assert!(
+                !message.use_candidate(),
+                "controlled agent must not send USE-CANDIDATE",
+            );
+            break trans;
+        };
+        let nominate_reply =
+            nominate_reply.expect("controlled agent should reply to the nominated request");
```
thomaseizinger: Is there a way we can express a bit more clearly, what we are trying to do here? It seems like we want to read two messages. Why are we doing that in a loop? Shouldn't it always be deterministic, which one is emitted first?

src/ice/mod.rs
thomaseizinger: Or, we change `deliver_manual_test_transmit` to include the `poll_transmit` inside but also return the message that got delivered so we can make assertions on it. I think that would make the test more readable.

src/ice/mod.rs
zRedShift: It does, extracted.

src/ice/mod.rs
zRedShift: Agreed. I removed the loop and made the ordering explicit now.

src/ice/mod.rs
zRedShift: Added `poll_next_stun_message` on `TestAgent`. I kept `deliver_manual_test_transmit` as the two-agent delivery helper, so polling/assertion stays separate from delivery.

thomaseizinger:
I've pushed another refactoring. Thank you for the fix! :)

zRedShift:
@thomaseizinger @algesten Thanks! One heads-up: the changelog currently lists this under str0m 0.17.0, but 0.17.0 was already released to crates.io on March 20, 2026, so this should probably move to `Unreleased` or `0.17.1` if you want it in a patch release.

For us this fix is patch-release worthy. We’ll pin to the commit for now, since peer-reflexive candidates arriving before or during out-of-band trickle are common in our product, and a limited rollout of this branch on top of 0.17.0 already produced a noticeable increase in session success rates.

I also noticed the firezone/firezone#12646 linkage and the crates.io version bump, so I wanted to call this out.

thomaseizinger:
Oh good catch, thank you!

PR #912, opened 2026-03-21
Refactor out crate: is

algesten:


is/Cargo.lock
thomaseizinger, line 1: Huh? Why is there a separate lockfile? Is this repo not a workspace?

is/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "is"
+version = "0.1.0"
+authors = ["Martin Algesten <martin@algesten.se>", "Thomas Eizinger <thomas@eizinger.io>"]
```
thomaseizinger, line 4: Thank you for the credit🙏

is/src/lib.rs
```diff
@@ -1,6 +1,102 @@
+//! # is -- ICE in Sans-IO style
```
thomaseizinger, line 1: Do you know about this?

```
#![doc = include_str!("../README.md")]
```

Would save you from duplicating the information from the README.

proto/src/stun/mod.rs
thomaseizinger, line 110: Perhaps worth depending on [`stun-types`](https://docs.rs/stun-types)?

thomaseizinger: Why are these moving to `str0m-proto`? Can't they just be defined (and exported) in `is`?

thomaseizinger (commented):
Overall LGTM

A few questions!

is/src/lib.rs
algesten: I've come to use `cargo readme` for this.

is/Cargo.lock
algesten: Good question

is/Cargo.toml
algesten: 🎉

proto/src/stun/mod.rs
algesten: Yeah, weird. Let's fix.

algesten:
@copilot make it run the CI tests

PR #913, opened 2026-03-23
fix: update deny.toml and lock files for cargo-deny & Dependabot

xnorpx:
- Move deprecated top-level 'targets' and 'exclude' keys to [graph]
- Add unused-allowed-license = "allow" to silence warnings for pre-approved licenses not yet used in the dependency tree
- Update aws-lc-sys 0.38 → 0.39 in sub-crate lock files to fix RUSTSEC AWS-LC X.509 Name Constraints Bypass (High) and CRL Distribution Point Scope Logic Error (High)
- Run cargo update on all sub-crate lock files to pick up latest compatible dependency versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -278,9 +278,9 @@ checksum = "2e93abca9e28e0a1b9877922aacb20576e05d4679ffa78c3d6dc22a26a216659"
 
 [[package]]
 name = "cc"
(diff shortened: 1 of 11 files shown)


xnorpx:
Some busywork to get dependabot happy  and silence some warnings in cargo deny

PR #915, opened 2026-03-23
Fix VLA serialization issue

Ramil980:
Hi, VLA serialization is currently broken when the number of simulcast and spatial layer combinations goes beyond 3. This PR fixes it


algesten:
LGtM

PR #916, opened 2026-03-23
chore: Update to latest dimpl API

efer-ms:
Update apple-crypto:
- Use the updated dimpl API
- Add test that simply runs the validate() function on the provider so we can verify dimpl related crypto

### DIFF
--- a/crypto/apple-crypto/src/dimpl_provider/mod.rs
+++ b/crypto/apple-crypto/src/dimpl_provider/mod.rs
@@ -40,13 +40,3 @@ impl SecureRandom for AppleSecureRandom {
 }
 
 static SECURE_RANDOM: AppleSecureRandom = AppleSecureRandom;
-
-#[cfg(test)]
-mod tests {
-    #[test]
-    fn validate_dimpl_provider() -> Result<(), String> {
-        super::default_provider()
-            .validate()
-            .map_err(|err| format!("{err:?}"))
-    }
-}


crypto/apple-crypto/src/dimpl_provider/mod.rs
```diff
@@ -44,3 +40,13 @@ impl SecureRandom for AppleSecureRandom {
...
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn validate_dimpl_provider() -> Result<(), String> {
+        super::default_provider()
+            .validate()
+            .map_err(|err| format!("{err:?}"))
+    }
+}
```
algesten: We are always validating in dimpl: https://github.com/algesten/dimpl/blob/main/src/config.rs#L392

efer-ms: Unclear on the suggestion. Are you suggesting we shouldn't bother with this test.. cause anything that calls build() on ConfigBuilder would fail anyways?

Isn't it preferable to have a narrower test, so if it fails you know it's in validate? I mean I'd actually love to have individual tests for the providers and test vectors.. but I don't like having to repeat all those test vectors everywhere.

algesten: The test vectors are run on every build(), i.e. each start. So if there's a single test with the crypto backend, it will be tested implicitly.

I don't care hugely, you can add these tests across the board if you think that's good.

efer-ms: I'll remove it for now, and go back to thinking aobut how best to share test vectors. I had added test vectors previously cause the PrfProvider had bugs.. having clear information about which vectors and ciphers failed is helpful vs having any one failure just tell you one failed. But that is a separate issue, which validate() doesn't solve for us anyways.

PR #918, opened 2026-03-24
Move subcrates to crates/ and set up Cargo workspace

algesten:
Move proto, is, netem, _str0m_test, and fuzz into crates/ subfolder. Configure str0m as both the top-level crate and the workspace root, with all dependency versions centralized in [workspace.dependencies]. Remove non-root Cargo.lock files. Update CI workflows and fuzz scripts for the new paths.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,20 +79,11 @@ p256 = { version = "0.13", features = ["ecdh"] }
 sha2 = "0.10"
 openssl-sys = "0.9.80"
 libc = "0.2"
-apple-cryptokit-rs = "0.2.1"
-core-foundation = "0.10"
-security-framework = "3.5"
 der = "0.7"
 pkcs8 = { version = "0.10", features = ["pem"] }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -23,6 +23,88 @@ rust-version = "1.85.0"
...
+libc = "0.2"
+apple-cryptokit-rs = "0.2.1"
+core-foundation = "0.10"
+security-framework = "3.5"
+der = "0.7"
+pkcs8 = { version = "0.10", features = ["pem"] }
+sec1 = "0.7"
+spki = "0.7"
+x509-cert = { version = "0.2", default-features = false }
+windows = { version = "0.62", features = [
```
xnorpx: nit: I would put platform specific crates in the crypto crates? Like Windows only in wincrypto etc.

algesten: sounds good

xnorpx (approved):
lgtm

PR #919, opened 2026-03-24
Feature: Dimpl-backed OpenSSL Crypto

efer-ms:
- Adds a new feature to Str0m: openssl-dimpl - which enables the str0m-openssl dep and enables the 'dimpl' feature on that crate.
- Added a 'dimpl' feature on the str0m-openssl crate
  - When enabled, this replaces the OpenSSL based DTLS implementation with dimpl, allowing DTLS 1.3.

Note: If you enable both openssl-dimpl AND openssl at the same time, then only openssl-dimpl is available.

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -21,6 +21,10 @@ jobs:
             os: ubuntu-latest
           - name: openssl
             os: macos-latest
+          - name: openssl-dimpl
+            os: ubuntu-latest
+          - name: openssl-dimpl
+            os: macos-latest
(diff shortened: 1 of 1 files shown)


efer-ms:
We should add a CI to test `openssl-dimpl`.

efer-ms:
> LGTM, but it's missing documentation on the new feature flag in `lib.rs` (followed by `cargo readme > README.md`)

Updated. Also fixed mention of CommonCrypto for apple-crypto.. since it uses CryptoKit (mostly, only AES Counter Mode needs CommonCrypto).

PR #921, opened 2026-03-26
Add local_init_string for snap

xnorpx:
Realized I had refactored out this API, so adding it back to make my life easier.


src/sctp/mod.rs
```diff
@@ -373,8 +373,7 @@ impl RtcSctp {
             // Established non-SNAP association — MUST NOT inject sctp-init.
             return None;
         }
-        let bytes = d.local_init.as_ref()?;
-        Some(b64_encode(bytes))
+        d.local_init.as_ref().map(|b| b64_encode(b))
```
algesten, line 376: This is the same?!

PR #925, opened 2026-03-28
Add DTLS cipher suite configuration to RtcConfig

algesten:
## Summary
- Add `Dtls12CipherSuite` and `Dtls13CipherSuite` enums (provider-agnostic, defined in str0m-proto)
- Add `DtlsConfig` struct bundling DTLS version + optional cipher suite allow-lists
- Add `RtcConfig::set_dtls12_cipher_suites()` and `set_dtls13_cipher_suites()` methods
- Update `DtlsProvider::new_dtls()` trait method to accept `&DtlsConfig`
- All 5 dimpl backends forward cipher preferences to dimpl's `Config` builder
- Native backends (OpenSSL, SChannel) accept but ignore cipher suite fields

Existing `set_dtls_version()` / `dtls_version()` continue to work, delegating to the `DtlsConfig` struct.

## Test plan
- [x] `cargo check` passes
- [x] `cargo test` passes (all 56 tests)
- [x] `cargo test --test handshake-direct` passes (all 9 handshake tests)
- [ ] Test with restricted cipher suites in a real handshake scenario

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/proto/src/crypto/mod.rs
+++ b/crates/proto/src/crypto/mod.rs
@@ -1,9 +1,8 @@
 mod provider;
 pub use provider::{AeadAes128Gcm, AeadAes128GcmCipher, AeadAes256Gcm, AeadAes256GcmCipher};
 pub use provider::{Aes128CmSha1_80, Aes128CmSha1_80Cipher, CryptoProvider, CryptoSafe};
-pub use provider::{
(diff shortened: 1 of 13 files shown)


xnorpx:
@algesten I don't think we need this now when we can disable cryptos with the feature flags. (But others might want it)

PR #929, opened 2026-03-30
Drop stale depacketizer state on stream pause

davibe:
This resets the packetizers so we dont't emit stale media on pause/unpase cycles.

It's also restoring a dead branch which was fixing timestamps in some cases. Previously we were doing

```
if self.paused {
    self.paused = false;
    self.need_paused_event = true;
}

...

if self.paused && Some(time_u32) < previous_time {
    // DEAD BRANCH
}
```

### DIFF
--- a/src/streams/receive.rs
+++ b/src/streams/receive.rs
@@ -389,7 +389,8 @@ impl StreamRx {
     ) -> RegisterUpdateReceipt {
         self.last_used = now;
 
-        if self.paused {
+        let was_paused = self.paused;
+        if was_paused {
             self.paused = false;
             self.need_paused_event = true;
         }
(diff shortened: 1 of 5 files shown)


src/streams/receive.rs
```diff
@@ -412,7 +413,7 @@ impl StreamRx {
         // Calculate the extended timestamp
         let mut time_u32 = extend_u32(previous_time, header.timestamp);
 
-        if self.paused && Some(time_u32) < previous_time {
+        if was_paused && Some(time_u32) < previous_time {
```
lolgesten, line 416: Is this fixing some bug, or what's going on here?

lolgesten (approved):

algesten (approved):

src/streams/receive.rs
davibe: I have explained in the GH issue description, maybe you were already reviewing.

This regressed to dead code, my change restores the intended behaviour

```
if self.paused {
    self.paused = false;
    self.need_paused_event = true;
}

...

if self.paused && Some(time_u32) < previous_time {
    // DEAD BRANCH
}
```

PR #931, opened 2026-04-04
refactor(stun): reduce bounds checks in encode/decode

mstyura:
These changes make compiler eliminate a lot of bounds check when accessing individual indices.
The idea was simple, since each index operation may panic just do the panic early via `assert!()` of the minimum expected buffer size. Relying on copying slices make compiler to inject single bounds checke rather than separate checks for each index read/write.

Links to Godbolt with coder generation before and after: ~~[encode_xor](https://rust.godbolt.org/z/arb3Px77M), [decode_xor](https://rust.godbolt.org/z/8oWE3qfvf)~~, updated links to latest state of MR: [encode_xor](https://rust.godbolt.org/z/W8xPxeTe1), [decode_xor](https://rust.godbolt.org/z/aoMhG69ss)

### DIFF
--- a/crates/is/src/stun.rs
+++ b/crates/is/src/stun.rs
@@ -431,7 +431,7 @@ impl<'a> StunMessage<'a> {
         }
 
         // Custom attributes
-        self.attrs.to_bytes(&mut buf, &self.trans_id.0)?;
+        self.attrs.to_bytes(&mut buf, &self.trans_id)?;
 
         if include_message_integrity {
             // Message integrity
@@ -473,7 +473,7 @@ impl<'a> StunMessage<'a> {
     }
 }
 
-const MAGIC: &[u8] = &[0x21, 0x12, 0xA4, 0x42];
+const MAGIC: &[u8; 4] = &[0x21, 0x12, 0xA4, 0x42];
 
 #[derive(Clone, Copy, Debug, PartialEq, Serialize, Deserialize)]
 pub enum Class {
@@ -766,7 +766,7 @@ impl<'a> Attributes<'a> {
             + error_code
     }
 
-    fn to_bytes(self, out: &mut dyn Write, trans_id: &[u8]) -> io::Result<()> {
+    fn to_bytes(self, out: &mut dyn Write, trans_id: &TransId) -> io::Result<()> {
         if let Some(v) = self.username {
             out.write_all(&Self::USERNAME.to_be_bytes())?;
             encode_str(Self::USERNAME, v, out)?;
@@ -1093,7 +1093,7 @@ fn decode_str(typ: u16, max_bytes: usize, buf: &[u8], len: usize) -> Result<&str
     }
 }
 
-fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
+fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &TransId) -> usize {
     let port = addr.port() ^ 0x2112;
     buf[2..4].copy_from_slice(&port.to_be_bytes());
     buf[1] = if addr.is_ipv4() { 1 } else { 2 };
@@ -1112,7 +1112,7 @@ fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
                 ip_buf[i] = bytes[i] ^ MAGIC[i];
             }
             for i in 4..16 {
-                ip_buf[i] = bytes[i] ^ trans_id[i - 4];
+                ip_buf[i] = bytes[i] ^ trans_id.0[i - 4];
             }
             20
         }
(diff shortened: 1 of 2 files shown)


crates/is/src/stun.rs
```diff
@@ -1100,45 +1100,38 @@ fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
     let ip_buf = &mut buf[4..];
     match addr {
         SocketAddr::V4(v) => {
-            let bytes = v.ip().octets();
-            for i in 0..4 {
-                ip_buf[i] = bytes[i] ^ MAGIC[i];
-            }
```
thomaseizinger, line 1106: So the bounds-checking happens because we were operating on a slice here, correct? What if we don't slice `buf` and instead directly access `buf` with `i + 4`? Can the compiler then infer that we will never be out of bounds here?

crates/is/src/stun.rs
```diff
@@ -1100,45 +1100,38 @@ fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
...
+    let port = u16::from_be_bytes([buf[2], buf[3]]) ^ 0x2112;
     let ip_buf = &buf[4..];
     let ip = match buf[1] {
         1 => {
-            let mut bytes = [0_u8; 4];
-            for i in 0..4 {
-                bytes[i] = ip_buf[i] ^ MAGIC[i];
-            }
-            IpAddr::V4(bytes.into())
+            let ip_bytes: [u8; 4] = ip_buf[..4].try_into().expect("IPv4 XOR buffer too short");
```
thomaseizinger: So this just panics if `buf` is too short? I realise we were also doing that before but why not return an error if we can? Probably more a question for @algesten.

thomaseizinger (commented):
Nice work! I have two questions.

crates/is/src/stun.rs
mstyura: I've just preserved old behaviour, but totally agree that returning error makes perfect sence.

crates/is/src/stun.rs
mstyura: Sorry for misleading PR description, in the `encode_xor` there were no bound checking triggered in this branch. All of bounds checks were coming from indexing `trans_id` in IPv6 branch.

crates/is/src/stun.rs
mstyura: Updated PR to keep and old for loop, but made compiler aware of size of the slices by providing compiler information via type (TransId for encode_xor) and by doing explicit precondition check for decode_xor.

crates/is/src/stun.rs
```diff
@@ -1112,25 +1112,43 @@ fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
...
+    if buf.len() < 4 {
+        return Err(StunError::Parse(format!(
+            "XOR buffer too short: {} < 4",
+            buf.len()
+        )));
+    }
+    let port = u16::from_be_bytes([buf[2], buf[3]]) ^ 0x2112;
     let ip_buf = &buf[4..];
     let ip = match buf[1] {
         1 => {
```
thomaseizinger, line 1132:
```suggestion
        1 if ip_buf.len() >= 4 => {
```

Does this achieve the same thing? Would be shorter because we can reuse the existing error branch of the `match`.

thomaseizinger: Also, we should add a test that ensures this doesn't regress.

crates/is/src/stun.rs
```diff
@@ -1112,25 +1112,43 @@ fn encode_xor(addr: SocketAddr, buf: &mut [u8; 20], trans_id: &[u8]) -> usize {
...
+                ip_buf[i] = bytes[i] ^ trans_id.0[i - 4];
             }
             20
         }
     }
 }
 
 fn decode_xor(buf: &[u8], trans_id: TransId) -> Result<SocketAddr, StunError> {
-    let port = (((buf[2] as u16) << 8) | (buf[3] as u16)) ^ 0x2112;
+    if buf.len() < 4 {
```
thomaseizinger, line 1123: Also add a test for this case please (i.e. that we don't panic if buffer is too short)

thomaseizinger (commented):
Thank you for addressing the feedback. I have a few more suggestions / questions.

crates/is/src/stun.rs
thomaseizinger: > Would be shorter because we can reuse the existing error branch of the `match`.

If we do that, we should tweak the error message to also say something about length.

mstyura: if we wan't 3 different error messages for 3 different cases: unknown IP version, short IPv4 buffer and short IPv6 buffer I think the guard if will not very helpful here.

thomaseizinger: I don't think we need 3 different error messages. We can make one generic one that says: "Unknown IP version ($version) or insufficient buffer length for it ($len)".

algesten:
@mstyura can we take this all the way? @thomaseizinger got some good points to address.

mstyura:
Yes, sorry for delay. Was busy this week, will address till Monday.

algesten:
@thomaseizinger @mstyura I believe all concerns are addressed. I think we can resolve comments and merge?

thomaseizinger (approved):
Thank you for addressing the feedback!

PR #934, opened 2026-04-13
Fix RUSTSEC-2026-0097

k0nserv:


algesten (approved):
I see. They released a 0.9.x fix.

PR #936, opened 2026-04-16
feat(ice): allow keeping remote candidates on ICE restart

thomaseizinger:
Currently, `IceAgent::ice_restart` allows to retain only the local candidates. The remote candidates are always cleared. This is somewhat limiting for more advanced use-cases of the ICE agent that does not rely on SDP offers triggering the ICE restart.

When the local agent roams, only the local candidates change. As such, it makes sense to only clear the local ones and form new pairs using the existing remote candidates.


thomaseizinger:
Obsolete now that we have #937

PR #940, opened 2026-04-17
Track SSRC and RTX SSRC acks separately

### LINKED ISSUE #938: Frame drops in Firefox related to remote_acked_ssrc
Thanks so much for sharing this great library!

### Issue

I have an application that streams video to clients using str0m that works fine in chrome but has bad performance in Firefox. Lots of frames are dropped after small packet losses that don't affect performance in Chrome.

I'm using Firefox 149 and Chromium 147.0.7727.55 and str0m 0.18.0

### Investigation

I bisected the problem to this commit: [Add simulcast send support](https://github.com/algesten/str0m/commit/425a851a2de2d3af378c15b55d7e9ed16ca6a1b7). In particular this section:

```
        if !remote_acked_ssrc {
            header_ref.ext_vals.mid = Some(mid);
            header_ref.ext_vals.rid = rid;
        }
```
If I change it to `if true` then the issue disappears.

Something else that might be interesting is that `about:webrtc` shows a big increase in "Discarded packets" after the dropped frames. I'm interpreting that as packets that are successfully muxed but can't be used?

### Hypothesis

We're tracking whether the SSRC is acked with that flag, but it's not the only SSRC. There's also the RTX SSRC: `rtx: Option<Ssrc>`. What if the `remote_acked_ssrc` flag is set, then we are sending RTX SSRC transmissions without mid and Firefox doesn't know what to do with them?

To test this, I just always set the mid in the Resend/Blank case to ensure the mid was always attached with using the RTX SSRC, but left everything else unchanged: https://github.com/dmm/str0m/commit/07624b924ddcde1f7249c108e172ae1a62f27805

That change fixes the issue in Firefox, and Chrome still works fine.

Though it seems like a real fix would be to track acks per SSRC?

dmm:
This more closely matches the behavior of libwebrtc: https://webrtc.googlesource.com/src/+/refs/heads/main/modules/rtp_rtcp/source/rtp_sender.cc#696

resolves #938

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix frame drops in Firefox due missing RTX SSRC #940
   * Drop stale depacketizer state on stream pause and restore paused timestamp repair #929
 
 # 0.18.0
--- a/src/streams/send.rs
+++ b/src/streams/send.rs
@@ -131,6 +131,9 @@ pub struct StreamTx {
     /// that the receiver has bound the Mid/Rid tuple to the SSRC and no longer
     /// needs to be sent on every packet
     remote_acked_ssrc: bool,
+
+    // Same as `remote_acked_ssrc`, but for the RTX SSRC
+    remote_acked_rtx_ssrc: bool,
 }
 
 impl StreamTx {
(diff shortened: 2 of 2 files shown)


algesten (approved):

algesten:
Thank you so much! Excellent finding and fixing!

PR #941, opened 2026-04-19
Sdpapi stop media

bekriebel:
I found that with the current SDP Api, it did not seem possible to completely disable media lines in a way that allowed the browser to completely release them.

1. To demonstrate: check out the first commit in this PR (a483369), this sets media lines to Inactive when a user leaves
2. Run the chat example
3. Connect a user
4. Open chrome://webrtc-internals and observe that there are no inbound-rtp entries
5. Connect a second user and start cam & mic
6. Refresh webrtc-internals for the first user and observer that there is one inbound-rtp (kind=video) and one inbound-rtp (kind=audio) - expected result
7. Close the tab of the second user and wait about 30s to ensure RTP has timed out
8. Refresh webrtc-internals and observe that there is still one inbound-rtp (kind=audio) entry
9. Repeat steps 5 - 8, you will see that each time the audio media line is stranded in the stats and not fully released

It is apparently a quirk of Chrome that the video media line is automatically closed when no packets are being received, but audio media lines remain opened and stats tracked (though frozen).

To fix this, implement `disable_media()` for the `SdpApi` that will set the direction to Inactive and also set `port=0` (which indicates `stopped` per RFCs and browser standards).

This did require some refactoring in how tracks are determined to be disabled. Previously, this was true if `self.remote_pts.is_empty()`. However, an offer will be rejected if the `remote_pts` contains no entries (this is only valid in answers to indicate rejection).

The third commit updates the chat example to instead use this new method to disable the media lines instead of just setting them to inactive. You can re-run the test above to see that the inbound-rtp entries are now properly released by the browser. I kept the chat changes separate so they can be removed if you don't actually want them to be included, however I think it is a fine change to keep in the example.

### DIFF
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -355,6 +355,30 @@ impl<'a> SdpApi<'a> {
         }
     }
 
+    /// Stop an already existing media.
+    ///
+    /// The next generated offer emits the m-line with port 0 and excludes
+    /// it from the BUNDLE group, per [RFC 8843] §7.5.3. The remote
+    /// transceiver transitions to the "stopped" state and the m-line slot
+    /// becomes eligible for recycling.
+    ///
+    /// Unlike [`SdpApi::set_direction()`] with [`Direction::Inactive`],
+    /// a stopped m-line cannot be reactivated.
+    ///
+    /// If the media doesn't exist, or is already stopped, [`SdpApi::apply()`]
+    /// will not require a negotiation.
+    ///
+    /// [RFC 8843]: https://datatracker.ietf.org/doc/html/rfc8843#section-7.5.3
+    pub fn stop_media(&mut self, mid: Mid) {
+        let changed = self.rtc.session.stop_media(mid);
+
+        if changed {
+            self.changes
+                .0
+                .push(Change::Direction(mid, Direction::Inactive));
+        }
+    }
+
     /// Add a new reliable ordered data channel and get the `id` that will be used.
     ///
     /// Use `add_channel_with_config` when unreliable or unordered data channels are preferred.
@@ -1219,7 +1243,7 @@ fn update_media(
                 media.mid()
             );
         }
-        media.mark_disabled();
+        media.mark_stopped();
         media.set_direction(Direction::Inactive);
         return;
     }
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1078,8 +1078,14 @@ fn add_pending_changes(session: &mut Session, pending: Changes) {
 /// Compares m-lines in Sdp with that already in the session.
 ///
 /// * Existing m-lines can apply changes (such as direction change).
-/// * New m-lines are returned to the caller.
-fn sync_medias<'a>(session: &mut Session, sdp: &'a Sdp) -> Result<Vec<&'a MediaLine>, String> {
+/// * New m-lines are returned to the caller paired with the session
+///   index they should occupy. The index is normally the next free slot,
+///   but can be a recycled slot (RFC 8829 §5.2.2) when the remote has
+///   replaced a previously disabled Media with a new mid.
+fn sync_medias<'a>(
+    session: &mut Session,
+    sdp: &'a Sdp,
+) -> Result<Vec<(usize, &'a MediaLine)>, String> {
     let mut new_lines = Vec::with_capacity(sdp.media_lines.len());
     let bundle_mids = sdp.bundle_mids();
 
@@ -1111,6 +1117,19 @@ fn sync_medias<'a>(session: &mut Session, sdp: &'a Sdp) -> Result<Vec<&'a MediaL
 
                     continue;
                 }
+
+                // Unknown mid at an index held by a disabled Media means
+                // the remote has recycled the slot (RFC 8829 §5.2.2).
+                // Retire the disabled Media so the replacement takes its
+                // place at the same index.
+                if let Some(pos) = session.medias.iter().position(|l| l.index() == idx) {
+                    if !session.medias[pos].disabled() {
+                        return index_err(m.mid());
+                    }
+                    let retired_mid = session.medias[pos].mid();
+                    session.medias.swap_remove(pos);
+                    session.streams.remove_streams_by_mid(retired_mid);
+                }
             }
             _ => {
                 continue;
@@ -1118,7 +1137,7 @@ fn sync_medias<'a>(session: &mut Session, sdp: &'a Sdp) -> Result<Vec<&'a MediaL
         }
 
         // Second, discover new m-lines.
-        new_lines.push(m);
+        new_lines.push((idx, m));
     }
 
     fn index_err<T>(mid: Mid) -> Result<T, String> {
@@ -1129,14 +1148,17 @@ fn sync_medias<'a>(session: &mut Session, sdp: &'a Sdp) -> Result<Vec<&'a MediaL
 }
 
 /// Adds new m-lines as found in an offer or answer.
+///
+/// Each entry in `new_lines` pairs an m-line with the session index it
+/// should occupy.
 fn add_new_lines(
     session: &mut Session,
-    new_lines: &[&MediaLine],
+    new_lines: &[(usize, &MediaLine)],
     is_offer: bool,
     bundle_mids: Option<&[Mid]>,
 ) -> Result<(), String> {
-    for m in new_lines {
-        let idx = session.line_count();
+    for (idx, m) in new_lines {
+        let idx = *idx;
 
         if m.typ.is_media() {
             let mut media = Media::from_remote_media_line(m, idx, is_offer);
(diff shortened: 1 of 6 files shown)


src/media/mod.rs
```diff
@@ -102,6 +102,13 @@ pub struct Media {
     /// If this is empty, the m-line is disabled/rejected (port=0 in SDP).
     remote_pts: Vec<Pt>,
 
+    /// Set when this m-line has been explicitly disabled via
+    /// [`SdpApi::disable_media`](crate::change::SdpApi::disable_media) or
+    /// rejected by the remote peer. Independent of `remote_pts` so that
+    /// explicit disable preserves the negotiated PT list in SDP output
+    /// (the SDP grammar requires at least one fmt on a port=0 m-line).
+    disabled: bool,
```
algesten, line 110: **Naming suggestion: `stopped` rather than `disabled`.**

`SdpApi` otherwise mirrors `RTCPeerConnection`. The browser analog of this feature is `RTCRtpTransceiver.stop()`, which is terminal/one-way and produces exactly the SDP this PR produces. `disabled` suggests reversibility (contrast `Direction::Inactive`, which *is* reversible); `stopped` communicates the one-way nature and lines up with the spec vocabulary.

Concretely:

- field `disabled: bool` → `stopped: bool`
- Keep `Media::disabled()` returning `self.stopped || self.remote_pts.is_empty()` — that already matches the SDP-level "port=0 in output" concern which is what `MediaLine.disabled` in `src/sdp/data.rs` means.

Two clean concepts: `stopped()` (terminal transceiver state) and `disabled()` (SDP serialization outcome).

src/media/mod.rs
```diff
@@ -282,16 +289,17 @@ impl Media {
...
     ///
     /// SDP level property.
     pub fn disabled(&self) -> bool {
-        self.remote_pts.is_empty()
+        self.disabled || self.remote_pts.is_empty()
     }
 
     pub(crate) fn mark_disabled(&mut self) {
-        self.remote_pts.clear();
+        self.disabled = true;
```
algesten, line 302: **Preserving `remote_pts` here is correct per RFC 4566 §5.14 (port=0 still needs ≥1 fmt), but it shifts one piece of observable semantics:**

`Media::remote_pts()` is **public API**. Before this PR, a remote-rejected mid returned `&[]`. After, it returns the negotiated list. Downstream code inferring "disabled ⇒ no PTs" now sees different behaviour. Technically a breaking change, but maybe ok to ship as patch version.

src/change/sdp.rs
```diff
@@ -355,6 +355,28 @@ impl<'a> SdpApi<'a> {
...
+    /// from the BUNDLE group, per [RFC 8843] §7.5.3. This is stronger than
+    /// [`SdpApi::set_direction()`] with [`Direction::Inactive`]: the remote
+    /// transceiver transitions to the "stopped" state and the m-line slot
+    /// becomes eligible for recycling by a subsequent [`SdpApi::add_media()`].
+    ///
+    /// If the media doesn't exist, or is already disabled, [`SdpApi::apply()`]
+    /// will not require a negotiation.
+    ///
+    /// [RFC 8843]: https://datatracker.ietf.org/doc/html/rfc8843#section-7.5.3
+    pub fn disable_media(&mut self, mid: Mid) {
```
algesten, line 370: **Two nits:**

1. Rename `disable_media` → `stop_media` to match the browser API (see note on the `stopped` field). One-way operation; the name should say so.
2. The docstring claim *"the m-line slot becomes eligible for recycling by a subsequent `SdpApi::add_media()`"* is aspirational — `add_media` unconditionally allocates a fresh mid at `session.line_count()` and we never recycle the slot. Suggest dropping the sentence, or rephrasing as *"the remote transceiver becomes eligible for the peer to reuse the slot via its own `addTransceiver`; str0m itself does not currently recycle."*

algesten (dismissed):
_Deleted because it was all wrong_

algesten (changes requested):
Thanks for this — the motivation is solid and the RFC 8843 §7.5.3 implementation looks correct (port=0 emitted, BUNDLE group excludes the mid, test asserts both). A few suggestions in line below on naming/docs, plus one bigger concern on browser interop that I'll explain here.

(Note: I posted an earlier review that I've now dismissed — it relied on an incorrect picture of browser behaviour that I got from shallow reading of pion/Janus rather than libwebrtc itself. This version is corrected.)

**Browser interop — libwebrtc does recycle m-sections, and this PR creates the conditions for it to happen against us.**

Direct evidence from libwebrtc source (`pc/sdp_offer_answer.cc` on main):

- `IsMediaSectionBeingRecycled` at line 298-305 defines the eligibility check (offer type, new section not rejected, old section rejected in local or remote description).
- The offer-construction loop at line 4744-4770 explicitly prefers a recyclable slot over appending, and generates a fresh mid via `mid_generator_.GenerateString()` when it recycles.
- `RemoveStoppedTransceivers` at line 5433-5472 dissociates the stopped transceiver before recycling.
- The test `RecycleMediaSectionWhenStoppingTransceiverOnAnswerer` in `pc/peer_connection_jsep_unittest.cc:659-715` asserts the recycled offer has only the new m-line (not the stopped port=0 one alongside) with a different mid.
- The sibling test `CreateOfferDoesNotRecycleMediaSectionIfFirstStopped` at line 636-654 pins the trigger: recycling only fires once the stop has been round-tripped via an offer/answer with port=0.

**Why pion and Janus look like they don't have this problem:**

- **pion** deviates from spec: `RTPTransceiver.Stop()` at `rtptransceiver.go:268` only sets direction to Inactive, never emits port=0. Their comment at `rtptransceiver.go:349-352` flags this as intentional to avoid SDP inflation. Browsers therefore never see the stop round-trip and never recycle against pion. Pion also happens to tolerate an incoming recycle because its answer generation iterates the remote SDP at `peerconnection.go:2976`, not its own transceiver list.
- **Janus** is stateless at the SDP layer. `janus_sdp_generate_answer_default` at `sdp-utils.c:1885` mirrors the offer's m-lines one-to-one. No internal state to disagree with a recycled mid.

**What breaks for str0m with this PR specifically:**

The risk is the combination of two things:

1. We properly signal stops via port=0 (this PR's new behaviour), which **enables** browser recycling against us.
2. `add_new_lines` in `src/change/sdp.rs:1115` uses `session.line_count()` to assign the index to an unknown mid, which assumes the incoming SDP's m-lines are always a suffix of ours. Under recycling they're not — the browser reuses an existing index with a new mid.

Concretely: if str0m stops a mid, the stop round-trips, and later the browser adds a new transceiver and offers, Chrome will recycle the slot. str0m's `sync_medias` won't find the new mid, falls into `add_new_lines`, which assigns it `line_count()` (wrong index, collides with the stopped Media still in `session.medias`). The subsequent answer emits both m-lines where the offer had one, and the browser rejects.

In SFU topologies where str0m is always the offerer, this never fires — browsers don't recycle in their answers. In P2P or mixed topologies where the browser can offer after a stop round-trip, it will.

**Recommendation:**

We need to add receive-side recycling in `sync_medias` / `add_new_lines` before shipping — detect an unknown mid at an index str0m considers stopped, dissociate the stopped `Media`, and use the offer's index rather than `line_count()`.

algesten:
I think the easy life is to not rock the boat with `port=0` and do what Pion does, but we can of course do it to be more compliant. It requires a bunch of integration tests.

bekriebel:
The concern makes sense. In my use case, I'm using Inactive direction on peers when "stopping" tracks, but maintaining state of those trancievers and reusing existing ones myself if they are compatible. I only plan to call the new disable/stop method when a peer is disconnected and I know those trancievers are truly gone. However, it should certainly handle the use case of it being used on a still-active peer.

I'll see if I can simulate the situation we are concerned about with a proper test and try to fully plumb that logic through.

I also need to finish testing on my local codebase. I encountered this while chasing a bug in my implementation, but I'm not convinced this was the actual root cause of my issue yet 😅. Still, it seemed worth trying to implement either way.

bekriebel:
Keeping this a draft for a bit longer as I'm checking one other aspect

bekriebel:
I finished testing and found that it actually isn't recycling that is causing failures in stopping browser peer transcievers, it's the clearing of the PT values. This can be reproduced easily on the chat example on the main branch:

1. Start the chat example
2. Connect a client with chrome and start cam and mic
3. In a dev console run: 
```js
const videoTx = rtc.getTransceivers().find(t => t.sender?.track?.kind === 'video');`
videoTx.stop();
await negotiate();
```
You'll get an error due to the PTs being empty.

Repeating the same test on this PR works properly, the negotiation can complete because the stopped m-line still has PTs listed.

src/media/mod.rs
bekriebel: The PTs being empty is actually causing a negotiation issue with Chrome, see my updates below.

algesten (approved):

algesten:
Thanks!

PR #942, opened 2026-04-20
sdp: reject m-line recycling in answers + expand recycle tests

algesten:
## Summary

Follow-up to #941. Two changes:

1. **Code fix**: `sync_medias` now rejects m-line recycling on the answer path. Per RFC 8829 §5.2.2, only the offerer recycles slots; an answer that introduces a new mid at the index of a stopped slot is malformed. Before this, `sync_medias` silently retired our stopped `Media` and installed the answerer's new mid — protocol-violating but undetected.

2. **Tests**: Move the recycling tests from `sdp-negotiation.rs` into a dedicated `sdp-m-recycle.rs`, and add coverage for gaps identified in the review thread on #941:
   - Post-recycle `Media` state assertions (kind, direction, `remote_pts`, not stopped)
   - L can generate a follow-up offer after accepting a recycled offer (no lingering inconsistency)
   - Recycling that changes media type (audio slot → video m-line, allowed per JSEP §5.2.2)
   - Answers that recycle a stopped slot are rejected with a clear error

The only behavioural change on the code side is returning an error where we previously accepted silently. No tests on `main` broke.

## Test plan
- [x] `cargo test` — full suite passes (486 + 6 new)
- [x] `cargo test --test sdp-m-recycle` — 6 pass, 0 fail, 0 ignored


algesten:
@k0nserv this one also ready.

bekriebel (approved):
Tested on my end and still works for all the manual and Playwright based testing I was doing. LGTM!

src/change/sdp.rs
```diff
@@ -1120,9 +1121,15 @@ fn sync_medias<'a>(
 
                 // Unknown mid at an index held by a disabled Media means
                 // the remote has recycled the slot (RFC 8829 §5.2.2).
-                // Retire the disabled Media so the replacement takes its
-                // place at the same index.
+                // Recycling is only permitted in offers; an answer that
+                // rewires a mid at a stopped slot is malformed.
                 if let Some(pos) = session.medias.iter().position(|l| l.index() == idx) {
+                    if !is_offer {
```
k0nserv, line 1127: Does our MSRV allow us to combine this with the if above? If not worth leaving a TODO here for when we bump it.

k0nserv (approved):

src/change/sdp.rs
algesten: @copilot does our MSRV allow it?

algesten: This is bullshit. if-let chains stabilized on 1.88

Issue #932, opened 2026-04-07
DTLS handshake fails for most clients in multi-client str0m-to-str0m setup


Teyk0o:
## Environment

  - str0m version: latest (both sides)                                                                                                                                                        
  - Setup: str0m-based SFU server + str0m-based load test client (str0m-stress), both using Rtc::builder().build()                                                                            
  - Network: client runs on a remote machine, SFU in Docker on a VPS (2 cores), UDP ports 16384-16385 mapped through Docker                                                                   
  - Scale: 2-30+ concurrent clients connecting to the same SFU   

## Symptom

ICE completes successfully for all clients (Checking → Connected → Completed), but DTLS only completes for ~20-30% of connections. The remaining clients never get past the DTLS handshake, no data channel opens, no media flows, and the connection eventually times out.

 ## Observed behavior

### Working client (minority):
ICE state -> Connected                                                                                                                                                                      
ICE state -> Completed                                                                                                                                                                      
DTLS setup is: true                                                                                                                                                                         
ClientHello: DTLS version=DTLS1_2, cookie_len=0, offering 3 cipher suites                                                                                                                   
[ServerHello received, handshake completes]                                                                                                                                                 
ChannelOpen("data")                                                                                                                                                                         
MediaData received ✓ 

### Failing client (majority): 
ICE state -> Connected                                                                                                                                                                      
ICE state -> Completed                                                                                                                                                                      
DTLS setup is: true                                                                                                                                                                         
ClientHello: DTLS version=DTLS1_2, cookie_len=0, offering 3 cipher suites                                                                                                                
[No ServerHello — nothing comes back]                                                                                                                                                       
Flight timeout in: 1.189s                                                                                                                                                                   
[Retransmit ClientHello]                                                                                                                                                                    
Flight timeout in: 0.802s                                                                                                                                                                   
[Retransmit again, eventually gives up after connect timeout (40s)] 

## Key data points

1. ICE works fine for all clients — STUN binding succeeds, candidate pairs are validated, consent checks flow at ~1 pps. UDP connectivity is not the issue.                                 
2. The server receives the DTLS ClientHello — our packet dispatch logs confirm the UDP packet is received on the correct socket, dispatched to the correct Rtc instance via                         handle_input(Input::Receive(...)). No packets are dropped.                                                                                                                                  
3. The server does not produce a DTLS response — after handle_input processes the ClientHello, poll_output() does not return a Transmit containing a ServerHello/HelloVerifyRequest. The    
[...]

PR #943, opened 2026-04-20
dtls: fix dropped handshake resend

algesten:
## Summary

Fixes a latent issue where DTLS flight retransmits are delayed (sometimes by many seconds) because `handle_timeout` runs at the end of `do_poll_output` — after both `dtls.poll_packet()` and the DTLS `poll_output` loop have already been consulted.

When a flight timeout elapses, dimpl's `handle_timeout` re-arms the flight timer and queues the retransmit packet in its tx queue. With the previous ordering:

- `dtls.poll_packet()` already ran earlier and returned `None`
- `next_dtls_timeout` is cleared to `None` after `handle_timeout`, so the computed `Output::Timeout` does not reflect the new flight deadline
- The queued retransmit sits until some unrelated subsystem (typically ICE consent checks) wakes the caller

Moving the `handle_timeout` check to run before the DTLS `poll_output` loop lets the loop pick up the retransmit into `pending_packets` and record the refreshed flight timer, and the existing `poll_packet()` call emits the `Transmit` in the same iteration.

Refs #932.

## Test plan

- [x] `cargo build --all-features`
- [x] `cargo test --all-features`
- [ ] Reporter to re-test the multi-client SFU scenario

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1494,6 +1494,16 @@ impl Rtc {
             }
         }
 
+        // Handle DTLS timeout before polling output so any retransmit packets
+        // queued by dimpl and the re-armed flight timer are picked up by the
+        // poll loop below in the same iteration.
+        if let Some(timeout) = self.next_dtls_timeout {
+            if timeout <= self.last_now {
+                let _ = self.dtls.handle_timeout(self.last_now);
+                self.next_dtls_timeout = None;
+            }
+        }
+
         // Poll DTLS output - collect packets, handle events
         let mut just_connected = false;
         loop {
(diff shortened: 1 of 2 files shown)


algesten:
@k0nserv this is ready for review.

k0nserv (approved):

k0nserv:
Are there any tests for the DTLS handshake with a simulated loss? If not it might be good to add that too, seems like a test like that might have also caught this

algesten:
Good call — added one. `dtls_handshake_completes_under_packet_loss` applies 30% random loss in both directions from the very first packet and drives the handshake to completion with a bounded iteration cap. Fails without the fix (handshake stalls past the connect deadline), passes with it.

Both regression tests are gated to dimpl-backed providers since the bug was specific to how str0m drives dimpl's tx queue via `next_dtls_timeout`.

algesten:
Confirmed locally that the 30% loss causes two resends during DTLS handshake.

PR #950, opened 2026-04-30
feat(is): handle role conflicts as per RFC

thomaseizinger:
This implements role conflict resolution as per the ICE RFC. The idea is simple: If both agents want to be the controlling side, the u64 tie breaker value decides which one gets to actually be the controlling side. By default, this tie breaker value is randomly generated but we allow users to change it via the `set_control_tie_breaker` method for advanced use cases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
   * Expose H265 some already referenced profile structs #948
+  * Implement ICE role conflict resolution #950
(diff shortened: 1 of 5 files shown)


thomaseizinger:
Disclaimer: Most of this was written by Claude but I reviewed all changes manually.

algesten (approved):

algesten:
Looks great!

PR #951, opened 2026-04-30
feat(is): expose `Candidate::local` publicly

thomaseizinger:
Right now, it is not possible to replace `default_local_preference` from outside the crate because `Candidate::local` is `pub(crate)`.

### DIFF
--- a/crates/is/src/candidate.rs
+++ b/crates/is/src/candidate.rs
@@ -499,6 +499,20 @@ impl Candidate {
         self.addr
     }
 
+    /// Returns the _local_ address of this ICE candidate.
+    ///
+    /// This is an extension to the ICE spec that we use to track
+    /// local interfaces also for relayed candidates.
+    /// Specifically, for relayed candidates, the local interface
+    /// is the address the agent uses to communicate with the TURN
+    /// server using the TURN protocol.
+    ///
+    /// For all other candidates, this refers to either the
+    /// candidate itself (host) or the base (srflx/prflx).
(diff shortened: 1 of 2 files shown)


thomaseizinger:
@algesten Let me know what you think. Now that the `IceAgent` is in its own crate, we could also just expose this publicly as a documented feature. It already is part of the public API of `Candidate::relayed` after all.

algesten:
Document the semantics precisely and make public?

thomaseizinger:
> Document the semantics precisely and make public?

SGTM, will do!

thomaseizinger:
@algesten Let me know what you think of the docs I wrote.

crates/is/src/candidate.rs
```diff
@@ -499,6 +499,20 @@ impl Candidate {
...
+    ///
+    /// This is an extension to the ICE spec that we use to track
+    /// local interfaces also for relayed candidates.
+    /// Specifically, for relayed candidates, the local interface
+    /// is the address the agent uses to communicate with the TURN
+    /// server using the TURN protocol.
+    ///
+    /// For all other candidates, this refers to either the
+    /// candidate itself (host) or the base (srflx/prflx).
+    pub fn local(&self) -> SocketAddr {
```
thomaseizinger, line 512: I moved it up here so it is next to `fn addr`.

algesten: Sounds good

algesten (approved):

PR #954, opened 2026-05-06
docs: clarify Instant semantics for Input::Receive

mstyura:
The meaning of the instant in the `Input::Receive` was a bit unclearly described in doc and I've found a confirmation in the code and in [previous discussion](https://discordapp.com/channels/1436725650302959829/1436796781462687817/1436796785019322388) that it's actually network receive time, so I propose to consider updating documentation.


algesten (approved):
Good improvement!

PR #955, opened 2026-05-10
feat: add stereo to opus format params

lherman-cs:
Technically, there's sprop-stereo for the sender side. But, I don't think browsers actually use this. I see Google Meet uses `stereo` only on both directions for both Chrome and Firefox user agents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+  * Add stereo to opus format params #955
+
 # 0.19.0
 
   * Expose H265 some already referenced profile structs #948
--- a/src/format/format_params.rs
+++ b/src/format/format_params.rs
@@ -13,10 +13,15 @@ pub struct FormatParams {
 
     /// Opus specific parameter.
     ///
-    /// Allow stereo Opus encoding; actual channel count follows the source track
-    ///
+    /// Decoder-side preference for stereo audio. Default is mono per
     /// RFC 7587 Section 7.1: <https://datatracker.ietf.org/doc/html/rfc7587#section-7.1>
     pub stereo: Option<bool>,
+
+    /// Opus specific parameter.
+    ///
+    /// Sender-side stereo hint for Opus. When set, it signals the intent to
+    /// transmit stereo from the sender, matching the symmetric m-line model.
+    pub sprop_stereo: Option<bool>,
 
     /// Opus specific parameter.
     ///
@@ -125,6 +130,7 @@ impl FormatParams {
         match param {
             MinPTime(v) => self.min_p_time = Some(*v),
             Stereo(v) => self.stereo = Some(*v),
+            SpropStereo(v) => self.sprop_stereo = Some(*v),
             UseInbandFec(v) => self.use_inband_fec = Some(*v),
             UseDtx(v) => self.use_dtx = Some(*v),
             LevelAsymmetryAllowed(v) => self.level_asymmetry_allowed = Some(*v),
@@ -150,6 +156,9 @@ impl FormatParams {
         }
         if let Some(v) = self.stereo {
             r.push(Stereo(v));
+        }
+        if let Some(v) = self.sprop_stereo {
+            r.push(SpropStereo(v));
         }
         if let Some(v) = self.use_inband_fec {
             r.push(UseInbandFec(v));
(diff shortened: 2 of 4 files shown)


algesten:
Thanks for picking this up. A few things to address before merging:

### 1. The `stereo` doc comment is technically wrong

> `Allow stereo Opus encoding; actual channel count follows the source track`

Per [RFC 7587 §7.1](https://datatracker.ietf.org/doc/html/rfc7587#section-7.1), `stereo` is a **decoder-side preference** ("the decoder prefers receiving stereo signals"), with default 0 (mono). It does not, by itself, configure encoding. The encoder-side hint is `sprop-stereo`.

What does happen in practice is that libwebrtc's send path treats a remote `stereo=1` as a trigger to encode stereo. From [`audio_encoder_opus.cc`](https://chromium.googlesource.com/external/webrtc/+/refs/heads/main/modules/audio_coding/codecs/opus/audio_encoder_opus.cc):

```cpp
int GetChannelCount(const SdpAudioFormat& format) {
  const auto param = GetFormatParameter(format, "stereo");
  if (param == "1") {
    return 2;
  } else {
    return 1;
  }
}
```

And [`webrtc_voice_engine.cc::UpdateSendCodecSpec`](https://webrtc.googlesource.com/src/+/refs/heads/main/media/engine/webrtc_voice_engine.cc):

```cpp
// Encoder will only use two channels if the stereo parameter is set.
const auto& it = send_codec_spec.format.parameters.find("stereo");
if (it != send_codec_spec.format.parameters.end() && it->second == "1") {
  num_encoded_channels_ = 2;
}
```

So adding `stereo` is functional for the common "make Chrome send us stereo" case — but the doc should describe what the parameter actually is (a decoder preference) and note that browser encoders treat it as a trigger.

### 2. We need `sprop-stereo` too — m-lines are symmetric

str0m treats m-lines as symmetric (a single PT carries both send and recv direction), so leaving out `sprop-stereo` is not really an option: without it, str0m has no way to signal its own send-side stereo intent to a remote. The claim in the PR description that browsers don't use `sprop-stereo` is too strong:

- [Mozilla bug 1524145](https://bugzilla.mozilla.org/show_bug.cgi?id=1524145) is a direct regression report where Firefox stopped sending stereo when `sprop-stereo=1` was missing — i.e., Firefox does honor it.
- [Mozilla bug 818618](https://bugzilla.mozilla.org/show_bug.cgi?id=818618) is the original Opus stereo SDP handling implementation in Firefox covering both `stereo` and `sprop-stereo`.
- [w3c/webrtc-extensions#63](https://github.com/w3c/webrtc-extensions/issues/63) treats `stereo=1; sprop-stereo=1` as the canonical pair browsers should expose without SDP munging.

Google Meet may emit only `stereo` (perhaps because it doesn't itself produce stereo, or for older-endpoint compat), but generalizing that to "browsers don't use sprop-stereo" doesn't hold up.

Please add `SpropStereo(bool)` as a sibling `FormatParam`, with a `sprop_stereo: Option<bool>` field on `FormatParams`, and round-trip it the same way as `stereo`.

For reference: [pion's default Opus registration](https://github.com/pion/webrtc/blob/master/mediaengine.go) is `"minptime=10;useinbandfec=1"` — neither parameter by default. [pion/sdp util.go](https://github.com/pion/sdp/blob/master/util.go) keeps fmtp as an opaque string, so it doesn't constrain us either way.

### 3. The match logic comparison is wrong

The new block in `match_opus_score`:

```rust
let either_stereo_specified = c0.format.stereo.is_some() || c1.format.stereo.is_some();
if either_stereo_specified && c0.format.stereo != c1.format.stereo {
    score = score.saturating_sub(8);
}
```

This penalizes `Some(false)` vs `None`, even though both semantically mean "mono" per RFC 7587 (default 0). The existing Opus params have the same bug, but please don't extend the pattern. Match what AV1 does at lines 369–386 and compare with explicit defaults:

```rust
let c0_stereo = c0.format.stereo.unwrap_or(false);
let c1_stereo = c1.format.stereo.unwrap_or(false);
if c0_stereo != c1_stereo {
    score = score.saturating_sub(8);
}
```

Same treatment should apply for `sprop_stereo` once added.

Also: a penalty of 8 is the largest among the Opus params (minptime=1, FEC=2, DTX=4). Stereo mismatch is more impactful than FEC/DTX preference mismatch, so 8 is defensible, but worth a sentence justifying it in a comment.

### Minor

- Test `opus_default_to_mono` only asserts that `Default` produces no `stereo` field — the name implies more than the assertion checks. Either rename to `opus_default_stereo_absent` or actually assert the mono-equivalent semantics (`parsed.stereo.unwrap_or(false) == false`).

algesten:
Sorry about the AI answer. I do stand by all the points though ;)

lherman-cs:
I added the sprop-stereo format.

When I read those references, I still read "sprop-stereo" is spec-compliant, but not really used in practice. It sounds more like we leave the "stereo" vs "mono" decision to the opus decoder on the receiving end. I agree to follow the spec anyway, though.

algesten (approved):
This looks good to me now.

Thanks!

algesten:
@lherman-cs can you add row in the CHANGELOG.md for this change? There should be a new section at the top

```text
# Unreleased

  * Open feat: add stereo to opus format params #955

# 0.19.0
…
```

lherman-cs:
@algesten done.

PR #956, opened 2026-05-12
Handle DTLS close_notify shutdown

kvasilye:
Summary

  * Send DTLS close_notify alerts from Rtc::close().
  * Emit Event::Closed when a DTLS close_notify is received and respond with a reciprocal alert.
  * Wire close handling through the DTLS provider implementations and document the change in the changelog.

### DIFF
--- a/crypto/wincrypto/src/sys/provider.rs
+++ b/crypto/wincrypto/src/sys/provider.rs
@@ -409,4 +409,10 @@ impl DtlsInstance for WinCryptoDtlsInstance {
     fn is_active(&self) -> bool {
         self.dtls.is_client().unwrap_or(false)
     }
+
+    fn close(&mut self) -> Result<(), DtlsImplError> {
+        Err(DtlsImplError::CryptoError(
+            "SChannel native DTLS does not support close_notify".into(),
+        ))
+    }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Send DTLS `close_notify` on `Rtc::close()` #956
   * Fix `Simulcast::add_recv_layer` to push to recv instead of send #968
   * Accept any non-empty SDP session name (`s=`), not just `s=-` (RFC 8866 section 5.3)
 
--- a/crates/proto/src/crypto/provider.rs
+++ b/crates/proto/src/crypto/provider.rs
@@ -192,6 +192,9 @@ pub trait DtlsInstance: CryptoSafe {
 
     /// Return the negotiated DTLS protocol version. May return `None` before handshake completion.
     fn protocol_version(&self) -> Option<ProtocolVersion>;
+
+    /// Initiate graceful shutdown by sending a DTLS `close_notify` alert.
+    fn close(&mut self) -> Result<(), DtlsImplError>;
 }
 
 // ============================================================================
--- a/crypto/apple-crypto/src/dtls.rs
+++ b/crypto/apple-crypto/src/dtls.rs
@@ -169,6 +169,10 @@ impl DtlsInstance for AppleCryptoDtlsInstance {
     fn protocol_version(&self) -> Option<ProtocolVersion> {
         self.dtls.protocol_version()
     }
+
+    fn close(&mut self) -> Result<(), DtlsImplError> {
+        self.dtls.close()
+    }
 }
 
 #[cfg(test)]
--- a/crypto/aws-lc-rs/src/dtls.rs
+++ b/crypto/aws-lc-rs/src/dtls.rs
@@ -111,4 +111,8 @@ impl DtlsInstance for AwsLcRsDtlsInstance {
     fn protocol_version(&self) -> Option<ProtocolVersion> {
         self.dtls.protocol_version()
     }
+
+    fn close(&mut self) -> Result<(), DtlsImplError> {
+        self.dtls.close()
+    }
 }
--- a/crypto/openssl/src/dtls_dimpl.rs
+++ b/crypto/openssl/src/dtls_dimpl.rs
@@ -112,4 +112,8 @@ impl DtlsInstance for DimplDtlsInstance {
     fn protocol_version(&self) -> Option<ProtocolVersion> {
         self.dtls.protocol_version()
     }
+
+    fn close(&mut self) -> Result<(), DtlsImplError> {
+        self.dtls.close()
+    }
 }
--- a/crypto/openssl/src/dtls_ossl.rs
+++ b/crypto/openssl/src/dtls_ossl.rs
@@ -434,6 +434,10 @@ pub(super) struct OsslDtlsInstance {
     queued_app_data: VecDeque<Vec<u8>>,
     next_timeout: Option<Instant>,
     connected_emitted: bool,
+    /// Close handling
+    close_notify_received: bool,
+    close_notify_emitted: bool,
+    close_notify_sent: bool,
 }
 
 impl std::fmt::Debug for OsslDtlsInstance {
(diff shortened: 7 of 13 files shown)


algesten:
@kvasilye Thanks for looking at this, and this seems like a promising start.

I want to take a step back and think about what we mean with `Event::Close` – right now it's about DTLS closing, but we don't do any state change on the back of that.

* Is DTLS closing the _only_ signal for transitioning the entire `Rtc` instance to closed?
* Should we make some state change?
* Should we continue to accept media writes/rtp input and/or network input?

kvasilye:
Hello,

- Is DTLS closing the only signal for transitioning the entire Rtc instance to closed?

It means the remote peer tells us that they're done - or if "we" called `close` it means "we" are done.

The incoming close should be propagated to the application (which the new event does) and it's expected that in either case the application will stop working with this `Rtc` object (exception: item 3 below).

- Should we make some state change?

If you want the change to closed to be permanent, we could set some state variable. The existing `alive` won't work though because of next item:

- Should we continue to accept media writes/rtp input and/or network input?

I can only see one use case where an application would want to continue to work with the `Rtc` object for some (sort) duration after receiving a close from the remote: to send the reciprocal DTLS close_notify.

For this, it would need to be able to call `poll_output` (and then send any generated packets - there will be one, close_notify).

Likewise, if the application initiated close from its side, it first needs to send the actual UDP close message which would come from `poll_output`, it may want to wait for the remote to send its close (although it won't change anything really, but it may be the application's "wish"), this requires that `handle_input` processes the DTLS close message.

If the application received the close message from the remote, it may want to send the reciprocal close. For this it also needs `poll_output` to work.

In theory there should not be any incoming media packets after the peer sent us its close, but there could be in practice because of network jitter. If we are the ones initiating close, then there could be incoming media packets until the remote has received and processed our close (and then there could be packets in-flight).

If we added a new state variable - `close_in_progress` or something - in `Rtc` then we could:

- Return an error from media write functions (or ignore any media written) - note that `Rtc::writer` returns `Option` not `Result` so we could return `None` but can't return `Error`
- In `do_poll_output` only process DTLS messages
- Likewise, `do_handle_receive` would only process DTLS messages

What do you think?

crypto/wincrypto/src/sys/provider.rs
```diff
@@ -404,4 +404,10 @@ impl DtlsInstance for WinCryptoDtlsInstance {
     fn is_active(&self) -> bool {
         self.dtls.is_client().unwrap_or(false)
     }
+
+    fn close(&mut self) -> Result<(), DtlsImplError> {
+        Err(DtlsImplError::CryptoError(
```
efer-ms, line 414: Just FYI. This is fine.

Barring objections I plan on removing the schannel code entirely and just making wincrypto use dimpl for the dtls implementation.

kvasilye: Sounds good to me!

kvasilye:
@algesten 

I've updated the PR.
- Added `closing: bool` to `Rtc` - set after generating or receiving `close_notify`.
- If in closing state, only process DTLS messages and ignore media, ice, etc.
- Media writers and channel writers return `None`.

kvasilye:
@algesten 

Martin - would it be possible to take a look at my latest proposed changes?

algesten:
> @algesten
> 
> Martin - would it be possible to take a look at my latest proposed changes?

Sorry, I haven't forgotten, I'm just ostriching. Trying to come to some sort of settlement in my head how to define:

* it's over, this Rtc is shut down
* which signals triggers it
* how should it affect the instance

And I keep going back/forth between this being just informational, to let's tear it all down…

kvasilye:
Anything I can do to help?

Right now there is a separate "closing" state where we don't accept or emit media, but it doesn't transition to "alive = false". 

The reason is that if the close is initiated by the remote, we need to send our own close (reciprocal / confirmation) and even though we know when we emit it from the library into the application, we don't know when the application is done sending it.

Maybe we can stop at the "we emitted the reciprocal close to the application" boundary and mark the Rtc instance as "not alive".

Or maybe we can leave it forever in a "closing" state - we don't accept or emit media in that state anyway, so what are the drawbacks?

algesten:
Right, I think I've landed my thoughts.

A DTLS close is end of the road. Unlike ICE `Disconnected`, which is recoverable (a new candidate shows up and we resume), once close_notify is exchanged the keys are gone and there's no resuming.

So of your two options it's the first - stop once the close is flushed and mark it not alive. Not the forever one. We have one lifecycle bit, `alive`, and apps reap on it (`clients.retain(|c| c.rtc.is_alive())`).

The reciprocal close_notify isn't ours to model. That's DTLS-internal (it's what `SSL_shutdown` and the dimpl state machine do). Whatever bytes it produces come out the normal `poll_output` / `poll_packet` path and we transmit them like any handshake record. `Rtc` doesn't need to know one of them is a reciprocal.

So it collapses to:

- Surface the close signal from DTLS (the `DtlsOutput::CloseNotify` you added). Keep.
- Emit `Event::Closed` as the reason marker. Keep.
- `Rtc::close()` triggers the engine's shutdown, we drain output, flip `alive = false`. Keep.
- Drop `closing: bool`, the media-write `None` guards and the receive-path filtering. Once `alive` is false those paths are inert anyway, and the rest is DTLS-internal.

Don't worry about the SChannel gap, we can take that as a follow-up.

kvasilye:
How can we know that we've emitted the reciprocal DTLS close if the close is initiated from the remote side?

We can't control when the application sends it - which is probably fine - but how do we know when we've emitted our close response (it's all in str0m so it must be possible, I'm just not sure how exactly).

Also when closing from "our" side - in the model you're proposing we'll never let the application receive the remote's reciprocal close because we're going to have `alive = false` right away.

Finally in terms of the API, even if close is initiated from our side, close doesn't return a list of packets - if it sets `alive = false` then `poll_output` will return an empty collection and the application will never know to send it.

Do you want to change `close` to return a list of packets? This would allow us to set alive to false.

Still doesn't solve waiting for the reciprocal close though (which applications should not wait on indefinitely btw, because it can get lost).

kvasilye:
@algesten I've updated the PR

- Kept `closing` in `Rtc` instance - it's necessary to carry state from `close()` to `poll_output()`
- Removed special logic around `closing` except in `poll_output()` which just collects and emits the (expected) DTLS close_notify
- Once `poll_output()` emits the close_notify packet, it marks the `Rtc` instance as `alive = false`
- When "we" initiate the closing from our side there is no way to wait for a reciprocating close from the remote side

The last item is a drawback of this update, but may even be good because that reciprocating close packet can be lost. If the application decided to close a connection, it's not like it's going to change its "mind" if it doesn't receive the reciprocating close.

User tries to close their video meeting app -> "sorry, no reciprocating close in 3000 ms, you can't leave" - absurd, right?

kvasilye:
I got some PR check failures all of a sudden, and then updated from `main` and now getting different failures:

```
error[E0119]: conflicting implementations of trait `From<format_description::parse::format_item::HourBase>` for type `OtherNameValue`
   --> /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rcgen-0.14.8/src/lib.rs:359:1
    |
359 | / impl<T> From<T> for OtherNameValue
360 | | where
361 | |     T: Into<String>,
    | |____________________^
    |
    = note: conflicting implementation in crate `time`:
            - impl From<format_description::parse::format_item::HourBase> for <format_description::parse::format_item::HourBase as format_description::parse::format_item::ModifierValue>::Type;
    = note: upstream crates may add a new impl of trait `std::convert::Into<std::string::String>` for type `time::format_description::parse::format_item::HourBase` in future versions
```

I don't think it's my code, the failure is somewhere else - could you help @algesten ?

algesten:
@kvasilye we had a temporary test failure on main. What happens if you rebase off latest main now?

kvasilye:
@algesten i merged from main and now all checks are passing.

The latest revision is ready for your review, thank you!

algesten (approved):
Approved for merge after conflict resolution, changelog update, and passing CI.

kvasilye:
Thank you Martin!

PR #958, opened 2026-05-19
Expose WinCrypto DTLS provider for external DTLS usage

tyhik:
To be able to externally do DTLS using wincrypto crate with dimpl+CNG, the easiest is to make dimpl_provider public. This change would allow setting MTU and querying the selected DTLS version after the completed DTLS handshake.  As of now, DtlsInstance does not expose these APIs and exposing them via DtlsInstance would need more code changes.


xnorpx:
This should not be solved in this way, please open an issue to start discussion how what APIs is required.

algesten:
@xnorpx talked to me in Discord. We should discuss arch before doing this.

PR #959, opened 2026-05-20
use Arc<[u8]> for RTP and media payloads

ThanhDodeurOdoo:
Forwading currently turn each received payload into owned byte storage at the API boundary and then repeats payload copies when the same packet or frame is passed to several destinnations. This make SFU fanout cost scale with destination count even though the payload bytes are immutable.

Use `Arc<[u8]>` for incoming RtpPacket payloads and MediaData frames. Accept `Into<Arc<[u8]>>` in the direct RTP and media writer APIs so callers can pass `Vec<u8>`, `&[u8]`, `Box<[u8]>` or an already shared `Arc<[u8]>`


src/streams/receive.rs
```diff
@@ -468,7 +468,7 @@ impl StreamRx {
             seq_no,
             time,
             header,
-            payload: data,
+            payload: data.into(),
```
algesten: This I think is a bit dodgy. This causes a reallocation where there before was none. Question is why we have `data: Vec<u8>` - shouldn't we use `Arc<[u8]>` throughout?

src/streams/rtx_cache.rs
```diff
@@ -89,7 +89,7 @@ mod test {
             header: RtpHeader::default(),
             seq_no: seq_no.into(),
             time: MediaTime::from_90khz(0),
-            payload: millis.to_be_bytes().to_vec(),
+            payload: millis.to_be_bytes().to_vec().into(),
```
algesten: Here we had one allocation, now we have two. Unsure if the compiler optimizes away one?

src/streams/send_queue.rs
```diff
@@ -251,7 +251,7 @@ mod test {
             seq_no: 0.into(),
             time: MediaTime::from_90khz(10),
             header: RtpHeader::default(),
-            payload: vec![42, 42],
+            payload: vec![42, 42].into(),
```
algesten: All of these can potentially avoid double allocation by going `[42, 42].into()` or maybe static slice.

algesten (commented):
Looks ok! I left some comments to consider.

src/streams/rtx_cache.rs
ThanhDodeurOdoo: oh nice catch

src/streams/receive.rs
ThanhDodeurOdoo: `data` is still a `Vec<u8>` here because it is mutable working storage until ths point,
after SRTP decrypt, the receive path may truncate SRTP padding with `RtpHeader::unpad_payload(&mut data)` and may rewrite RTX with `stream.un_rtx(&mut header, &mut data, pt)`. both mutate the decrypted payload before `RtpPacket` is made

I think you're right it should be Arc<[u8]>, but it would need bigger changes: padding removal and RTX rewriting would need a refactor into slice/range, or SRTP decrypt would need to make the final shared representation after.
Vec -> Arc<[u8]> copies with Arc because Arc needs room for its refcount sowe can't just convert for free

I was thinking about making a follow up PR for the bigger changes beyond the API layer (upstream like mentioned above, and also downstream with even less copies at the encryption step) but If you want to avoid that reallocation now I can add commit that does the receiver side

xnorpx:
Can you present some benchmark numbers how much it is improved on some workload?

k0nserv:
I feel like it's desirable, but even `Arc<[u8]>` might be too restrictive as a type. Three motivating examples:

1. A user might want to statically allocate all their memory at process launch and then deal with `&[u8]`'s borrowed from that static allocation
2. A user might want to use `Bytes`
3. A user using a thread per core approach wants `Rc<[u8]>` to avoid needlessly paying the cost of atomic reference counting when no threading is involved

Potentially controversial idea: `pub struct Rtc<BufferKind> { .. }` or even

```rust
pub trait Meta {
  type OwnedBuffer: OwnedBuffer;
}

pub trait OwnedBuffer: Into<Vec<u8>> {}
impl Ownedbuffer for Vec<u8> {}

pub struct DefaultMeta;

impl Meta for DefaultMeta {
  type Ownedbuffer = Vec<u8>;
}

pub struct Rtc<M: Meta = DefaultMeta> { .. }
```

ThanhDodeurOdoo:
Hello @xnorpx @k0nserv thanks for the interest in the change, 

@k0nserv I didnt want to make the changes too heavy but I think that would be a very useful API if you are willing to take the added complexity

one detail, instead of 
```rs
pub trait OwnedBuffer: Into<Vec<u8>> {}
```
I think maybe something like
```rs
pub trait OwnedBuffer: Clone + AsRef<[u8]> + 'static {
    fn from_vec(data: Vec<u8>) -> Self;
}
```
it would be closer to the a "less copies" goal

@xnorpx yes so I used this situations, with `aws-lc-rs`

1. feed 1 encrypted inbound RTP datagram into str0m with `Input::Receive`
2. parse RTP, decrypt SRTP
3. emit `Event::RtpPacket`
4. fan out th RTP payload to 1, 2, 10  and 50 RTP senders
5. call `StreamTx::write_rtp` for each destination
6. `Input::Timeout` poll `Output::Transmit`
7. serialize RTP and encrypt outbound SRTP

for the cpu/clock It's probably realistically around 1-2%, so it's usually noise-level compared to the encryption and decryption cost, I will try to get a table from a stable machine with decent sample size

memory (64 inbound packets)

| payload | users | baseline vec | arc + rx | saved |
|---|---:|---:|---:|---:|
| 160 B | 1 | 134 KiB / 296 allocs | 125 KiB / 232 allocs | ~7% |
| 160 B | 2 | 208 KiB / 438 allocs | 189 KiB / 310 allocs | ~9% |
| 160 B | 10 | 805 KiB / 1574 allocs | 706 KiB / 934 allocs | ~12% |
| 160 B | 50 | 3.70 MiB / 7254 allocs | 3.21 MiB / 4054 allocs | ~13% |
| 1200 B | 1 | 329 KiB / 296 allocs | 255 KiB / 232 allocs | ~23% |
| 1200 B | 2 | 533 KiB / 438 allocs | 384 KiB / 310 allocs | ~28% |
| 1200 B | 10 | 2.12 MiB / 1574 allocs | 1.39 MiB / 934 allocs | ~35% |
| 1200 B | 50 | 10.11 MiB / 7254 allocs | 6.45 MiB / 4054 allocs | ~36% |
| 1350 B | 1 | 358 KiB / 296 allocs | 274 KiB / 232 allocs | ~23% |
| 1350 B | 2 | 582 KiB / 438 allocs | 414 KiB / 310 allocs | ~29% |
| 1350 B | 10 | 2.32 MiB / 1574 allocs | 1.49 MiB / 934 allocs | ~36% |
| 1350 B | 50 | 11.07 MiB / 7254 allocs | 6.95 MiB / 4054 allocs | ~37% |

xnorpx:
@ThanhDodeurOdoo are you using custom allocator such as mimalloc or jemalloc on those benchmarks? If not can you add one with that?

We are running single core servers (SYNC BABY) on one thread, so we are mostly passing reference around and then clone before passing to str0m. But if we can get Rc and save some cycles that would be nice.

ThanhDodeurOdoo:
@xnorpx sure, i was using the default allocator, I will run benchmarks using jemalloc (I remember that algesten mentioned that it's the one you're using), in the meantime I did run some more benchmarks with criterion and valgrind:



### Criterion
1200 inbound packets per run, sample size 80

| payload | users | baseline vec | arc + rx | gain |
|---|---:|---:|---:|---:|
| 160 B | 1 | 2.60 ms | 2.62 ms | ~-1% |
| 160 B | 30 | 43.61 ms | 43.07 ms | ~1% |
| 1350 B | 1 | 3.16 ms | 3.19 ms | ~-1% |
| 1350 B | 30 | 53.82 ms | 51.81 ms | ~4% |

(the performance gain is there but it's a bit lost in the noise)

### Valgrind (IR / cpu instruction requests)
1200 inbound packets

| path | payload | users | baseline vec | arc + rx | gain |
|---|---:|---:|---:|---:|---:|
| RTP event fanout | 160 B | 1 | ~1.47M | ~1.44M | ~2% |
| RTP event fanout | 160 B | 30 | ~3.68M | ~1.91M | ~48% |
| RTP event fanout | 1350 B | 1 | ~1.94M | ~1.90M | ~2% |
| RTP event fanout | 1350 B | 30 | ~5.34M | ~2.36M | ~56% |
| full relay | 160 B | 1 | ~3.02M | ~2.93M | ~3% |
| full relay | 160 B | 30 | ~47.92M | ~46.31M | ~3% |
| full relay | 1350 B | 1 | ~3.86M | ~3.79M | ~2% |
| full relay | 1350 B | 30 | ~63.27M | ~60.07M | ~5% |

ThanhDodeurOdoo:
@xnorpx  Valgrind does indeed show that jemalloc is a better allocator here, still noticeable gains but a bit smaller

(1200 inbound packets)

| path | payload | users | baseline vec | arc | gain |
|---|---:|---:|---:|---:|---:|
| RTP event fanout | 160 B | 1 | ~1.41M | ~1.41M | ~0% |
| RTP event fanout | 160 B | 30 | ~2.44M | ~1.88M | ~23% |
| RTP event fanout | 1350 B | 1 | ~1.85M | ~1.84M | ~1% |
| RTP event fanout | 1350 B | 30 | ~4.23M | ~2.31M | ~45% |
| full relay | 160 B | 1 | ~2.85M | ~2.83M | ~1% |
| full relay | 160 B | 30 | ~46.00M | ~45.27M | ~2% |
| full relay | 1350 B | 1 | ~3.78M | ~3.74M | ~1% |
| full relay | 1350 B | 30 | ~60.07M | ~58.81M | ~2% |

detail for 1350B / 30 users

| type | vec : arc | diff | 
|---|---:|---:|
|  Instructions |                60074611 : 58814864   |   (-2.09697%) [-1.02142x] |
|  L1 Hits |                    83008308 : 81166845    |     (-2.21841%) [-1.02269x] |
|  LL Hits |                        2898982 : 2703622  | (-6.73892%) [-1.07226x] |

for the fanout usecase, the small speed gain and the lower cache occupancy is pretty nice for a relatively shallow change and it opens the door for more gains with deeper changes

Where I work we have (albeit rare) uses cases for 150+ user fanouts (it may be spread over several cores but the less we need to split the better)

k0nserv:
> @k0nserv I didnt want to make the changes too heavy but I think that would be a very useful API if you are willing to take the added complexity

Appreciate it! My thinking was that changing to `Arc<[u8]>` is a breaking change, whereas I think using the generic approach is both more flexible and, because of the default, is non-breaking. If we are going to do something like this, better to get it right in the first instance and make it as flexible as possible I think.


> one detail, instead of[..]

Yeah, I hadn't really thought too deeply about the exact trait bounds. `Meta` would probably need a few different associated types to make that approach work well.

k0nserv:
Had Claude make a PoC https://github.com/algesten/str0m/compare/meta-type?expand=1. It's only the input half though

ThanhDodeurOdoo:
@k0nserv looks promising, if you add the RTP payload to meta it would cover my use case

algesten:
@k0nserv i'm not convinced the types we are building for with such a mechanism.

> * A user might want to statically allocate all their memory at process launch and then deal with `&[u8]`'s borrowed from that static allocation

While it would be nice, I don't think it be possible to remove allocations from inside str0m. Passing in some allocator or something, maybe. It's a stretch.

> * A user might want to use `Bytes`

Yeah, and some don't :)

While I'm not against supporting it, I'm not prepared to bend over backwards for it.

> * A user using a thread per core approach wants `Rc<[u8]>` to avoid needlessly paying the cost of atomic reference counting when no threading is involved

I think sharing data on the same thread with multiple Rtc instances would be a quite weird scenario. The main work in str0m is crypto, so you are effectively data sharing between multiple instances doing crypto on the same thread. The gain `Rc` does over `Arc` will be dwarfed by the cost of the crypto.

-----

To me it seemed `Arc<[u8]>` is a quite neat type. The one bit that suck about it is that going from an idiomatic `Vec<u8>` to `Arc<[u8]>` forces an allocation. So `Arc<Vec<u8>>` maybe is the "best" built-in type that avoids most allocations - but I feel a bit irritated by looking at that type signature.

k0nserv:
All of those were just examples, my point is just that there's benefit in being flexible because there are many potential strategies. There's a bunch of extra ceremony in str0m to achieve that though.

Too bad the Allocator API isn't stable yet, would be nice if all of str0m was generic over that.

> To me it seemed Arc<[u8]> is a quite neat type. The one bit that suck about it is that going from an idiomatic Vec<u8> to Arc<[u8]> forces an allocation. So Arc<Vec<u8>> maybe is the "best" built-in type that avoids most allocations - but I feel a bit irritated by looking at that type signature.

`bytes` does solve this nicely :)

algesten:
@ThanhDodeurOdoo if we land this PR as is, does that solve your problems?

ThanhDodeurOdoo:
@algesten yes the current shape is already good, it solves the copies on the API user side

algesten:
@k0nserv you ok with us merging it like this for now?

k0nserv:
Yes! No intent to block on my part, just keen we consider all the options since the change will be breaking

algesten (approved):

PR #963, opened 2026-05-23
RTP write API

ThanhDodeurOdoo:
consists of:

- [REF] rtp: replace direct write overloads: that adds a builder to the RTP write API (breaking change: `rtp_write` now takes a `RtpWrite` builder as param)
- [IMP] rtp: support VP8 descriptor patches: that adds `vp8_patch` to provide specific rewrites allowing us to leverage the optimisation brought by the recent refactor towards Arc<[u8]> packets
- [FIX] packet: reject malformed VP8 TL0PICIDX see ( https://github.com/algesten/str0m/pull/963#issuecomment-4543161315 ) (breaking change: added `ErrVP8CorruptedPacket` to the exhaustive (now non-exhaustive) public `PacketError` enum)

Usage example:

```rs
// usually done once per payload, rx side
// errors for malformed descriptor
let descriptor = Vp8Descriptor::parse(packet.payload.as_ref())?;
```

```rs
// can generate multiple patches, usually one per receiver, tx side
// errors for invalid values given
let patch = descriptor.patch()
    .with_picture_id(new_picture_id)
    .with_tl0picidx(new_tl0picidx)
    .with_keyidx(new_keyidx)
    .build()?;
```

```rs
// still infallible as the patch was pre-validated on build
stream_tx.write_rtp(
    RtpWrite::new(pt, seq_no, timestamp, wallclock, payload)
        .marker(marker)
        .ext_vals(ext_vals)
        .nackable(nackable)
        .csrc(csrc)
        .vp8_patch(patch),
);
```

benchmarks for commit 2:

steps:
1. feed one encrypted inbound RTP datagram into str0m with `Input::Receive`
2. parse RTP and unprotect/decrypt SRTP
3. drain `poll_output()` until `Event::RtpPacket`
4. fan out that RTP payload to each destination
5. rewrite VP8 PictureID either by full payload copy then mutate or by payload patch
6. call `StreamTx::write_rtp` or `write_rtp_with_payload_patches`
7. call `Input::Timeout`
8. drain `poll_output()` until `Output::Transmit`
9. serialize outbound RTP and protect/encrypt SRTP

Conditions:
packet: VP8-like RTP payload, 1,200 bytes  
rewrite: 2-byte VP8 PictureID at payload offset 2  
runs: 16 inbound RTP packets  
fanout: 1 user and 30 users  
allocator: `jemalloc`  
tool: `Valgrind`

**Results**
| workload | variant | IR | estimated cycles |
|---|---:|---:|---:|
| 1 user | copy then `write_rtp` | 489,759 | 798,695 |
| 1 user | payload patches | 484,535 | 788,514 |
| 1 user | gain | 1.07% | 1.27% |
| 30 users | copy then `write_rtp` | 8,584,193 | 13,841,546 |
| 30 users | payload patches | 8,393,559 | 13,479,493 |
| 30 users | gain | 2.22% | 2.62% |

(these are the same gains as the basic case of moving from `Vec<u8>` to `Arc<[u8]>`, since they basically just allow to do that even when we want to modify the payload)


src/streams/payload_patch.rs
```diff
@@ -0,0 +1,320 @@
...
+/// Inputs are `(payload_offset, replacement_bytes)` pairs. Offsets are
+/// relative to the RTP payload, not the RTP header.
+///
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub(crate) struct PayloadPatches {
+    storage: PayloadPatchStorage,
+}
+
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+enum PayloadPatchStorage {
```
ThanhDodeurOdoo: maybe the file is a bit "too much" and we can have a smaller diff by just letting the payloadwrite hold the ArrayVec of patches directly, but the enum allows to save some space and the noise of the feature is contained in its own file

src/streams/payload_patch.rs
```diff
@@ -0,0 +1,320 @@
...
+
+const MAX_INLINE_PAYLOAD_PATCHES: usize = 2;
+
+/// Maximum replacement byte length for one payload patch.
+pub const MAX_PAYLOAD_PATCH_BYTES: usize = 8;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct PayloadPatch {
+    offset: usize,
+    replacement_bytes: ArrayVec<u8, MAX_PAYLOAD_PATCH_BYTES>,
```
algesten: I wasn't quite expecting an opaque change of the RTP data. Since we got full RTP knowledge in str0m. What is the use case?

src/streams/payload_patch.rs
```diff
@@ -0,0 +1,304 @@
...
+        }
+
+        dst[copied_until..].copy_from_slice(&base_payload[copied_until..]);
+    }
+}
+
+/// Errors returned when RTP payload patches cannot be represented safely.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum PayloadPatchError {
```
ThanhDodeurOdoo: maybe this should be panics instead

src/streams/payload_patch.rs
ThanhDodeurOdoo: I thought it could be used in a generic way but for me a more vp8 aware API is fine too if you prefer that direction

my use case is for VP8 PictureID or TL0PICIDX which are in the payload descriptor

algesten: Yeah, I think so. After looking into it, you'd not be the only that wants to rewrite this and we also have a parser for VP8 already.

algesten:
@ThanhDodeurOdoo this is ready for review?

ThanhDodeurOdoo:
@algesten yes I think it's ready now, the patch building is still a separate, it feels more ergonomic to have the Result separate, but if you prefer to have it integrated to the `RtpWrite` it's not a big deal for me either way

algesten:
I took a pass over the API and VP8 patch path. Overall the `RtpWrite` builder shape looks cleaner, and carrying the patch on `RtpPacket` makes the regular-send/RTX behavior line up nicely.

A few things I think are worth tightening before merge:

1. Public rewrite inputs probably should not panic. `Vp8PatchBuilder::picture_id()` and `key_idx()` currently assert on out-of-range values, while `build()` already has a `Result`-returning validation boundary. For SFU-style rewrite inputs, it would be friendlier to report these as `Vp8PatchError` variants from `build()` instead of panicking.

2. 7-bit PictureID rewrites currently silently truncate. If the parsed descriptor is using the 7-bit representation and the caller asks for `0x80`, the patch writes `0x00`. Since the patch cannot grow the descriptor to 15-bit while preserving payload length, I think this should either return an error for `!is_15_bit && picture_id > 0x7f`, or the modulo/truncation behavior should be explicit in the API docs. Given the “validated patch” framing, I would prefer an error.

3. Minor docs typos in `Vp8PatchBuilder`: “pasred” and “repalced”.

algesten:
Oh, and add a row to CHANGELOG.md

algesten:
Right. I checked RFC 7741 and `L=1, T=0` is invalid.

Section 4.2 says `L` means `TL0PICIDX` is present, and if `L` is set then `T` MUST also be set. So this check looks right to me:

```rust
if has_tl0_pic_idx && !has_tid {
    return Err(Vp8DescriptorError::MalformedDescriptor);
}
```

This also means the old depacketizer is a bit too permissive here.

https://datatracker.ietf.org/doc/html/rfc7741#section-4.2

algesten (approved):

algesten:
@xnorpx are you cool with this API change?

xnorpx:
Can you eli5 why the patches is needed?

algesten:
> Can you eli5 why the patches is needed?

I can now! :)

If you're doing simulcast over VP8, it's not enough to just fiddle with the RID headers on the RTP layer. You also need to rewrite some track/layer index in the VP8 payload itself.

ThanhDodeurOdoo:
@xnorpx so it's conceptually a follow up to https://github.com/algesten/str0m/pull/959 (the Arc<[u8]> payloads),

Most of the time we don't need to touch the content of the payload so we can pass the arc and avoid rewriting the payload, because each destination can get the same identical payload.

But for vp8 (and vp9 eventually), we sometimes want to edit the descriptor (part of the payload) because it can contain information that depends on the destination.

The main use case is for simulcast: we can have multiple different origin streams (each with their own distinct descriptor "continuity") but a "consumer" that only sees 1 stream. 

That 1 destination stream has a single identity and must have a coherent descriptor continuity while potentially swapping between the different simulcast sources (which don't have correlated descriptors (or at least nothing guarantees it)), so we can't just copy the descriptor of one source then copy the descritor of the next source after a swap because that would break the continuity of the destination descriptor, so what we want to do is to rewrite them so that it is consistent regardless of the payload origin.

The PR gives a relatively intuitive way of manipulating that (it used to be up to the API user to reconstruct the payload) and also performant because it allows us to not touch the Arc<[u8]> and let str0m take care of it downstream.

xnorpx:
I see, that makes sense then. I am all good.

xnorpx (approved):

algesten:
Thanks!

PR #964, opened 2026-05-26
fix: abs-capture-time dropped on receiver

lherman-cs:
The motivation is to use this abs-capture-time to measure the end-to-end latency from the publisher -> sfu -> subscriber. In my case, all of these use str0m. But the publisher and subscriber run on the default mode (frame mode). 

Changes:

1. Add a public API for setting the frame's absolute capture time
2. Fix the missing abs-capture-time on the receiver
    * This is a breaking change because we need to merge the first packet's value. On the sender side, we only include abs-capture-time in the first packet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-  * Add `abs-capture-time` writer support and fixed receiver-side extension propagation (breaking)
+  * Add `abs-capture-time` writer support and fixed receiver-side extension propagation (breaking) #964
   * Expose ICE candidate pair RTT in CandidatePairStats #962
   * Add stereo to opus format params #955
   * Use `Arc<[u8]>` as input/output of media data (breaking) #959
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Add `abs-capture-time` writer support and fixed receiver-side extension propagation (breaking)
   * Expose ICE candidate pair RTT in CandidatePairStats #962
   * Add stereo to opus format params #955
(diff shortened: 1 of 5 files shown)


algesten:
Wonder if we should also preserve VLA here.

`payload.rs` strips both first-packet-only values on non-first packets:

```rust
if !first {
    pkt_ext_vals.abs_capture_time = None;
    pkt_ext_vals.user_values.remove::<VideoLayersAllocation>();
}
```

This PR restores `abs_capture_time` from the first packet when creating `MediaData.ext_vals`, but leaves `VideoLayersAllocation` behind for multi-packet frames.

Maybe extend the merge to copy VLA too:

```rust
if let Some(first_val) = first.user_values.get_arc::<VideoLayersAllocation>() {
    merged.user_values.set_arc(first_val);
}
```

And make `vla_frame_mode` use a payload large enough to force more than one RTP packet. Right now it uses 10 bytes, so it does not cover the split case.

Lil nit: `Writer::abs_capture_time` says packet, but `Writer` is frame-level API. Should say frame.

algesten:
Oh, and not sure I'd call this breaking. It's breaking-ish on observed behavior. Doesn't matter hugely since we got breaking changes cued up.

lherman-cs:
I was wondering about the VLA field too. It makes sense to include it after all. The vla_frame_mode test with a big frame reproduced the missing vla before the fix. 

I also wonder about other clients, whether it makes sense to even iterate from the last packet to the first. But this creates non-trivial work for a big frame.

It's not big. But the old code will not compile, so "breaking" it is 🙂. Btw, I'm referring to the `ext_vals` return value to be no longer a reference.

algesten (approved):

algesten:
Thanks!

PR #967, opened 2026-05-29
Make MTU size configurable

tyhik:
Make MTU configurable via RtcConfig .  Tested with Dimpl+aws-lc-rs, Dimpl+CNG, and Schannel DTLS backends so far. 

As of now, not deliverable; depends on sctp-proto > 0.9.0 .

### DIFF
--- a/crypto/openssl/src/dtls_ossl.rs
+++ b/crypto/openssl/src/dtls_ossl.rs
@@ -12,10 +12,10 @@ use openssl::ssl::{SslContext, SslContextBuilder, SslMethod};
 use openssl::ssl::{SslOptions, SslStream, SslVerifyMode};
 use openssl::x509::X509;
 
+use str0m_proto::DATAGRAM_MTU_TARGET;
 use str0m_proto::crypto::dtls::{DtlsCert, KeyingMaterial, SrtpProfile};
 use str0m_proto::crypto::dtls::{DtlsImplError, DtlsInstance, DtlsOutput, DtlsProvider};
 use str0m_proto::crypto::{CryptoError, DtlsVersion};
-use str0m_proto::{DATAGRAM_MTU, DATAGRAM_MTU_WARN};
 
 // ============================================================================
 // IO Buffer
@@ -290,9 +290,9 @@ struct OsslDtlsImpl {
 }
 
 impl OsslDtlsImpl {
-    fn new(cert: &DtlsCert) -> Result<Self, CryptoError> {
+    fn new(cert: &DtlsCert, mtu: usize) -> Result<Self, CryptoError> {
         let context = dtls_create_ctx(cert)?;
-        let ssl = dtls_ssl_create(&context)?;
+        let ssl = dtls_ssl_create(&context, mtu)?;
         Ok(OsslDtlsImpl {
             tls: TlsStream::new(ssl, IoBuffer::default()),
         })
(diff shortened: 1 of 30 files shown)


crypto/openssl/src/dtls_ossl.rs
```diff
@@ -347,8 +349,8 @@ impl OsslDtlsImpl {
     fn poll_datagram(&mut self) -> Option<Vec<u8>> {
         let x = self.tls.inner_mut().pop_outgoing();
         if let Some(x) = &x {
-            if x.len() > DATAGRAM_MTU_WARN {
-                warn!("DTLS above MTU {}: {}", DATAGRAM_MTU_WARN, x.len());
+            if x.len() > self.mtu {
+                warn!("DTLS above MTU {}: {}", self.mtu, x.len());
```
algesten: I think we have a small logic shift here that needs to be ironed out.

```rs
const DATAGRAM_MTU: usize = 1200;
const DATAGRAM_MTU_WARN: usize = 1280;
```

We have these two values. In some cases we can calculate to hit the MTU perfectly (RTP), and in others we can't (DTLS). The idea is that we aim for a certain MTU (1200), but warn if don't manage to this it - with some flex. In some of these subsystems, you ask for size X, but it has some extra headers etc, so you don't get a clamped X but `X + <extra>`.

Now that extra is OK, but we also need to set some rather conservative limits for warnings. I think the 1280 here is a matgic number lifted fro libWebRTC (double check that). And that is further informed by VPN tunnels where the outer MTU might be 1500, but by the time you can send media data, it's more like 1280 left.

What if the setting you can configure is a range `1200..1280` - which means we aim for 1200, we allow any value from 1200-1280, and warn if above 1280?

tyhik: Okay, sure, I'll bring the warn logic back. I may have underestimated the overhead of the protocol headers.

You mentioned 1200 as the target MTU.  Currently in main, DATAGRAM_MTU is 1150 (crates/proto/src/lib.rs) and _WARN is 1280, so there's 130 bytes cushion.  dtls_schannel.rs has its own local DATAGRAM_MTU=1200, but my PR removes it and switches to the common one.  I'd prefer leaving the defaults (target 1150 and the warning cushion 130) unmodified in this PR.

Do I read your 1200..1280 example as proposal to use Range<usize> for the setter?  Any preference of the names/signatures? set_mtu(), set_mtu_range(), anything else ?  For usage convenience, mtu() and mtu_warn() returning lower and upper end, respectively, would be convenient getters.

algesten: Yeah, some of these MTU settings have been a bit trial and error. It could be that depending on subsystem you need to have offsets in negative direction to hit the wanted value.

The warning has served as canary that I'm not hitting the values I'm aiming for.

I was thinking a range, but it could be a struct too. I just don't want two independent settings on RtcConfig, since they will always be related.

tyhik: Updated the PR. Please review.

Main changes this time:
1. Warn logic restored.
2. The warning previously in crypto/openssl/src/dtls_ossl.rs was removed and replaced with the warning in src/dtls.rs to cover the packets of all the endpoints.
3. RtcConfig.mtu option was made a range, and so were the other structs that use MTU warning like Dtls, Session, etc. 

Re. 3, MTU carrying members are now everywhere named mtu, regardless of being Range or usize. I think it would be clearer to rename the .mtu members that contain range rather as .mtu_range . Not sure. Just say if you'd like them renamed.

Also thinking now that to be consistent with mtu as a range, perhaps instead of the two defaults DATAGRAM_MTU_TARGET and DATAGRAM_MTU_WARN there should be a single DATAGRAM_MTU as a range. Just let me know if you want that.

algesten (approved):

algesten:
Looks good! Thanks!

PR #969, opened 2026-06-03
fix: accept any session name in SDP, not just "s=-"

mstyura:
The FFMpeg WHIP does generate custom session name not equal to `s=-`. The custom names are allowed by spec. I think it make sense to accepts such `sdp`s in `str0m`.

https://datatracker.ietf.org/doc/html/rfc4566#section-5.3
```
5.3 Session Name ("s=")

      s=<session name>

   The "s=" field is the textual session name.  There MUST be one and
   only one "s=" field per session description.  The "s=" field MUST NOT
   be empty and SHOULD contain ISO 10646 characters (but see also the
   "a=charset" attribute).  If a session has no meaningful name, the
   value "s= " SHOULD be used (i.e., a single space as the session
   name).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Accept any non-empty SDP session name (`s=`), not just `s=-` (RFC 8866 section 5.3)
+
 # 0.20.0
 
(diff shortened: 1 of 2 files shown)


algesten (approved):
Looks good! Thanks!

PR #973, opened 2026-06-12
Add optional drv feature deriving drv::Input on identity types

lolgesten:
## Summary

Adds a default-off `drv` feature that derives [`drv::Input`](https://docs.rs/drv) — drv's memoization cache-key plumbing — on str0m's public identity types, so downstream `#[drv::memo]` query functions can take them by value (or project them in a `drv::Input` struct) as cache keys.

Without it, a drv-based consumer can't satisfy drv's `ToStatic` bound on these types: the orphan rule blocks a consumer-side `impl drv::ToStatic for str0m::media::Mid`, forcing local newtypes (with conversions at every boundary) or pre-collapsing values into primitives.

Types covered:

- **rtp ids:** `Mid`, `Rid`, `Ssrc`, `Pt`, `SessionId`, `SeqNo`
- **enums:** `Direction`, `MediaKind`, `Codec`, `VideoOrientation`
- **format:** `Frequency`, `CodecSpec`, `FormatParams`

`Frequency(NonZeroU32)` relies on drv 0.4.3's new `NonZero*` `ToStatic` impls, hence `drv = { version = "0.4.3", optional = true }`.

The feature is **off by default**, gated with `#[cfg_attr(feature = "drv", derive(drv::Input))]`, so non-drv consumers compile exactly as before and pay no extra dependency. A feature-gated test (`tests/drv_input.rs`) confirms the types work as `#[drv::memo]` inputs.

### Out of scope

- `PayloadParams` — its hand-written `PartialEq` deliberately ignores `locked`/`fb_remb`; drv's field-by-field `eq_static` would diverge from that equality, so it's left out.
- `Extension` — its `UnknownUri(String, Arc<dyn ExtensionSerializer>)` variant holds a trait object, which isn't `ToStatic`.
- `MediaTime`, `CodecExtra`, and the H.265 payload/NALU types — value-/payload-carrying, not identity.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -296,6 +296,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "drv"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cb67531e2bd07035971a7b45c7ca013024e6691a348736801675914847f03b21"
+dependencies = [
+ "drv-macros",
+]
+
+[[package]]
+name = "drv-macros"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4d1fa8647349fd4f292c87823cc96987a404b38181da2736ce7773d9d5601298"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.90",
+]
+
+[[package]]
 name = "errno"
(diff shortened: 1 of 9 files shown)


lolgesten:
Retargeting: this branch is based on lolgesten/str0m main; opening against that base instead of upstream algesten/str0m (which is 345 commits ahead and conflicts).

PR #976, opened 2026-06-15
Allow sctp outgoing buffer limit to be configured

wdouglass:
I'm continuing to search for ways to increase datachannel throughput for my application, which pairs a 1080p h264 stream with an extra ~256k of data for each frame on a datachannel. Increasing this outgoing buffer size to 512k achieved some extra throughput. Let me know what you think.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,10 +2,9 @@ use std::ops::RangeInclusive;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use crate::Rtc;
 use crate::config::DtlsCert;
-use crate::crypto::CryptoProvider;
 use crate::crypto::dtls::DtlsVersion;
+use crate::crypto::CryptoProvider;
 use crate::format::CodecConfig;
 use crate::format::Vp9PacketizerMode;
 use crate::ice::IceCreds;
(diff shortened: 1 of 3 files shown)


algesten:
@wdouglass I seem to recall that there are some limitations/interactions with `sctp-proto` here that were not obvious, meaning the net-effect of how to document this as public API was not straightforward.

Also. I don't understand how this could increase throughput since MTU is MTU, and larger payloads are needing to be fragmented into packets anyway. In some respects it's treating str0m as some middle-storage – something we've said we should not do.

I'm not challenging that you experience more throughput, but we need to get to the bottom of why you see that.

wdouglass:
Fair enough, I think I understand why -- I'll explain. My sensor generates about 256k of extra data per video frame, and I have it set such that if the data channel kicks back a packet it skips processing the next frame. Because my data is so bursty, increasing that buffer size (from it's default 128k) let's me keep a full frames worth do data queued up for when the first part gets acknowledged by the receiver (which is a web assembly program running in Firefox)

wdouglass:
just noticed another aspect of your comment
```In some respects it's treating str0m as some middle-storage – something we've said we should not do.```

you're right, and feel free to reject this PR on these grounds. i have a lot of raw headroom on the wire (currently sending about 50 mbps on a gigabit connection), i suspect my client isn't acking the sctp packets as fast as I'd like

algesten:
Parking this until we can answer the question:

Why would increasing this buffer improve str0m throughput when MTU is what decides the sctp-proto chunking?

PR #980, opened 2026-06-19
feat: support NACK retransmission without RTX via in-band resend on main SSRC

mstyura:
I'd like to propose to decouple `NACK` from `RTX`. This PR gates `NACK` on negotiated `nack` feedback rather than the presence of `RTX`. When no `RTX` `SSRC` is configured, retransmissions are resent in-band on the main `SSRC` (original PT + sequence number, as `libwebrtc` [does](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/modules/rtp_rtcp/source/rtp_sender.cc;l=299-300;drc=888fb5647ce3fe5740a33500eaf731ec8c6b1b0d)).

The main motivation: `str0m` can now NACK-retransmit `Opus` packets, which previously had no loss recovery since audio never gets `RTX`. In practice this is audible - a long sustained sound (e.g. "AAAAAAA" for 5 seconds) recorded over a lossy network (`5%`) now plays back without the audible glitches that the dropped packets used to cause.

Also adds `PayloadParams::set_resend` to configure/disable RTX.

image

### DIFF
--- a/src/format/payload_params.rs
+++ b/src/format/payload_params.rs
@@ -208,6 +208,12 @@ impl PayloadParams {
         self.resend
     }
 
+    /// Sets the RTX resend payload type. Use `None` to disable RTX, in which
+    /// case retransmissions (if NACK is enabled) are sent in-band on the main SSRC.
+    pub fn set_resend(&mut self, resend: Option<Pt>) {
+        self.resend = resend;
+    }
+
     /// The codec with settings for this group of parameters.
     pub fn spec(&self) -> CodecSpec {
         self.spec
@@ -578,6 +584,14 @@ impl PayloadParams {
         let mut remote_pt = first.pt;
         let mut remote_rtx = first.resend;
 
+        // The RTX payload type (RFC 4588 §8.6) is a media format, and for unicast
+        // the answerer selects formats (RFC 3264 §6.1, Unicast Streams) "from
+        // amongst those listed in the offer". So if we didn't offer RTX for this
+        // codec, don't adopt the remote's RTX PT.
+        if self.resend.is_none() {
+            remote_rtx = None;
+        }
+
         if self.locked {
             // This can happen if the incoming PTs are suggestions (send-direction) rather than demanded
             // (receive-direction). We only want to warn if we get receive direction changes.
@@ -645,6 +659,13 @@ impl PayloadParams {
             if let Some(rtx) = remote_rtx {
                 claimed.assert_claim_once(rtx);
             }
+
+            // The answerer removes rtcp-fb attributes it does not support, and per
+            // RFC 4585 §4.2 "both offerer and answerer MUST only use feedback
+            // mechanisms negotiated in this way". So intersect our local `fb_nack`
+            // with what the remote signaled (`a=rtcp-fb <pt> nack`). Runs once at
+            // lock time, when `self.fb_nack` still holds the local desired value.
+            self.fb_nack = self.fb_nack && first.fb_nack;
 
             // This is now locked.
             self.locked = true;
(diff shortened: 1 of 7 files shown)


algesten:
@mstyura not necessarily arguing against this, but why wouldn't you use RTX also for the Opus case?

RTX is a cryptographic safety to avoid an evil MITM forcing us to repeat the same data over the main SRTP channel. Seems like you want to enable it, no?

algesten:
I guess the answer is: libWebRTC doesn't do that.

src/format/payload_params.rs
```diff
@@ -208,6 +208,12 @@ impl PayloadParams {
         self.resend
     }
 
+    /// Sets the RTX resend payload type. Use `None` to disable RTX, in which
+    /// case retransmissions (if NACK is enabled) are sent in-band on the main SSRC.
+    pub fn set_resend(&mut self, resend: Option<Pt>) {
+        self.resend = resend;
+    }
+
```
algesten, line 216: `pt` and `resend` are set on construction and immutable. There deliberately is no setter for `resend`.

src/streams/send.rs
```diff
@@ -505,12 +505,10 @@ impl StreamTx {
...
-                } else {
-                    // If the PT we're sending on doesn't have a corresponding RTX PT,
-                    // the packet is de-facto not nackable.
-                    //
-                    // This blocks incoming NACK requests and thus ensures there are no
-                    // entries in self.retries without a RTX PT.
+                } else if !param.fb_nack() {
+                    // No RTX PT and no `fb_nack`: the packet cannot be resent (neither
+                    // via RTX nor in-band), so it is de-facto not nackable. This ensures
+                    // there are no entries in the rtx cache we can't resend.
```
algesten, line 511: This causes a confusion in the code base. "rtx" is in many places used to mean "can resend" interchangeably. I'm a bit frustrated with myself for doing that back in the day, because at this point this PR is working against language and patterns that are consistent across the code base.

I don't know what the solution is.

src/streams/send.rs
```diff
@@ -527,6 +525,13 @@ impl StreamTx {
 
                 header_ref.clone()
             }
+            NextPacketKind::ResendInband => {
+                // In-band resend on the main SSRC: reuse the original PT and sequence
+                // number so SRTP reproduces the original ROC/IV and the encrypted
+                // payload decrypts at the receiver.
+                header_ref.ext_vals.rid_repair = None;
+                header_ref.clone()
+            }
```
algesten, line 534: I'm uncertain about having a new enum value here. Whether we want just NextPacketKind::Resend, but discover the "mode" of whether we are operating with or without RTX.

However this has consequences for the BWE and Pacer, so again. I'm not certain.

algesten (commented):
I'm on the fence about this one.

I do think supporting Opus resends without RTX is a nice goal, but I'm not sure this PR goes all the way. In my mind, I've used the "rtx" meaning "can resend" interchangeably, which means to make the code base consistent in language and logic, there's a deeper take that is needed.

I would want to do that manually myself.

mstyura:
I have the following use-case of `str0m`. The `libwebrtc` clients (both native and browsers) connects to `str0m`-powered backend to send media. The media is then being processed with `ffmpeg-next`. Sometimes audio packets are lost in such a way that causes audible glitches during playback. `libwebrtc` only supports `nack` for audio, so I was thinking that `str0m` just not yet handle inband retransmits, rather than intentionally not implemented it.

algesten:
@mstyura yes. It surely is an omission.

xnorpx:
Note that RTX usage for audio is not so common due to re-order of packets and causing extra delay (if that matters) In bacnd fec or DRED in Opus is better in terms of latency.

PR #981, opened 2026-06-19
Support AppSpecific Feedback

sravanthivelidandla:


xnorpx:
@sravanthivelidandla  please add integration tests as examples

src/rtp/rtcp/psfbapp.rs
```diff
@@ -0,0 +1,157 @@
...
+            feedback_message_type: FeedbackMessageType::PayloadFeedback(
+                PayloadType::ApplicationLayer,
+            ),
+            words_less_one: (self.length_words() - 1) as u16,
+        }
+    }
+
+    fn length_words(&self) -> usize {
+        // 1 (RTCP header) + 2 (sender_ssrc + media_ssrc) + payload words
+        let payload_words = (self.payload.len() + 3) / 4;
```
pthatcher: Nit: Is this different than using div_ceil(4)?

src/rtp/rtcp/psfbapp.rs
```diff
@@ -0,0 +1,157 @@
...
+        1 + 2 + payload_words
+    }
+
+    fn write_to(&self, buf: &mut [u8]) -> usize {
+        self.header().write_to(&mut buf[..4]);
+        buf[4..8].copy_from_slice(&self.sender_ssrc.to_be_bytes());
+        buf[8..12].copy_from_slice(&self.media_ssrc.to_be_bytes());
+
+        let payload_len = self.payload.len();
+        buf[12..12 + payload_len].copy_from_slice(&self.payload);
```
pthatcher: Nit:  I always prefer it like this:  "buf[12..][..payload_len].copy_form_slice(&self.payload)"

src/rtp/rtcp/psfbapp.rs
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    fn length_words(&self) -> usize {
+        // 1 (RTCP header) + 2 (sender_ssrc + media_ssrc) + payload words
+        let payload_words = (self.payload.len() + 3) / 4;
+        1 + 2 + payload_words
+    }
+
+    fn write_to(&self, buf: &mut [u8]) -> usize {
```
pthatcher, line 40: Does this method just blindly assume that buf will be big enough?  Seems risky.

src/rtp/rtcp/psfbapp.rs
```diff
@@ -0,0 +1,157 @@
...
+        buf[4..8].copy_from_slice(&self.sender_ssrc.to_be_bytes());
+        buf[8..12].copy_from_slice(&self.media_ssrc.to_be_bytes());
+
+        let payload_len = self.payload.len();
+        buf[12..12 + payload_len].copy_from_slice(&self.payload);
+
+        // Pad to 4-byte boundary
+        let padded_payload_len = (payload_len + 3) / 4 * 4;
+        for b in &mut buf[12 + payload_len..12 + padded_payload_len] {
+            *b = 0;
```
pthatcher: Could you not use something like "buf[12 + paylad_len..][..pad_len].fill(0)"?

src/rtp/rtcp/psfbapp.rs
```diff
@@ -0,0 +1,157 @@
...
+    type Error = &'static str;
+
+    fn try_from(buf: &'a [u8]) -> Result<Self, Self::Error> {
+        if buf.len() < 8 {
+            return Err("AppSpecificFeedback less than 8 bytes");
+        }
+
+        let sender_ssrc = u32::from_be_bytes([buf[0], buf[1], buf[2], buf[3]]).into();
+        let media_ssrc = u32::from_be_bytes([buf[4], buf[5], buf[6], buf[7]]).into();
+        let payload = buf[8..].to_vec();
```
pthatcher: Nit: I'm always a fan of using split_at for stuff like this.

src/change/direct.rs
```diff
@@ -387,4 +387,21 @@ impl<'a> DirectApi<'a> {
...
+    /// with a Receiver Report for RFC 3550 compliance. The `sender_ssrc` is used as
+    /// the RR's sender SSRC, which allows remote RTCP demuxers that route compound
+    /// packets by the first SSRC to deliver the feedback to the correct media session.
+    ///
+    /// The `payload` is opaque and application-defined.
+    pub fn send_app_specific_feedback(
+        &mut self,
+        sender_ssrc: Ssrc,
+        media_ssrc: Ssrc,
+        payload: Vec<u8>,
```
algesten: Our new input datatype in other situations is this:

```rs
impl Into<Arc<[u8]>>
```

https://docs.rs/str0m/0.20.0/str0m/media/struct.Writer.html#method.write
https://docs.rs/str0m/0.20.0/str0m/rtp/struct.RtpWrite.html#method.new

I think we should leave it either `Into<Arc<[u8]>>` or `&[u8]` not `Vec`.

src/session.rs
```diff
@@ -113,8 +114,18 @@ pub(crate) struct Session {
     feedback_tx: VecDeque<Rtcp>,
     feedback_rx: VecDeque<Rtcp>,
 
+    // Standalone feedback queue — sent in their own SRTCP compound datagrams (not bundled
+    // with the regular SR/RR). This is needed when the remote RTCP demuxer routes compound
+    // packets based on the first SSRC; bundling feedback with an unrelated SR/RR would cause
+    // misrouting. Each entry is sent as a minimal compound packet (empty RR + feedback).
+    standalone_feedback_tx: VecDeque<Rtcp>,
```
algesten: That's already the case for TWCC. This should not require an intermediary storage like this. Write straight to `feedback_tx`.

src/session.rs
```diff
@@ -624,7 +645,19 @@ impl Session {
             }
         }
 
-        for fb in RtcpFb::from_rtcp(self.feedback_rx.drain(..)) {
+        // Extract application-specific feedback directly before stream routing.
+        // These are not stream-routed and are surfaced as events to the application.
+        let drained: Vec<Rtcp> = self.feedback_rx.drain(..).collect();
```
algesten: This is not ok. We cannot have allocation in this hot path.

src/session.rs
```diff
@@ -624,7 +645,19 @@ impl Session {
...
+        let mut for_rtcpfb = Vec::new();
```
algesten: This is not ok. We cannot have allocations here.

src/session.rs
```diff
@@ -113,8 +114,18 @@ pub(crate) struct Session {
...
+    // with the regular SR/RR). This is needed when the remote RTCP demuxer routes compound
+    // packets based on the first SSRC; bundling feedback with an unrelated SR/RR would cause
+    // misrouting. Each entry is sent as a minimal compound packet (empty RR + feedback).
+    standalone_feedback_tx: VecDeque<Rtcp>,
+
     raw_packets: Option<VecDeque<Box<RawPacket>>>,
 
+    // Pending application-specific feedback (PSFB FMT=15) messages to emit as events.
+    // Stored as (sender_ssrc, media_ssrc, payload) tuples.
+    pending_app_feedback: VecDeque<(Ssrc, Ssrc, Vec<u8>)>,
```
algesten: I dislike this. Queues and intermediary storage. Since it seems these packets cannot be stapled together with SR/RR (though I'm not entirely sure why). There should be max one of these per rtcp packet right?

If so, we don't need a queue, only a `Option<XXX>`

algesten (commented):
I think overall:

1. Construct the packet directly into `feedback_tx`
2. Consider if there really can be more than on of these.
3. No frivolous allocations.

src/change/direct.rs
sravanthivelidandla: updated to use Into<Arc<[u8]>>

src/lib.rs
```diff
@@ -1019,6 +1020,19 @@ pub enum Event {
...
+    /// Emitted when a non-REMB FMT=15 RTCP PSFB message is received. The payload
+    /// is opaque and application-defined (RFC 4585 Section 6.4).
+    AppSpecificFeedback {
+        /// SSRC of the sender of this feedback message.
+        sender_ssrc: rtp::Ssrc,
+        /// SSRC of the media source this feedback relates to.
+        media_ssrc: rtp::Ssrc,
+        /// Application-dependent payload.
+        payload: Arc<[u8]>,
+    },
```
algesten: Can we have a wrapper type `AppSpecificFeedback` (like we do `SenderFeedback`) instead of the raw fields on the enum.

src/session.rs
```diff
@@ -115,6 +115,10 @@ pub(crate) struct Session {
 
     raw_packets: Option<VecDeque<Box<RawPacket>>>,
 
+    // Pending application-specific feedback (PSFB FMT=15) messages to emit as events.
+    // Stored as (sender_ssrc, media_ssrc, payload) tuples.
+    pending_app_feedback: Option<(Ssrc, Ssrc, Arc<[u8]>)>,
```
algesten: And we use the wrapper type already here.

src/session.rs
```diff
@@ -326,6 +331,14 @@ impl Session {
         Some(())
     }
 
+    /// Enqueue an application-specific feedback message (PSFB FMT=15, PT=206) for
+    /// transmission as part of the regular RTCP compound packet.
+    pub(crate) fn send_app_specific_feedback(&mut self, sender_ssrc: Ssrc, media_ssrc: Ssrc, payload: impl Into<std::sync::Arc<[u8]>>) {
+        self.feedback_tx.push_back(Rtcp::AppSpecificFeedback(
+            crate::rtp_::AppSpecificFeedback { sender_ssrc, media_ssrc, payload: payload.into() }
```
algesten: Follow the precedence, and add a `use` statement instead of full path.

algesten (commented):
This looks good now! Almost ready to merge this now. Some tiny tweaks.

algesten (changes requested):
Hey @sravanthivelidandla @xnorpx,

I get the feeling my comments are going straight into Claude and the generated follow-up commit are not reviewed by a human. While it's fine you use LLMs, I refuse to drive code changes through your bots. You need to review your code before I do.

* Don't repeat problems I already pointed out (use "use" import statements, not full paths)
* Follow code precedence (why rename `AppSpecificFeedbackEvent` in the import, when my previous comment was clearly about following established pattern?)

sravanthivelidandla:
Hi @algesten , 
I am sorry for the previous push, I will definitely make sure that this is not repeated again and review my changes thoroughly before pushing them to the remote.

src/session.rs
```diff
@@ -326,6 +331,14 @@ impl Session {
         Some(())
     }
 
+    /// Enqueue an application-specific feedback message (PSFB FMT=15, PT=206) for
+    /// transmission as part of the regular RTCP compound packet.
+    pub(crate) fn send_app_specific_feedback(&mut self, sender_ssrc: Ssrc, media_ssrc: Ssrc, payload: impl Into<Arc<[u8]>>) {
+        use crate::rtp_::AppSpecificFeedback as RtcpAppFeedback;
```
algesten, line 342: 👍

algesten (approved):

algesten (approved):
Looks good!

PR #982, opened 2026-06-23
Add RFC 2198 RED (redundant audio) support for all audio codecs 

farit2000:
## What

Adds optional [RFC 2198](https://www.rfc-editor.org/rfc/rfc2198) RED (redundant audio) for all audio codecs (Opus, G722, PCMU, PCMA), off by default. Enable it per codec via the `use_red` argument of the audio enable fns (e.g. `RtcConfig::enable_opus(true, true)`). RED carries a redundant copy of one or more previous frames in each packet, letting the receiver recover isolated packet losses without retransmission, at the cost of extra audio payload size (roughly one primary's worth per redundancy level).

## Design

Modeled on the existing RTX support:

- A `Codec::Red` marker + `PayloadParams.red: Option<Pt>` linkage, folded onto the primary audio codec in SDP (`a=rtpmap:63 red/48000/2`, `a=fmtp:63 111/111`), mirroring how `apt` links RTX. Kept only when both peers offer it. Each audio codec gets its own RED PT, advertised with that codec's RTP clock rate (so G722's RED sits on 8000, its wire rate, not the 16000 it samples at).
- **Frame mode (default):** transparent — outgoing audio is wrapped with a configurable number of redundancy levels; incoming RED is unwrapped back to the primary codec and lost packets are recovered from later packets' redundancy via the existing depayload path.
- **RTP mode:** RED packets pass through unchanged; `RedEncoder` / `RedDecoder` are public for callers that want to build or parse RED themselves.
- A self-contained `packet::red` module does the byte-level RFC 2198 work and is fuzzed.

## Send-side control

RED availability is fixed at negotiation (or set up via DirectApi) and never renegotiated. On top of that:

- `RtcConfig::set_red_distances(&[u32])` sets the redundancy pattern — how many packets back each level carries. `[1]` (default) sends the previous frame, as browsers do; `[1, 3, 5]` sends three levels for higher loss. The pattern is sanitised (sorted, de-duped, capped at the receiver recovery depth) and always anchored on distance 1.
- `DirectApi::set_red_send(mid, bool)` toggles wrapping on and off at runtime with no renegotiation (reachable from an SDP-negotiated session too, via `rtc.direct_api()`), so an SFU can turn RED on for a leg that starts losing packets and off again when it's clean.
- If a redundant block would push the packet over the MTU, the deepest levels are shed so the wrapped packet still fits.

## Robustness

RED payloads are untrusted remote input, so the receive path is bounded:

- The decoder rejects payloads with more than 32 redundant blocks (matching libwebrtc's RED receiver limit; conformant Opus-RED senders use one, occasionally up to ~4).
- Recovery is limited to the most-recent `MAX_RED_RECOVERY_DEPTH` (8) packets and only to blocks carrying the primary PT, so one packet can't expand into unbounded recovery work, and a redundant block of a different codec (RFC 2198 permits mixed PTs) is never fed to the primary depayloader.
- A redundant block carries a timestamp offset, not a sequence number, so recovery places it by bracketing its time between the nearest received packets and filling the missing sequence number between them. This is exact for a single gap; for a multi-packet burst it assumes uniform frame duration (the fixed-ptime norm) and is otherwise best-effort. A recovered frame is only ever placed at a still-missing sequence number, so it never overwrites a correctly received frame. Recovered sequence numbers are marked so they are neither re-recovered nor NACKed, while reception reports still reflect the real wire loss.

## Tests

Unit (byte-level encode/decode round-trips, no-panic fuzz-style decode, block-count cap, send-side wrap/history/budget-shedding, and `red_locate_seq` placement across single-hole, uniform-burst, depth-limit and already-recovered cases) plus integration in `tests/red.rs`: transparent round-trip, single-loss and multi-packet-burst recovery vs. plain Opus under seeded loss, recovery correctness for patterns without a `main-1` block and under varying frame duration, recovered frames not inheriting the carrier's audio level, receive/ingress byte accounting matching the sender, the runtime send toggle, interop fallback when only one side enables RED (both directions), rtp-mode pass-through, and that the offer lists the RED PT in the m-line fmt list.

## Notes

- RED is *advertised* but not *preferred* (the primary codec stays first in the m-line). This is sufficient for str0m as a sender; a browser that gates RED sending on RED being the preferred codec would need the caller to reorder.
- Real-browser interop hasn't been exercised in-repo (no browser in the test harness) — a sanity check against Chrome/Firefox would be appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+  * Add RFC 2198 RED (redundant audio) for Opus via `RtcConfig::enable_red`: transparent send/recv
+    with single-loss recovery in frame mode, pass-through in rtp mode, and public `RedEncoder`/`RedDecoder` #982
+
 # 0.21.0
 
   * Upgrade to  dimpl 0.7.0 (breaking)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 # 0.23.1
 
   * Guard SRTP/SRTCP and RTCP parsing against malformed input #1029
+  * Add RFC 2198 RED (redundant audio) via `RtcConfig::enable_red` for all audio codecs (Opus, PCMU,
+    PCMA, G722): transparent send/recv with best-effort loss recovery in frame mode, a configurable
+    redundancy pattern via `RtcConfig::set_red_distances`, a runtime send-side on/off toggle via
+    `SdpApi`/`DirectApi::set_red_send`, pass-through in rtp mode, and public `RedEncoder`/`RedDecoder` #982
 
 # 0.23.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,10 +3,7 @@
 # 0.23.1
 
   * Guard SRTP/SRTCP and RTCP parsing against malformed input #1029
-  * Add RFC 2198 RED (redundant audio) via `RtcConfig::enable_red` for all audio codecs (Opus, PCMU,
-    PCMA, G722): transparent send/recv with best-effort loss recovery in frame mode, a configurable
-    redundancy pattern via `RtcConfig::set_red_distances`, a runtime send-side on/off toggle via
-    `SdpApi`/`DirectApi::set_red_send`, pass-through in rtp mode, and public `RedEncoder`/`RedDecoder` #982
+  * Add RFC 2198 RED (redundant audio) support for all audio codecs #982
 
 # 0.23.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,10 @@
 # Unreleased
 
+  * Add RFC 2198 RED (redundant audio) support for all audio codecs #982
+
 # 0.23.1
 
   * Guard SRTP/SRTCP and RTCP parsing against malformed input #1029
-  * Add RFC 2198 RED (redundant audio) support for all audio codecs #982
 
 # 0.23.0
 
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -1521,6 +1521,9 @@ impl AsSdpMediaLine for Media {
             if let Some(rtx) = p.resend() {
                 pts.push(rtx);
             }
+            if let Some(red) = p.red() {
+                pts.push(red);
+            }
         }
 
         if let Some(s) = self.simulcast() {
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -362,17 +362,6 @@ impl<'a> SdpApi<'a> {
         }
     }
 
-    /// Turn RFC 2198 RED wrapping on or off on the send side at runtime.
-    ///
-    /// RED availability is fixed when the m-line is negotiated; this only decides whether outgoing
-    /// packets are wrapped in RED right now, and needs no renegotiation (so it applies immediately,
-    /// with or without a following [`SdpApi::apply()`]). Enabling has no effect on media where RED
-    /// was not negotiated. This is the send-side lever an SFU uses to turn redundancy on for a leg
-    /// that is losing packets and off again when it recovers.
-    pub fn set_red_send(&mut self, mid: Mid, enabled: bool) {
-        self.rtc.session.set_red_send(mid, enabled);
-    }
-
     /// Stop an already existing media.
     ///
     /// The next generated offer emits the m-line with port 0 and excludes
--- a/src/change/sdp.rs
+++ b/src/change/sdp.rs
@@ -362,6 +362,17 @@ impl<'a> SdpApi<'a> {
         }
     }
 
+    /// Turn RFC 2198 RED wrapping on or off on the send side at runtime.
+    ///
+    /// RED availability is fixed when the m-line is negotiated; this only decides whether outgoing
+    /// packets are wrapped in RED right now, and needs no renegotiation (so it applies immediately,
+    /// with or without a following [`SdpApi::apply()`]). Enabling has no effect on media where RED
+    /// was not negotiated. This is the send-side lever an SFU uses to turn redundancy on for a leg
+    /// that is losing packets and off again when it recovers.
+    pub fn set_red_send(&mut self, mid: Mid, enabled: bool) {
+        self.rtc.session.set_red_send(mid, enabled);
+    }
+
     /// Stop an already existing media.
     ///
     /// The next generated offer emits the m-line with port 0 and excludes
@@ -1520,6 +1531,9 @@ impl AsSdpMediaLine for Media {
             pts.push(p.pt());
             if let Some(rtx) = p.resend() {
                 pts.push(rtx);
+            }
+            if let Some(red) = p.red() {
+                pts.push(red);
             }
         }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -229,6 +229,19 @@ impl RtcConfig {
         self
     }
 
+    /// Enable RFC 2198 RED (redundant Opus audio). Off by default. Requires Opus.
+    ///
+    /// RED carries a redundant copy of the previous Opus frame in each packet so the
+    /// receiver can recover single packet losses without retransmission, at the cost of
+    /// roughly doubling audio payload size.
+    ///
+    /// Call this after [`Self::enable_opus`]: RED folds onto the enabled Opus payload, so
+    /// enabling it while no Opus codec is configured is a no-op (and is logged).
+    pub fn enable_red(mut self, enabled: bool) -> Self {
+        self.codec_config.enable_red(enabled);
+        self
+    }
+
     /// Enable PCM μ-law audio codec.
     ///
     /// This is 14-bit audio compressed to 8-bit as specified by G.711
--- a/src/config.rs
+++ b/src/config.rs
@@ -224,7 +224,7 @@ impl RtcConfig {
     /// // For the session to use only OPUS and VP8.
     /// let mut rtc = RtcConfig::default()
     ///     .clear_codecs()
-    ///     .enable_opus(true)
+    ///     .enable_opus(true, false)
     ///     .enable_vp8(true)
     ///     .build(Instant::now());
     /// # }
@@ -234,28 +234,16 @@ impl RtcConfig {
         self
     }
 
-    /// Enable opus audio codec.
+    /// Enable opus audio codec, optionally with RFC 2198 RED (redundant audio).
     ///
-    /// Enabled by default.
-    pub fn enable_opus(mut self, enabled: bool) -> Self {
-        self.codec_config.enable_opus(enabled);
-        self
-    }
-
-    /// Enable RFC 2198 RED (redundant audio). Off by default. Requires an audio codec.
-    ///
-    /// RED carries redundant copies of earlier frames in each packet so the receiver can recover
-    /// packet losses without retransmission, at the cost of extra audio payload size. It applies to
-    /// every enabled audio codec (Opus, PCMU, PCMA, G722), each getting its own RED payload type.
-    ///
-    /// Call this after enabling audio codecs: RED folds onto the enabled audio payloads, so
-    /// enabling it while no audio codec is configured is a no-op (and is logged). The redundancy
-    /// depth is set with [`Self::set_red_distances`] (default one level, the previous frame), and
-    /// wrapping can be toggled at runtime on the send side without renegotiation via
-    /// [`SdpApi::set_red_send`][crate::change::SdpApi::set_red_send] /
+    /// Enabled by default (without RED). `use_red` folds a RED payload type onto Opus so the
+    /// receiver can recover packet loss without retransmission, at the cost of extra audio payload
+    /// size. Each audio codec carries its own RED payload type. The redundancy depth is set with
+    /// [`Self::set_red_distances`] (default one level, the previous frame), and wrapping can be
+    /// toggled at runtime on the send side without renegotiation via
     /// [`DirectApi::set_red_send`][crate::change::DirectApi::set_red_send].
-    pub fn enable_red(mut self, enabled: bool) -> Self {
-        self.codec_config.enable_red(enabled);
+    pub fn enable_opus(mut self, enabled: bool, use_red: bool) -> Self {
+        self.codec_config.enable_opus(enabled, use_red);
         self
     }
 
@@ -265,25 +253,27 @@ impl RtcConfig {
     ///
     /// The pattern is sanitised (zeros and values deeper than the recovery depth are dropped, then
     /// sorted and de-duped), falling back to `[1]` if empty. This only sets the pattern; enable RED
-    /// with [`Self::enable_red`] for it to take effect.
+    /// via the `use_red` argument of an audio codec (e.g. [`Self::enable_opus`]) for it to take effect.
     pub fn set_red_distances(mut self, distances: &[u32]) -> Self {
         self.codec_config.set_red_distances(distances);
         self
     }
 
-    /// Enable PCM μ-law audio codec.
+    /// Enable PCM μ-law audio codec, optionally with RFC 2198 RED (see [`Self::enable_opus`] for
+    /// `use_red`).
     ///
     /// This is 14-bit audio compressed to 8-bit as specified by G.711
-    pub fn enable_pcmu(mut self, enabled: bool) -> Self {
-        self.codec_config.enable_pcmu(enabled);
+    pub fn enable_pcmu(mut self, enabled: bool, use_red: bool) -> Self {
+        self.codec_config.enable_pcmu(enabled, use_red);
         self
     }
 
-    /// Enable PCM a-law audio codec.
+    /// Enable PCM a-law audio codec, optionally with RFC 2198 RED (see [`Self::enable_opus`] for
+    /// `use_red`).
     ///
     /// This is 13-bit audio compressed to 8-bit as specified by G.711
-    pub fn enable_pcma(mut self, enabled: bool) -> Self {
-        self.codec_config.enable_pcma(enabled);
+    pub fn enable_pcma(mut self, enabled: bool, use_red: bool) -> Self {
+        self.codec_config.enable_pcma(enabled, use_red);
         self
     }
 
@@ -294,8 +284,10 @@ impl RtcConfig {
     /// codec samples at 16 kHz. str0m treats G722 as a 16 kHz codec in the API and
     /// handles the 8 kHz RTP timestamp mapping internally. See
     /// <https://en.wikipedia.org/wiki/RTP_payload_formats#cite_note-55>
-    pub fn enable_g722(mut self, enabled: bool) -> Self {
-        self.codec_config.enable_g722(enabled);
+    ///
+    /// See [`Self::enable_opus`] for `use_red`.
+    pub fn enable_g722(mut self, enabled: bool, use_red: bool) -> Self {
+        self.codec_config.enable_g722(enabled, use_red);
         self
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -242,6 +242,35 @@ impl RtcConfig {
         self
     }
 
+    /// Enable RFC 2198 RED (redundant audio). Off by default. Requires an audio codec.
+    ///
+    /// RED carries redundant copies of earlier frames in each packet so the receiver can recover
+    /// packet losses without retransmission, at the cost of extra audio payload size. It applies to
+    /// every enabled audio codec (Opus, PCMU, PCMA, G722), each getting its own RED payload type.
+    ///
+    /// Call this after enabling audio codecs: RED folds onto the enabled audio payloads, so
+    /// enabling it while no audio codec is configured is a no-op (and is logged). The redundancy
+    /// depth is set with [`Self::set_red_distances`] (default one level, the previous frame), and
+    /// wrapping can be toggled at runtime on the send side without renegotiation via
+    /// [`SdpApi::set_red_send`][crate::change::SdpApi::set_red_send] /
+    /// [`DirectApi::set_red_send`][crate::change::DirectApi::set_red_send].
+    pub fn enable_red(mut self, enabled: bool) -> Self {
+        self.codec_config.enable_red(enabled);
+        self
+    }
+
+    /// Set the RFC 2198 RED send-side redundancy pattern: how many packets back each redundant
+    /// level carries. `[1]` (the default) sends one level, the previous frame, as browsers do;
+    /// `[1, 3, 5]` sends three levels for higher loss at the cost of more bandwidth.
+    ///
+    /// The pattern is sanitised (zeros and values deeper than the recovery depth are dropped, then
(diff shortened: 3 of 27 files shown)


src/format/codec_config.rs
```diff
@@ -119,6 +122,7 @@ impl CodecConfig {
                 format,
             },
             resend,
+            red: None,
```
algesten: Scoot this down to before `locked: false`

src/format/codec_config.rs
```diff
@@ -232,6 +236,21 @@ impl CodecConfig {
...
+    pub fn enable_red(&mut self, enabled: bool) {
+        if enabled && !self.params.iter().any(|p| p.spec.codec == Codec::Opus) {
+            warn!("enable_red(true) ignored: Opus is not enabled");
+        }
+        for p in &mut self.params {
+            if p.spec.codec == Codec::Opus {
+                p.red = if enabled { Some(PT_RED) } else { None };
+            }
+        }
+    }
```
algesten, line 309: In the `Codec` enum, we treat `Red` as a "doc(hidden)" value and I believe that is the correct abstraction: ignore the SDP-level technicality of it being a "codec".

In that spirit, I think the better API here would be `enable_opus(enabled: bool, use_red: bool)` - a breaking change, but I think that's fine. If we absolutely don't want to break, we can do `enable_opus_with_red(enabled: bool)`

src/format/payload_params.rs
```diff
@@ -116,6 +120,7 @@ impl PartialEq for PayloadParams {
     fn eq(&self, other: &Self) -> bool {
         self.pt == other.pt
             && self.resend == other.resend
+            && self.red == other.red
```
algesten: Scoot to last

src/format/payload_params.rs
```diff
@@ -155,6 +160,7 @@ impl PayloadParams {
         PayloadParams {
             pt,
             resend,
+            red: None,
```
algesten: Move to before locked

src/format/payload_params.rs
```diff
@@ -183,6 +189,7 @@ impl PayloadParams {
         PayloadParams {
             pt,
             resend: None,
+            red: None,
```
algesten: Ditto

src/format/payload_params.rs
```diff
@@ -209,6 +216,11 @@ impl PayloadParams {
         self.resend
     }
 
+    /// The RFC 2198 RED payload type wrapping this primary codec, if negotiated.
+    pub fn red(&self) -> Option<Pt> {
+        self.red
+    }
+
```
algesten: Move to before `match_score` fn below.

src/format/payload_params.rs
```diff
@@ -688,6 +700,13 @@ impl PayloadParams {
                     self.resend, remote_rtx
                 );
             }
+
+            if self.red != first.red {
+                warn!(
+                    "Ignore remote PT RED change {:?} => {:?}",
+                    self.red, first.red
+                );
```
algesten, line 716: Is this an expected or unexpected situation? We use `warn` for unexpected - but if a browser is consistently doing it, we rather have this on info or debug.

src/format/payload_params.rs
```diff
@@ -85,6 +85,10 @@ pub struct PayloadParams {
     /// This is used to, via PT, separate RTX resend streams from the main stream.
     pub(crate) resend: Option<Pt>,
 
+    /// The RFC 2198 RED payload type that wraps this primary codec, if negotiated.
+    /// Like `resend`, this links a secondary PT to this primary via SDP.
+    pub(crate) red: Option<Pt>,
+
```
algesten: Move to before `locked`

src/media/mod.rs
```diff
@@ -442,11 +442,16 @@ impl Media {
...
+        let payloader = self.payloaders.entry((pt, rid)).or_insert_with(|| {
             // Unwrap is OK, the pt should be checked already when calling this function.
             let params = params.iter().find(|p| p.pt == pt).unwrap();
             Payloader::new(params.spec, vp9_mode)
-        })
+        });
+        // Keep the RED linkage current with the negotiated params so a remapped or dropped RED
+        // PT can't go stale in the cached payloader.
+        let red = params.iter().find(|p| p.pt == pt).and_then(|p| p.red);
+        payloader.set_red(red, pt);
```
algesten: This is not great. We should not force mutating business logic into what is essentially a getter.

src/packet/payload.rs
```diff
@@ -2,16 +2,50 @@ use crate::format::CodecSpec;
...
 use super::PacketError;
-use super::{CodecPacketizer, Packetizer};
+use super::{CodecPacketizer, Packetizer, RedEncoder, RedundantBlock};
 
 #[derive(Debug)]
 pub struct Payloader {
     pack: CodecPacketizer,
     clock_rate: Frequency,
+    red: Option<RedState>,
+}
```
algesten, line 19: This is not the right level for this code. `Payloader` is generic handling all payloader, and we don't want to see any payload specific state keeping here.

I notice there is some special vp9 and h265 treatment that has sneaked in on this level. I will do a PR to fix that once we landed this.

src/sdp/parser.rs
```diff
@@ -600,6 +600,27 @@ where
...
+            } else {
+                None
+            };
+            match pts.and_then(|v| v.first().copied()) {
+                Some(primary) => Ok(MediaAttribute::Fmtp {
+                    pt: red_pt,
+                    values: vec![FormatParam::Red(Pt::from(primary))],
+                }),
+                None => Err(StreamErrorFor::<Input>::message_format("not a RED fmtp")),
+            }
```
algesten: We use `combine` to avoid frivolous allocations and have consistent expressions for reading numbers etc. The above code block does inelegant manual parsing via collect(), split(), etc - which also goes against the rest of the file.

src/streams/receive.rs
```diff
@@ -452,6 +452,25 @@ impl StreamRx {
...
+    /// Build an [`RtpPacket`] WITHOUT counting it in receive stats. Used for RFC 2198
+    /// FEC-recovered packets, which were never received on the wire (the carrying packet's
+    /// bytes are already counted by `handle_rtp`).
+    pub(crate) fn make_rtp_packet(
+        &mut self,
+        now: Instant,
+        header: RtpHeader,
+        payload: Arc<[u8]>,
+        seq_no: SeqNo,
+        time: MediaTime,
```
algesten, line 600: Is this some drive-by cleanup not related to the PR?

src/config.rs
```diff
@@ -229,6 +229,19 @@ impl RtcConfig {
...
+    /// receiver can recover single packet losses without retransmission, at the cost of
+    /// roughly doubling audio payload size.
+    ///
+    /// Call this after [`Self::enable_opus`]: RED folds onto the enabled Opus payload, so
+    /// enabling it while no Opus codec is configured is a no-op (and is logged).
+    pub fn enable_red(mut self, enabled: bool) -> Self {
+        self.codec_config.enable_red(enabled);
+        self
+    }
+
```
algesten, line 249: Let's break API and make `enable_opus(mut self, enabled: bool, use_red: bool)`

algesten:
I started a review this morning, and I'm still working on it. I have some concerns of the placement of the changes in session.rs, and I want to explore myself if there is some better solution I would feel happier with.

xnorpx:
In general one would not enable RED unless there is packetloss that shows there is a need for it. So it should be enabled on demand. Also RED can support all codecs, so it should not be bound to Opus.

algesten:
> In general one would not enable RED unless there is packetloss that shows there is a need for it. So it should be enabled on demand. Also RED can support all codecs, so it should not be bound to Opus.

You mean all _audio_ codecs, right?

xnorpx:
> > In general one would not enable RED unless there is packetloss that shows there is a need for it. So it should be enabled on demand. Also RED can support all codecs, so it should not be bound to Opus.
> 
> You mean all _audio_ codecs, right?

Yes, let me add some more information here.

RED should be available for all Audio Codecs.

You should be able to switch it on demand, so you start sending RED when you need and not all the time unless you want to.

RED is not limited to 1 redundant payload or just take main-1.

You could do a RED packet with main, main-1, main-3, main-5 etc. If we anyway do the work here I think this should be doable.

algesten:
@xnorpx 

> You should be able to switch it on demand, so you start sending RED when you need and not all the time unless you want to.

I was debating this with Claude. It seems a browser doesn't have an "easy" way to turn RED on/off. You need to do an SDP negotiation to set the codec preference "red, opus" then the browser starts sending RED.

This is not true for SFU -> Browser, because any PT that is declared for the m-line is valid, so we can "just switch" at any point.

str0m currently doesn't support [renegotiating codecs](https://github.com/algesten/str0m/blob/main/docs/SDP.md#renegotiation-and-narrowed-pt-lists).

Do you think runtime enabling of RED is so important that we should work on renegotiating codecs? I know you guys don't use SDP at all, so the question is maybe meaningless :P

xnorpx:
Right, so Chrome you need to re-negotiate to be able to start sending RED or not. I think we can keep that as a separate issue.

We just need to make sure str0m can depacketize RED and use the main packets depacketizer

For sending, (str0m is also used for clients) I think there should be a way to indicate that we now want to to send this codec in RED with this set of distances. We could limit the amount of RED packets to maybe 4 then just have a fixed array that indicate the distance from the main packet. 0 means not used.

[1, 3, 5] would mean that we keep a buffer around to save the last 5 sent packets and then resend them in that pattern.

But I am not sure where this sort of dynamic configure would fit in the sample api.

algesten:
@xnorpx 

> Right, so Chrome you need to re-negotiate to be able to start sending RED or not. I think we can keep that as a separate issue.

It sort of affects the shape of the API. Whether it involved SDP API or not. I will ponder.

> For sending, (str0m is also used for clients) I think there should be a way to indicate that we now want to to send this codec in RED with this set of distances. We could limit the amount of RED packets to maybe 4 then just have a fixed array that indicate the distance from the main packet. 0 means not used.

Are you saying you want to dynamically be able to change the sets of distances, or can that be configured once and for all and then you chose RED on/off?

farit2000:
Thanks, this has been really useful. I went deep on the receive side and it convinced me we should treat this as a proper complete RED rather than the minimal Opus version the PR is now, so let me lay out where I landed, flag one small existing bug I found along the way, and then a question about how you want to take it.

The thing that reframed it for me: RFC 2198 tags each redundant block with a timestamp offset, not a sequence number, but our DepacketizingBuffer is keyed and deduped by seq_no. So to actually recover a lost packet we have to hand it a correct seq_no. The current code works it out positionally, rec_seq = seq - (n - i), which is only right when the redundancy is contiguous, main-1 then main-2. The moment you allow main-1, main-3, main-5 like you mentioned @xnorpx, it produces wrong seq_nos and the buffer silently drops both the recovered frame and the real one. It doesn't bite today only because we send a single main-1 block. The fix is to derive the distance from the timestamp offset instead, using the negotiated ptime for the frame size, offset / frame_size. That reproduces the contiguous case exactly and also handles the non-contiguous one, and I'd make it correct-or-skip: if an offset doesn't land on a clean packet boundary, or the seq isn't actually missing, just skip that block rather than guess. So it never corrupts, and the plain unwrap underneath is always correct regardless. Frame size isn't truly fixed (Opus can be 10/20/40/60ms), so recovering arbitrary distances is best effort by nature, which I think is fine as long as we're honest about it.

That pushes me to split the receive path into the two things it currently tangles together, which might be close to the session.rs restructuring you had in mind @algesten. First, just unwrapping RED and handing the primary block to the normal depacketizer, no seq mapping, works for any audio codec and any block count. That's your must-have @xnorpx and it's unconditionally correct. Second, the best effort recovery sitting on top, isolated. It extracts cleanly too, a method on StreamRx that returns the recovered packets with the delivery staying in the session loop, same shape as un_rtx.

Audio agnostic turned out to be a bit more than swapping == Opus for is_audio(). The RED rtpmap carries a clock rate per pt, so a single RED pt can't cover two audio codecs with different clock rates, and today we'd actually hand the same pt 63 to both and collide in SDP. So the right thing is a RED pt per red-enabled codec, each advertised with that codec's rtp clock rate. While I was in there I noticed the RED rtpmap emits the codec's user-facing clock rate rather than the rtp clock rate, which is invisible for Opus but would be wrong for G722, rtx already does it the right way. Small fix regardless.

On static vs dynamic distances, I'd keep the distance pattern static and add a plain on/off at runtime. We can't renegotiate codecs, so the pt and the pattern are fixed up front anyway, but flipping wrapping on and off needs no renegotiation. It is new api though, nothing like it exists yet, so it would follow the set_direction shape. Browser side on/off and auto enabling on loss both still look like separate issues to me.

Where I'd like your steer: this is a fair bit more than the minimal PR. The byte codec and the negotiation model carry over as is, but the recovery, the per-codec RED pt, the send side and the on/off api all need real rework to become the complete thing. Would you rather I evolve this PR toward that, or settle the shape first, especially where you want the recovery code to live, and I redo the changed parts against that. Happy either way, I just don't want to build the recovery out in a spot you're about to restructure.

algesten:
> Where I'd like your steer: this is a fair bit more than the minimal PR. The byte codec and the negotiation model carry over as is, but the recovery, the per-codec RED pt, the send side and the on/off api all need real rework to become the complete thing. Would you rather I evolve this PR toward that, or settle the shape first, especially where you want the recovery code to live, and I redo the changed parts against that. Happy either way, I just don't want to build the recovery out in a spot you're about to restructure.

The first junction is to decide if we are letting go of the "str0m does not further negotiate m-lines" or stick with it. In practice, I never seen the need to renegotiate m-lines after the initial OFFER/ANSWER, but then my experience might be too limited. I haven't used RED, and I don't know if toggling RED on/off is common.

The fastest/simplest way forward is to say that:

* RED via SDP is toggled ON by the initial OFFER/ANSWER and we don't support toggling it off/on again.
* Using DirectAPI you can toggle ON/OFF without renegotiation

farit2000:
That works for me and it keeps str0m's model intact, which I like. The initial offer/answer decides whether RED is available on the m-line, and we don't renegotiate to add or remove it after. Agreed.

One nuance on the toggle though, coming back to your own point that an SFU sending to a browser can just switch at any time. Once the RED pt is negotiated, whether a given outgoing packet is actually wrapped in RED is purely a send side choice with no renegotiation needed, since both the RED pt and the plain pt are valid on the m-line and the receiver already handles both. So I'd hang the on/off on the send side and let it work for an SDP negotiated sender too, not just DirectAPI. SDP decides RED is available, a runtime flag decides whether we wrap right now. That's the case where toggling actually earns its keep, an SFU turning RED on for a leg that's losing packets and off again when it's clean, without touching SDP. A pure client without SDP gets the same flag through DirectAPI.

So the shape I'm converging on is this: RED availability on the m-line is fixed at negotiation or set up via DirectAPI, the distance pattern is static config, and the only runtime lever is wrap on/off on the send side, available to both paths, with no renegotiation anywhere. And picking up @xnorpx's earlier point, this is for all audio codecs rather than just Opus, which in practice means a RED pt per codec the same way RTX already gets a resend pt per codec, each with that codec's rtp clock rate. I can't say how common toggling is either, but as a plain send side flag it's cheap enough to just have, and it covers the loss driven case @xnorpx had in mind.

algesten:
To me this sounds great.

farit2000:
Reworked this from the minimal Opus-only version into the full design we agreed on above, and rebased on current main (it sits on top of the Comfort Noise work now).

RED covers all four audio codecs now, not just Opus, with a payload type per codec the way RTX gets a resend pt per codec, each on its own rtp clock rate (so G722's RED sits on 8000, not the 16000 it samples at). Availability is fixed at negotiation or via DirectApi and nothing renegotiates to add or remove it. The only runtime lever is wrapping on and off on the send side, on both SdpApi and DirectApi with no renegotiation, so an SFU can flip it on a leg that starts losing packets.

On receive, recovery now works out the distance from each block's timestamp offset instead of guessing it positionally. The depacketizing buffer is keyed by seq, so a redundant block has to map back to a real seq, and the positional version only held for contiguous main-1 then main-2; going off the offset also handles the [1,3,5] patterns. It's correct-or-skip: if a block doesn't land on a clean packet boundary or its seq isn't actually missing, it's dropped rather than mapped to the wrong seq. The primary is always unwrapped to the normal depacketizer regardless. It stays best-effort since there's no reliable frame size at that point.

Distances are static config (set_red_distances), and the send side drops the deepest levels when a packet would go over the MTU, which also fixed a latent overflow panic on large audio frames.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
   * Support RFC 3389 Comfort Noise payloads #1013
   * Apply reliability parameters to a DCEP-receiving in-band channel #1004
+  * Add RFC 2198 RED (redundant audio) via `RtcConfig::enable_red` for all audio codecs (Opus, PCMU,
+    PCMA, G722): transparent send/recv with best-effort loss recovery in frame mode, a configurable
+    redundancy pattern via `RtcConfig::set_red_distances`, a runtime send-side on/off toggle via
+    `SdpApi`/`DirectApi::set_red_send`, pass-through in rtp mode, and public `RedEncoder`/`RedDecoder` #982
```
algesten: This does not follow how the changelog is written.

src/change/sdp.rs
```diff
@@ -362,6 +362,17 @@ impl<'a> SdpApi<'a> {
...
+    ///
+    /// RED availability is fixed when the m-line is negotiated; this only decides whether outgoing
+    /// packets are wrapped in RED right now, and needs no renegotiation (so it applies immediately,
+    /// with or without a following [`SdpApi::apply()`]). Enabling has no effect on media where RED
+    /// was not negotiated. This is the send-side lever an SFU uses to turn redundancy on for a leg
+    /// that is losing packets and off again when it recovers.
+    pub fn set_red_send(&mut self, mid: Mid, enabled: bool) {
+        self.rtc.session.set_red_send(mid, enabled);
+    }
+
```
algesten: Let's just leave the toggling in Direct API. The SDP API should lead to SDP negotation, at least sometimes. An API call that never does that is weird. On balance, let's just remove this one here.

algesten (commented):
Some thoughts

src/format/codec_config.rs
farit2000: Done. Folded use_red into all four audio enable fns (enable_opus/pcmu/pcma/g722) and dropped enable_red, since RED is all audio now. Codec::Red stays hidden.

src/config.rs
farit2000: Done, the RtcConfig builders mirror it: enable_opus/pcmu/pcma/g722 take use_red, enable_red removed.

src/format/payload_params.rs
farit2000: This one sits next to the PT and RTX "Ignore remote ... change" warns and shares their "shouldn't happen, validate earlier" comment, and only fires on a demanded receive-direction change since the send-direction case hits the local_is_controlling return just above. Kept it a warn to match those two, but happy to drop all three to debug together if you'd prefer.

src/media/mod.rs
farit2000: Done. payloader_for is a plain getter again and the RED sync moved into do_payload.

src/packet/payload.rs
farit2000: Agreed it's not the right level. Left it here for now since you mentioned a separate PR to pull the codec-specific state (vp9/h265 too) out of the payloader, so it can go with the rest.

src/sdp/parser.rs
farit2000: Done, rewrote it with combine to match the rest of the file.

src/streams/receive.rs
farit2000: Not drive-by. It's the make_rtp_packet split so a RED-recovered packet is built without counting in receive stats, since it never arrived on the wire. Normal handle_rtp still counts as before.

src/change/sdp.rs
farit2000: Done, removed it. The toggle lives on DirectApi only now.

CHANGELOG.md
farit2000: Done, one line now.

src/format/codec_config.rs
farit2000: Done.

src/format/payload_params.rs
farit2000: Done.

src/format/payload_params.rs
farit2000: Done.

src/format/payload_params.rs
farit2000: Done.

src/format/payload_params.rs
farit2000: Done.

src/format/payload_params.rs
farit2000: Done.

farit2000:
Pushed a commit addressing the review, replies inline. Main one: folded RED enabling into the audio codec enable fns (use_red) and dropped enable_red.

algesten:
@farit2000 I found some edge cases. I added failing test to illustrate them.

farit2000:
Thanks, these were good catches, all real. Pushed fixes for all of them.

Quick rundown: the default RED pts for G722/PCMU/PCMA were 64/65/66, which land in the 64-95 RTP/RTCP mux range, so I moved them to 62/61/60 (Opus RED at 63 was already fine). Conflicting remote RED mappings no longer panic, they drop the duplicate. RED wrapping now reserves the 1-byte primary header at packetization so the wrapped packet never exceeds the MTU. Recovery now bails when it can't establish the packet duration from offsets alone (a [3,5] pattern with no main-1), instead of treating the smallest offset as one packet, and set_red_distances now always includes distance 1 as the anchor so a pattern like [2,4] can't be configured into silent mis-recovery. The public encoder is capped at the decoder's block bound so its output always decodes. un_red now drops a primary block whose pt isn't the negotiated codec. And receive stats now count the full RED wire payload, matching the sender.

The audio-level one needs a word. A recovered frame genuinely can't carry its own audio level: RFC 2198 redundant blocks are codec payload only, the audio-level extension lives on the carrier packet's RTP header, so a recovered frame's original level simply isn't on the wire (browsers can't recover it either). So the fix clears the carrier's extensions on recovered frames, and they're delivered with no audio level rather than the carrier's wrong one. That's what your test's name asks for, but the assertion checked for the frame's own level, which nothing can satisfy, so I adjusted it to flag a recovered frame only if it carries a level that isn't its own. Shout if you'd rather handle that differently.

farit2000:
Rebased on latest main so it's mergeable again, and CI is green across the matrix. I think the review is fully addressed now, other than the RedState-in-Payloader placement you said you'd take in a separate PR. Anything else you'd like before this can go in, or is it good from your side? No rush.

algesten (approved):

algesten:
Thanks!

PR #983, opened 2026-06-24
fix: reject SDP payload types > 127 instead of panicking

farit2000:
## What

A remote SDP offer that advertises a codec at a payload type outside the 7-bit RTP range (e.g. `a=rtpmap:200 opus/48000/2`) panics the `Rtc`: the out-of-range payload type parses into a `Pt`, and when it matches a locally-enabled codec it is claimed against a 128-entry table (`[bool; 128]`), indexing out of bounds.

This violates str0m's "never panic on user input" invariant — a single malformed or hostile offer crashes the instance. It is pre-existing and not specific to a codec (the main and RTX payload-type paths reach the same claim), so the fix is at the shared `a=rtpmap` parser.

## Fix

Bound the `a=rtpmap` payload-type parser to the valid `0..=127` range. An out-of-range payload type makes the `a=rtpmap` line fall through to `Unused` (ignored), so no `PayloadParams` with an out-of-range PT is ever created. The offer is then rejected gracefully at parse (the m-line PT has no matching rtpmap) instead of panicking later when PTs are claimed.

## Tests

- `rtpmap_payload_type_over_127_is_rejected` (unit): an out-of-range `a=rtpmap` PT does not parse as an `RtpMap`; PT 127 still does.
- `offer_with_out_of_range_pt_is_rejected` (integration): an offer advertising a codec at PT 200 is rejected, not a panic.

Both fail without the fix (the integration test previously panicked on the out-of-bounds index).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Reject SDP payload types > 127 instead of panicking when claiming PTs #983
   * Add H266/VVC RTP packetizer and depacketizer #971
   * Negotiate SCTP max message size for data channels #852
   * Expose the negotiated DTLS protocol version #979
--- a/src/sdp/parser.rs
+++ b/src/sdp/parser.rs
@@ -547,9 +547,17 @@ where
 
     let pt = || {
         not_sp().and_then(|s| {
-            s.parse::<u8>()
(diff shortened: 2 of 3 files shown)


algesten (approved):
LGTM!

Issue #922, opened 2026-03-27
Performance


Razzwan:
I really want to improve the library’s performance. Right now, as I mentioned in [that post](https://github.com/algesten/str0m/issues/911), my code is too slow after the `set_current_bitrate` method was removed.

What I want to do:

1. Optimize the existing code to make it as fast as possible.

2. Add three entities: consumer, audio publisher, and publisher-consumer (instead of Client from the examples), plus the routing logic for SDP data between them. This should fix the related errors and improve performance.

What would you recommend as a starting point?
Are there any performance tests available for the library?
What are the existing performance bottlenecks in the library?

PS: I'm not sure I can handle it, so I want to start with something simpler. 

PPS: On the other hand, this is currently the weakest point in my app, and without addressing it, further growth is impossible. That's why I want to invest time in developing these features.

algesten:
> I really want to improve the library’s performance. Right now, as I mentioned in [that post](https://github.com/algesten/str0m/issues/911), my code is too slow after the `set_current_bitrate` method was removed.

I don't know why your code is slow there, but we're running 10+ participant room video conferencing and it's fine (we are also using BWE). My hunch is that it's not str0m itself that is the bottleneck.

Having said that, optimization is always good.

> What would you recommend as a starting point?

The optimization work I've done so far is using [`flamegraph`](https://crates.io/crates/flamegraph). My test process is:

1. `cargo run --release --example chat` - it's VERY important to use `--release` or you don't see actual performance problems.
2. Connect two browsers to the chat server, and start camera in each direction.
3. Find the pid of the chat server `ps auxwww | grep chat | grep -v grep`
4. `flamegraph -p <pid>`
5. Run the test for 10-15 seconds
6. Ctrl-C
7. `open -a safari flamegraph.svg`

The below image is maybe not clickable, but in Safari you can drill into the stack to see what takes time.

![Image](https://github.com/user-attachments/assets/babe7fbc-3d29-4b32-8f78-f9611e1b4173)

As you can see, we're spending more time in send/recv data from the UDP port than processing data in str0m.

<img width="1052" height="449" alt="Image" src="https://github.com/user-attachments/assets/76d987de-e0af-4333-a5d3-269142085233" />

If you click into the str0m processing you'll find a lot of time is spent in encryption/decryption, and there's not really anything we can do about that.

There is definitely room for improvement in testing this.

* More participants. A two-way 2 participant is small. What if we connect 10?
[...]

PR #984, opened 2026-06-24
[poc] binary timer heap

lherman-cs:
WARNING: this is just a POC.

Issue: https://github.com/algesten/str0m/issues/922

The idea here is to reduce the amount of work per time update. Currently, we have to traverse every subsystem every time we get a newer time. If N is the number of subsystems (the number of streams and other fixed subsystems like DTLS, ICE, SCTP, etc), then we have to spend O(N) time each. 

Each of this system unlikely to have the same deadline. So, ideally, we should only do work when needed, aka bringing us closer to a constant time per time update (Polling -> Event Driven).

**Before:**

Screenshot 2026-06-23 at 9 20 24 PM

**After**

Screenshot 2026-06-23 at 9 20 32 PM

**Result**

It's interesting to see how the number of spikes reduced on the profiler timelines.
The CPU usage reduction is marginal for now, from ~56% -> ~54%. But there is plenty of room for expansion.

Edit: With many rooms of 4 fully interactive participants, the CPU usage went down from ~72% to ~60% CPU. The above workload consisted of rooms with mostly idle participants.

Screenshot 2026-06-23 at 9 25 12 PM

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -910,7 +910,6 @@ impl Session {
         let pacing_at = self.pacer.poll_timeout();
         let packetize_at = self.medias.iter().flat_map(|m| m.poll_timeout()).next();
         let paused_at = self.paused_at();
-        let send_stream_at = self.streams.send_stream();
 
         // Gives us built-in reason
         let bwe_at = self
@@ -926,7 +925,6 @@ impl Session {
             .soonest(pacing_at)
             .soonest((packetize_at, Reason::Packetize))
             .soonest((paused_at, Reason::PauseCheck))
-            .soonest((send_stream_at, Reason::SendStream))
             .soonest(bwe_at)
     }
 
--- a/src/streams/mod.rs
+++ b/src/streams/mod.rs
@@ -55,7 +55,6 @@ struct StreamDeadline {
 
 impl Ord for StreamDeadline {
     fn cmp(&self, other: &Self) -> Ordering {
-        // Reverse comparison to turn BinaryHeap into a Min-Heap (earliest time first)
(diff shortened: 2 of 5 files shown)


src/streams/mod.rs
```diff
@@ -423,50 +454,47 @@ impl Streams {
...
-            stream.maybe_create_remb_request(sender_ssrc, feedback);
-
-            // All StreamRx belonging to the same Mid are reported together.
-            if self.mids_to_report.contains(&stream.mid()) {
-                stream.create_rr_and_update(now, sender_ssrc, feedback);
-            }
-
-            if do_nack {
-                stream.maybe_create_nack(sender_ssrc, feedback);
+        while let Some(entry) = self.deadline_heap.peek() {
```
lherman-cs, line 459: TODO: group the same MID streams in the SR/RR reports

src/session.rs
```diff
@@ -910,7 +910,6 @@ impl Session {
         let pacing_at = self.pacer.poll_timeout();
         let packetize_at = self.medias.iter().flat_map(|m| m.poll_timeout()).next();
         let paused_at = self.paused_at();
-        let send_stream_at = self.streams.send_stream();
```
lherman-cs: The send queue packet's timestamps are populated during output polling. I was trying to remove N streams iterations. It's currently failing on the bwe test, not ramping quickly enough. I might've misunderstood the timing bit here.

lherman-cs:
Related conversation, https://discordapp.com/channels/1436725650302959829/1436796781462687817

algesten:
Superseded by #1008

PR #985, opened 2026-06-24
Make DTLS client certs configurable

jshanson7:
Make client certificate requests configurable for dimpl-backed DTLS.

Default stays true. This lets callers disable `CertificateRequest` when talking to Chrome/WebRTC over DTLS 1.3.

### DIFF
--- a/crates/proto/src/crypto/provider.rs
+++ b/crates/proto/src/crypto/provider.rs
@@ -150,12 +150,16 @@ pub trait DtlsProvider: CryptoSafe {
     ///
     /// mtu: Backends that can size their record layer according to the provided MTU should do so;
     /// backends that cannot ignore it. `None` means use the backend's default.
+    ///
+    /// client_certificate_required: Servers request and require a client certificate when true;
+    /// backends that cannot configure this behavior ignore it.
     fn new_dtls(
         &self,
         cert: &DtlsCert,
         now: Instant,
         dtls_version: DtlsVersion,
         mtu: Option<usize>,
+        client_certificate_required: bool,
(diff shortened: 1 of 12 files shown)


algesten:
@jshanson7 in what situation would you want this?

Does libwebrtc allow you
to configure this?

jshanson7:
> @jshanson7 in what situation would you want this?
> 
> Does libwebrtc allow you
> to configure this?

@algesten hi! so with ice-lite and DTLS 1.3, dimpl sends a `CertificateRequest` and then chrome immediately aborts the handshake.

afaict the browser WebRTC API only lets you provide local `RTCCertificate`s, not configure client auth. chrome doesn't seem to expose lower level libwebrtc options via `RTCPeerConnection`.

so this is basically an opt-out, with the default left unchanged.

happy to update this if you think there's a better approach.

algesten:
@xnorpx DTLS 1.3 does that change anything?

@jshanson7 i never heard of a WebRTC situation where you don't validate the remote certificate. 

In standard SDP you send the fingerprint (hash) of the certificate and then during DTLS handshake you request remote cert and compare the fingerprint. 

This is the basis for security since regular WebRTC uses self-signed certs.

Turning off remote cert is like saying "i don't care who you are" and it would be a very odd thing to do in regular WebRTC. 

DTLS 1.2 or 1.3 shouldn't make
a difference. Nor should ice lite.

jshanson7:
makes sense. i think there must be a bug in either chrome or dimpl causing the dtls 1.3 handshake to fail. will see if i can track that down instead.

jshanson7:
@algesten figured it out: https://github.com/algesten/dimpl/pull/153

PR #986, opened 2026-06-25
Handle SCTP AssociationLost as closed

gab8i:
Summary

- Normalize SCTP association loss to the existing `Event::Closed` event.
- Keep the top-level RTC event abstraction independent of whether closure came from DTLS or SCTP.
- Bump `sctp-proto` to 0.10.0.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1030,6 +1030,11 @@ pub enum Event {
     /// We have received the DTLS close packet
     Closed,
 
+    /// The sctp association has been closed.
+    ///
+    /// Emitted when the peer emitted an SCTP abort packet.
+    SctpAssociationClosed,
+
     /// Debug output of incoming and outgoing RTCP/RTP packets.
     ///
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1018,10 +1018,15 @@ pub enum Event {
...
     RtpPacket(RtpPacket),
-
+    
     /// We have received the DTLS close packet
     Closed,
 
+    /// The sctp association has been closed.
+    ///
+    /// Emitted when the peer emitted an SCTP abort packet.
+    SctpAssociationClosed,
```
algesten: The abstraction level is not correct. On this level we are not really interested in _how_ the WebRTC session was closed. I think we should merge the logic into `Closed` event, i.e. on DTLS close or SCTP association we normalize to a single closed.

algesten:
@gab8i sctp-proto 0.10.0 is released.

PR #995, opened 2026-06-28
Add support for AMR-WB

xnorpx:
Not ready, need more iterations

## Architecture / mental model

This change adds an **AMR-WB RTP payload adapter**, not an AMR-WB encoder or decoder. Applications provide and receive already encoded 3GPP IF frames; str0m negotiates the supported transport profile and converts between IF frames and RFC 4867 RTP payloads.

There are three representations to keep distinct:

1. **SDP representation:** `CodecSpec` and `FormatParams` describe the negotiated payload shape.
2. **Application representation:** 3GPP IF frames, consisting of a one-byte `FT/Q` header followed by encoded speech bytes.
3. **Network representation:** RFC 4867 RTP payloads containing CMR, ToC entries, and speech bits.

```mermaid
flowchart LR
    subgraph Control["Control plane: SDP negotiation"]
        Config["CodecConfig<br/>PT 122, 16 kHz, mono"]
        Params["FormatParam / FormatParams"]
        Gate["AMR-WB compatibility gate"]
        Spec["Negotiated CodecSpec"]

        Config --> Params --> Gate --> Spec
    end

    subgraph Send["Send path"]
        Writer["Writer::write<br/>one 3GPP IF frame + MediaTime"]
        Payloader["Payloader<br/>select negotiated layout"]
        Packetizer["AmrWbPacketizer<br/>IF frame -> RFC 4867 payload"]
        RTPOut["RTP / SRTP"]

        Writer --> Payloader --> Packetizer --> RTPOut
    end

    subgraph Receive["Receive path"]
        RTPIn["RTP / SRTP"]
        Buffer["DepacketizingBuffer"]
        Depacketizer["AmrWbDepacketizer<br/>RFC 4867 -> IF frames"]
        Event["Event::MediaData"]

        RTPIn --> Buffer --> Depacketizer --> Event
    end

    Spec -. configures .-> Payloader
    Spec -. configures .-> Depacketizer
```

### Control plane

`enable_amr_wb(true)` adds an opt-in mono codec configuration using dynamic PT 122, a 16 kHz RTP clock, and the common IMS/VoLTE format:

```text
octet-align=1;mode-change-capability=2;max-red=0
```

The SDP layer parses AMR-WB parameters into typed `FormatParams`. Compatibility matching then accepts only configurations the packet layer can actually process. Parameters that change the wire format must match and are echoed unchanged in the answer.

Unsupported or malformed configurations are declined rather than negotiated and misdecoded. This includes CRC, interleaving, robust sorting, multiple channels, restricted mode sets, and unsupported mode-change constraints.

### Send path

Each `Writer::write` call must contain **exactly one 3GPP IF frame** and one media timestamp. AMR-WB frames represent 20 ms, or 320 ticks at 16 kHz, so accepting concatenated frames would make timestamp ownership ambiguous.

The packetizer:

- validates the IF header, frame type, encoded length, and MTU;
- rejects reserved frame types, concatenated frames, and standalone `NO_DATA`;
- emits exactly one RFC 4867 RTP payload;
- uses either octet-aligned or bandwidth-efficient packing according to SDP;
- emits CMR 15, meaning no codec mode request.

The generic RTP layer supplies sequence numbers, the rebased 16 kHz timestamp, extensions, and the talkspurt marker.

### Receive path

Incoming RTP is mapped to the negotiated payload type, and the depacketizer is configured from the same `octet-align` setting as the sender.

Unlike the send API, the receiver accepts valid compound RFC 4867 payloads because remote peers may place multiple ToC entries in one RTP packet. These are returned as concatenated IF frames in `Event::MediaData`.

The receiver rejects reserved frame types, truncated speech data, unexpected trailing octets, and nonzero terminal padding.

### Code map

- SDP parsing and serialization: data.rs
- Typed codec parameters: format_params.rs
- Compatibility and answer shaping: payload_params.rs
- RFC 4867 conversion: amr_wb.rs
- Runtime packetizer dispatch: mod.rs
- Send configuration: payload.rs
- Receive configuration: mod.rs
- End-to-end coverage: amr-wb.rs

A useful review order is: **SDP compatibility rules → packet conversion → runtime wiring → integration and malformed-input tests**.

### DIFF
--- a/src/format/codec.rs
+++ b/src/format/codec.rs
@@ -53,6 +53,12 @@ pub enum Codec {
     PCMU,
     PCMA,
     G722,
+    /// AMR Wideband (RFC 4867), 16 kHz mono telephony/IMS audio.
+    ///
+    /// Each call to [`Writer::write`](crate::media::Writer::write) must contain
+    /// exactly one 3GPP IF frame. Concatenated frames are rejected because each
+    /// frame requires its own RTP timestamp.
+    AmrWb,
     H264,
     // TODO show this when we support h265.
     #[doc(hidden)]
@@ -82,7 +88,7 @@ impl Codec {
     /// Tells if codec is audio.
     pub fn is_audio(&self) -> bool {
         use Codec::*;
-        matches!(self, Opus | PCMU | PCMA | G722)
+        matches!(self, Opus | PCMU | PCMA | G722 | AmrWb)
     }
 
     /// Tells if codec is video.
@@ -109,6 +115,7 @@ impl<'a> From<&'a str> for Codec {
             "pcmu" => Codec::PCMU,
             "pcma" => Codec::PCMA,
             "g722" => Codec::G722,
+            "amr-wb" => Codec::AmrWb,
             "h264" => Codec::H264,
             "h265" => Codec::H265,
             "h266" => Codec::H266,
@@ -128,6 +135,7 @@ impl fmt::Display for Codec {
             Codec::PCMU => write!(f, "PCMU"),
             Codec::PCMA => write!(f, "PCMA"),
             Codec::G722 => write!(f, "G722"),
+            Codec::AmrWb => write!(f, "AMR-WB"),
             Codec::H264 => write!(f, "H264"),
             Codec::H265 => write!(f, "H265"),
             Codec::H266 => write!(f, "H266"),
--- a/src/format/codec.rs
+++ b/src/format/codec.rs
@@ -55,6 +55,7 @@ pub enum Codec {
     G722,
     /// Comfort Noise payload, per RFC 3389.
     CN,
+    AmrWb,
     H264,
     // TODO show this when we support h265.
     #[doc(hidden)]
@@ -88,7 +89,7 @@ impl Codec {
     /// Tells if codec is audio.
     pub fn is_audio(&self) -> bool {
         use Codec::*;
-        matches!(self, Opus | PCMU | PCMA | G722 | CN)
+        matches!(self, Opus | PCMU | PCMA | G722 | CN | AmrWb)
     }
 
     /// Tells if codec is video.
@@ -116,6 +117,7 @@ impl<'a> From<&'a str> for Codec {
             "pcma" => Codec::PCMA,
             "g722" => Codec::G722,
             "cn" => Codec::CN,
+            "amr-wb" => Codec::AmrWb,
             "h264" => Codec::H264,
             "h265" => Codec::H265,
             "h266" => Codec::H266,
@@ -137,6 +139,7 @@ impl fmt::Display for Codec {
             Codec::PCMA => write!(f, "PCMA"),
             Codec::G722 => write!(f, "G722"),
             Codec::CN => write!(f, "CN"),
+            Codec::AmrWb => write!(f, "AMR-WB"),
             Codec::H264 => write!(f, "H264"),
             Codec::H265 => write!(f, "H265"),
             Codec::H266 => write!(f, "H266"),
--- a/src/format/format_params.rs
+++ b/src/format/format_params.rs
@@ -108,6 +108,68 @@ pub struct FormatParams {
     /// When > 0, DONL fields are included in H.265 RTP packets to support
     /// out-of-order NAL unit decoding.
     pub sprop_max_don_diff: Option<u16>,
+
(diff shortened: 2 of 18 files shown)


src/format/codec.rs
```diff
@@ -31,6 +31,8 @@ pub enum Codec {
     Opus,
     PCMU,
     PCMA,
+    /// AMR Wideband (RFC 4867), 16 kHz mono telephony/IMS audio.
```
xnorpx: remove comments

src/format/format_params.rs
```diff
@@ -108,6 +108,43 @@ pub struct FormatParams {
...
+    pub mode_change_capability: Option<u8>,
+
+    /// AMR-WB specific parameter (RFC 4867).
+    ///
+    /// The maximum amount of frame redundancy, in milliseconds, the receiver
+    /// accepts. `0` disables redundancy.
+    pub max_red: Option<u32>,
+
+    /// AMR-WB specific parameter (RFC 4867).
+    ///
```
xnorpx, line 151: make it clearer this is just for parsing and we don't support it

src/format/payload_params.rs
```diff
@@ -285,7 +285,7 @@ impl PayloadParams {
             return None;
         }
 
-        if c0.channels != c1.channels {
+        if !Self::channels_match(c0, c1) {
```
xnorpx: not related, leave alone

src/format/payload_params.rs
```diff
@@ -363,6 +367,18 @@ impl PayloadParams {
         score
     }
 
+    fn channels_match(c0: CodecSpec, c1: CodecSpec) -> bool {
+        if c0.channels == c1.channels {
+            return true;
+        }
+
+        if c0.codec == Codec::AmrWb && c1.codec == Codec::AmrWb {
```
xnorpx: no, this is weird.

src/format/payload_params.rs
```diff
@@ -406,6 +422,37 @@ impl PayloadParams {
...
+        }
+
+        // str0m can neither generate nor decode the per-frame payload CRC that
+        // crc=1 adds (RFC 4867 §4.3). Rather than agree to it and then mis-decode
+        // the stream, decline any peer that requires it. The default is no CRC.
+        if c0.format.crc.unwrap_or(false) || c1.format.crc.unwrap_or(false) {
+            return None;
+        }
+
+        // Interleaving and robust sorting add RTP payload fields/layouts that
```
xnorpx: put these higher up?

src/rtp/mtime.rs
```diff
@@ -48,6 +48,9 @@ impl Frequency {
     /// Cycles in a second of a 48 kHz signal.
     pub const FORTY_EIGHT_KHZ: Frequency = Self::make(48_000);
 
+    /// Cycles in a second of a 16 kHz signal.
+    pub const SIXTEEN_KHZ: Frequency = Self::make(16_000);
```
xnorpx, line 57: will be available in G.722

src/format/codec.rs
xnorpx: Removed the Codec::AmrWb variant comments in 18e1f2ce.

src/format/format_params.rs
xnorpx: Clarified that max-red is parsed and serialized for negotiation only; str0m does not implement AMR-WB frame redundancy.

src/format/payload_params.rs
xnorpx: Restored the generic channel comparison to direct equality. AMR-WB now routes to its codec-specific matcher before that unchanged check.

src/format/payload_params.rs
xnorpx: Removed channels_match; the omitted-channel-is-mono rule now stays entirely inside AMR-WB matching.

src/format/payload_params.rs
xnorpx: Moved the unsupported interleaving and robust-sorting checks to the top of AMR-WB matching.

src/rtp/mtime.rs
xnorpx: Confirmed. After rebasing, SIXTEEN_KHZ is provided by the G.722 work on main; this PR only extends its documentation for AMR-WB.

PR #999, opened 2026-07-01
vp9: don't emit an R=0 GOF in non-flexible SS for single-layer streams

### LINKED ISSUE #998: VP9 non-flexible packetizer emits an R=0 GOF for single-layer streams, corrupting picture on libwebrtc receivers
In non-flexible mode the keyframe SS advertises a GOF whose one entry has `R=0` (references 0 pictures), while inter frames are marked `P=1`. Single spatial + temporal layer, IPPP from libvpx.

`src/packet/vp9.rs`, keyframe first packet:
```
0x08  N_S=0, Y=0, G=1
0x01  N_G=1
0x00  GOF[0]: R=0
```

libwebrtc's `ParseSsData` treats the GOF as authoritative → concludes frames reference nothing → mis-tracks references → full-frame corruption even at 100% delivery (Chrome shows no loss).

It should be `G=0` (no GOF) for a single-layer stream — what libwebrtc itself sends (`g_bit = num_frames_in_gof > 0`); the receiver then derives refs from `P` + `tl0picidx`. Flexible mode avoids it but carries no reference info.

AdirAmsalem:
Fixes #998.

The non-flexible keyframe SS was `G=1` with a 1-entry GOF at `R=0` — telling receivers inter frames reference nothing, which contradicts `P=1` and the bitstream. libwebrtc trusts the GOF and mis-tracks references → corruption even at 100% delivery.

Emit `G=0` (no GOF) instead — the single-layer form libwebrtc sends; the receiver then uses `P` + `tl0picidx`. SS shrinks 3 bytes → 1.

Tests updated (plus a depacketizer assertion on the new SS byte); full suite green. Verified in production — fixes the corruption and recovers cleanly under loss.

(`VP9_NON_FLEXIBLE_HEADER_SIZE` left at 8; now over-reserves by 2, can drop to 6 if you'd prefer.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
+
 # 0.21.0
 
   * Upgrade to  dimpl 0.7.0 (breaking)
--- a/src/config.rs
+++ b/src/config.rs
@@ -642,7 +642,7 @@ impl RtcConfig {
     ///
     /// VP9 supports two RTP payload descriptor formats:
     ///
-    /// * [`Vp9PacketizerMode::NonFlexible`] (default) — 5-8 byte header with layer indices
+    /// * [`Vp9PacketizerMode::NonFlexible`] (default) — 5-6 byte header with layer indices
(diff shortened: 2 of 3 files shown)


algesten (approved):

algesten:
Thanks!

PR #1000, opened 2026-07-03
Normalize RTC close draining

ThanhDodeurOdoo:
Summary

Normalize RTC closing so `RtcState::Closing` means the connection is draining local close output rather than waiting for peer shutdown replies. Closing keeps the regular poll path active only until local close output has drained.

`Rtc::close()` now schedules RTCP BYE, requests SCTP shutdown, and starts DTLS close_notify together. DTLS close is no longer delayed on SCTP reaching a closed state; if SCTP shutdown output can no longer be carried because DTLS is already closing, it is dropped during the close drain.

Incoming DTLS close_notify and SCTP association loss enter the same closing path, so str0m responds with local close signals and lets `poll_output()` drain them before `is_alive()` becomes false.

Schedule RTCP BYE for local RTP senders on close, drop pending media and ordinary feedback, and stop scheduling new RTP/RTCP work while preserving close-related output.

Use the published close-state predicates from `sctp-proto` and `dimpl` so str0m delegates SCTP and DTLS closing/closed semantics to the owning subsystems.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -572,9 +572,9 @@ dependencies = [
 
 [[package]]
 name = "dimpl"
-version = "0.7.0"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0a11091ebc139b18d6b5878d8769ca43ad5ea8a7e2f5ed1e1f25e172951ef0ce"
+checksum = "bb2b8390bf0021eae92cf96a0e3a54c531b6315c47edfe9220f5fc61bc9e1f63"
 dependencies = [
  "aes",
  "aes-gcm",
@@ -1715,9 +1715,9 @@ dependencies = [
 
 [[package]]
 name = "sctp-proto"
-version = "0.10.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5376eaf8a764118abd6bee29673c68eab91f0913b448fe2944e74488b63c37b5"
+checksum = "199c5c38008c2c151e27afd228230b1b69f849b1f7629f5df86662ee1f456187"
 dependencies = [
  "bytes",
  "crc",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,12 +53,12 @@ _str0m_test = { path = "crates/_str0m_test" }
 # Core dependencies
 tracing = "0.1.37"
 fastrand = "2.0.1"
-sctp-proto = "0.10.0"
+sctp-proto = "0.10.1"
 combine = "4.6.6"
 subtle = "2.0.0"
 arrayvec = "0.7.6"
 serde = { version = "1.0.152", features = ["derive"] }
-dimpl = { version = "0.7.0", default-features = false }
+dimpl = { version = "0.7.1", default-features = false }
 drv = { version = "0.4.3" }
 time = "0.3"
 base64ct = "1"
--- a/crates/proto/src/crypto/provider.rs
+++ b/crates/proto/src/crypto/provider.rs
@@ -193,6 +193,16 @@ pub trait DtlsInstance: CryptoSafe {
     /// Return the negotiated DTLS protocol version. May return `None` before handshake completion.
     fn protocol_version(&self) -> Option<ProtocolVersion>;
 
+    /// Return true after close has started and while close output may still need polling.
+    fn is_closing(&self) -> bool {
+        false
+    }
+
+    /// Return true once close output has drained and the instance is closed.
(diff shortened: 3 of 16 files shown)


lolgesten:
@ThanhDodeurOdoo did you do a comparison pre- and post this PR?

ThanhDodeurOdoo:
@algesten yes I did, the commit completely reverts the perf loss

comparison: 0.21.0 -> 056e53569e3472ca74a5db1bc72f7c1d4e286211

| scenario | before instructions | after instructions | instruction change | before cycles | after cycles | cycle change | before read+write | after read+write | read+write change |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 1 session | ~3.9K | ~3.8K | -2.87% | ~15.8K | ~15.6K | -1.06% | ~5.8K | ~5.6K | -2.99% |
| 128 sessions | ~916.2K | ~901.7K | -1.58% | ~1.41M | ~1.39M | -1.67% | ~1.31M | ~1.29M | -1.68% |

xnorpx:
Congrats on #1000 !

algesten:
Original PR message before updating the PR scope:

Title: perf: avoid double lifecycle checks

Body:

Close-notify added a closing flag next to alive. Idle Rtc::poll_output() then paid two lifecycle checks even thou shutdown is rare

Replace both booleans with one enum so the hot poll path checks for Alive once.

Benchmarks:

tool: Valgrind
settings: jemalloc, 128 idle RTP-mode sessions, 1 drain round per run
comparison: 0.20.0 -> 0.21.0

| Rtc count | before instructions | after instructions | instruction change |
|---|---:|---:|---:|
| 1 | ~3.8K | ~3.9K | +3.20% | 
| 128 | ~900.5K | ~916.2K | +1.73% | 

| Rtc count | before cycles | after cycles | cycle change 
|---|---:|---:|---:|
| 1 | ~15.5K | ~15.9K | +2.30% |
| 128 | ~1.38M | ~1.41M | +2.33% |

| Rtc count | before read+write | after read+write | read+write change |
|---|---:|---:|---:|
| 1 | ~5.6K | ~5.8K | +3.10% |
| 128 | ~1.29M | ~1.31M | +1.72% |

(the commit removes the regression)

algesten:
@ThanhDodeurOdoo no fault of yours, but I felt the previous close weren't digging deep enough. I changed this PR to keep your fixes, but also normalize the closing behavior for SCTP and DTLS.

algesten (approved):

algesten:
@xnorpx any thoughts?

ThanhDodeurOdoo:
@algesten it looks great, there is just one part of the fix that was removed which had a good share of the performance optimisation: closing still needs to stay off the "hot" Alive path, otherwise evry idle poll_output() reaches the timeout tail and pays the Closing drain check even though it only maters during shutdown

roughly something like:

```rs
    fn do_poll_output(&mut self) -> Result<Output, RtcError> {
        if self.state != RtcState::Alive {
            return self.poll_inactive_output();
        }

        self.do_poll_output_inner()
    }

    #[cold]
    #[inline(never)]
    fn poll_inactive_output(&mut self) -> Result<> {
        if self.state == RtcState::Closed {
            self.last_timeout_reason = Reason::NotHappening;
            return Ok(Output::Timeout(not_happening()));
        }

        let out = self.do_poll_output_inner()?;

        if matches!(out, Output::Timeout)_)) && self.close_drain_complete() {
            //...
            return Ok(Output::Timeout(not_happening()));
        }

        Ok(out)
    }
    ```

algesten:
@ThanhDodeurOdoo going back to your original fix. It doesn't seem reasonable that it could affect performance much. It's replacing checking two booleans with one enum check.

Or am I misunderstanding the fix?

ThanhDodeurOdoo:
@algesten yes my commit message wasn't very good, the enum is more idiomatic to represent the state and it also indeed allows for a single check (which is also a small improvement) but the main performance gain is that close-drain logic stoped being part of the normal Alive -> Timeout main path

in the 0.21 shape, every idle alive poll_output() still reaches code that check if shutdown/drain handling applies

algesten:
I don't think this follows.

`Alive` hits the branch, but it doesn't call `close_drain_complete()`. So the cost is one enum check.

ThanhDodeurOdoo:
@algesten it's mainly that we still check state == Closing at the end of every idle poll, the fix moves closing/closed handling in an early cold branch, so the normal alive poll path does not have shutdown logic 

edit: chatted a bit on discord, it's not blocking for me, the small optimisation may not be worth the noise that it would add

PR #1006, opened 2026-07-11
Cache per-stream timeout state

algesten:
## Summary

Cache timeout state within `StreamTx` and `StreamRx`, invalidating it at mutable entry points. Keep the simple stream loops while using forced-inline deadline checks to skip child timeout work until a stream is dirty or due, preserving grouped RTCP reports and NACK processing.

Addresses the stream timeout traversal identified in #922.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Cache per-stream timeout state to skip idle timeout work #1006
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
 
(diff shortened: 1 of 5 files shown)


algesten:
Hey @lherman-cs - here's a smaller fix for the performance issues you are addressing with #984

Does this fix the problem?

algesten:
Excellent insight by Tekyo on Discord:

https://discord.com/channels/1436725650302959829/1525535885712429127/1525584930875244754

> ...the remaining poll_event scans (the four find_maps + the two medias
> loops) are untouched by #1006, and that's still ~20% of our thread.

The long story short: on a 30 participant SFU, the `values_mut().find_map()` dominates, meaning this PR in its current state is not enough. The work needs to short circuit higher up, more along #984

algesten:
Superseded by #1008

PR #1008, opened 2026-07-12
Replace timeout scans with a global scheduler

algesten:
This replaces repeated per-stream and per-media timeout scans in the normal `poll_output()` path with one global scheduler, while preserving the existing public timeout API.

## Architecture

`Rtc` now owns a global `Scheduler` keyed by private, precise `Timer` identities.

- Singleton timers use fixed slots; stream timers carry an SSRC and media packetization timers carry a MID.
- `Reason`, `Rtc::last_timeout_reason()`, and the public `Pacer::poll_timeout()` contract remain unchanged.
- Mutation paths append the smallest affected owner to a small, linearly deduplicated invalidation list. Before returning the next timeout, the scheduler recomputes only those scopes.
- Writer operations invalidate timers only after successful mutations: media writes invalidate their MID, while keyframe requests invalidate the exact receive-stream SSRC.
- Rearming an unchanged scoped timer is a no-op, avoiding unnecessary sorted-queue churn.
- Due timers are consumed through an iterator borrowing the scheduler, without a temporary allocation.
- Timer dispatch uses the precise identity to reach the affected component or stream. The SSRC-reset path preserves the invariant that the `StreamTx` map key equals `StreamTx::ssrc()`.
- Removed entities need no tombstone bookkeeping. A stale timer can cause an early wake, but exact lookup finds no entity and the timer disappears when consumed.

The pacer retains one internal timer identity and separately preserves its public `PacerReason` diagnostic. The three BWE deadline sources remain distinct timers while keeping the existing coupled BWE process step when any one fires.

## Correctness oracle

Integration-test builds retain an independent exhaustive O(N) timeout calculation. At every completed poll-to-timeout drain it asserts that:

- every live timer found by the exhaustive calculation exists in the production scheduler; and
- the production deadline is no later than the exhaustive deadline.

Extra production entries are allowed, which supports removal without explicit disarming. The scheduler also has focused coverage for move-on-rearm behavior, unchanged rearming, pacer sub-reasons, borrowed due polling, exact keyframe invalidation, and the deduplicated LIFO invalidation list.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -687,6 +687,10 @@ use streams::RtpPacket;
 use streams::StreamPaused;
 use util::InstantExt;
 
+mod scheduler;
+pub use scheduler::Timer;
+use scheduler::{Scheduler, TimeoutOwner};
+
 // Identity `drv::ToStatic` (`Static = Self`) for the `Copy` identity types.
 // Used instead of `#[derive(drv::Input)]` because the derive's generated
 // shadow type isn't `Eq`/`Hash`/`Borrow<Self>`, so it can't be a `HashMap`
@@ -837,7 +841,7 @@ use media::{MediaAdded, MediaChanged, MediaData};
 pub mod change;
 
 mod util;
-use util::{Soonest, not_happening};
+use util::{already_happened, not_happening};
 
 mod session;
 use session::Session;
@@ -917,7 +921,8 @@ pub struct Rtc {
     peer_bytes_rx: u64,
     peer_bytes_tx: u64,
     change_counter: usize,
-    last_timeout_reason: Reason,
+    scheduler: Scheduler,
+    last_timeout: Option<Timer>,
     crypto_provider: Arc<crate::crypto::CryptoProvider>,
     fingerprint_verification: bool,
     close_dtls_started: bool,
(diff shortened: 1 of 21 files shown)


algesten:
@xnorpx @k0nserv or @davibe a review?

src/scheduler.rs
```diff
@@ -0,0 +1,691 @@
...
+        if self.singletons.arm(timer, at) {
+            return;
+        }
+
+        if let Some(index) = self
+            .queued
+            .iter()
+            .position(|&(_, queued_timer)| queued_timer == timer)
+        {
+            if self.queued[index].0 == at {
```
k0nserv, line 295: Is this liable to ever happen? Don't we avoid this case because of this PR?

src/lib.rs
```diff
@@ -1826,29 +1763,29 @@ impl Rtc {
...
     /// let rtc = Rtc::new(Instant::now());
     ///
-    /// // Before any call to poll_output(), the reason is the default.
-    /// assert_eq!(rtc.last_timeout_reason(), Reason::NotHappening);
+    /// // Before any call to poll_output(), there is no timer.
+    /// assert_eq!(rtc.last_timeout(), None);
     /// ```
-    pub fn last_timeout_reason(&self) -> Reason {
-        self.last_timeout_reason
+    pub fn last_timeout(&self) -> Option<Timer> {
```
k0nserv: Is this the only breaking change, if so maybe we should keep the previous API surface here

src/pacer/leaky.rs
```diff
@@ -489,6 +492,12 @@ mod test {
     use queue::{PacketKind, Queue, QueuedPacket};
     use std::time::{Duration, Instant};
 
+    fn pacer_timeout(pacer: &impl Pacer) -> Option<Instant> {
```
k0nserv: This is test only so should be `#[cfg(test)]`

k0nserv (approved):
The performance improvement looks great!

I don't love the approach, but I also don't have a better idea myself. I do think avoiding the public API change would be good, this way we can continue to iterate on this.


## What I don't like

If you fail to invalidate it will affect correctness. It would be better if failing to invalidate only caused a performance regression, while maintaining correctness. Because it's the former not the latter, we have many potential new footguns.

Before knowledge about timeouts was scoped to each component, now it leaks out and is spread in several places.

src/lib.rs
```diff
@@ -1466,6 +1394,7 @@ impl Rtc {
         // This does not catch potential RIDs required to send simulcast, but
         // it's a good start. An error might arise later on RID mismatch.
         self.session.media_by_mid_mut(mid)?;
+        self.scheduler.invalidate(TimerScope::Media(mid));
```
k0nserv: Don't we also need to invalidate the keyframe request scope here?

algesten:
> ## What I don't like
> 
> If you fail to invalidate it will affect correctness. It would be better if failing to invalidate only caused a performance regression, while maintaining correctness. Because it's the former not the latter, we have many potential new footguns.
> 
> Before knowledge about timeouts was scoped to each component, now it leaks out and is spread in several places.

I 100% agree with you there. This is why I been so slow at accepting #984 - what we have is so clean in terms of not locking up str0m if we fuck up. However even #1006 is not enough to get the performance, so I'm slowly accepting the fact that we MUST do it and rely on tests etc to avoid problems.

src/scheduler.rs
algesten: Yes. Recomputing an invalidated scope can re-arm an unchanged timer at the same deadline, so this is expected. I added a comment and a regression test showing that it is a no-op.

src/lib.rs
algesten: Restored the previous public surface: `Reason` and `Rtc::last_timeout_reason()` are unchanged, `Timer` is private, and the public `Pacer::poll_timeout()` contract is preserved.

src/lib.rs
algesten: Yes. `request_keyframe()` now captures the exact receive-stream SSRC and invalidates `TimerScope::StreamRx(ssrc)` after the successful request. I also added a regression asserting that the feedback timeout is visible immediately.

src/pacer/leaky.rs
algesten: The helper is gone now that the public `Pacer::poll_timeout()` API has been restored; these tests call that API directly again.

lherman-cs:
> > ## What I don't like
> > 
> > If you fail to invalidate it will affect correctness. It would be better if failing to invalidate only caused a performance regression, while maintaining correctness. Because it's the former not the latter, we have many potential new footguns.
> > 
> > Before knowledge about timeouts was scoped to each component, now it leaks out and is spread in several places.
> 
> I 100% agree with you there. This is why I been so slow at accepting #984 - what we have is so clean in terms of not locking up str0m if we fuck up. However even #1006 is not enough to get the performance, so I'm slowly accepting the fact that we MUST do it and rely on tests etc to avoid problems.

It's unfortunate, I don't like the brittleness either here. Tbh, I started with caching the timeouts too. But, I couldn't figure out how to make the change impactful to real world performance. At the end, I just went back to this binary heap idea again..

Lanius-collaris:
@algesten @lherman-cs 
Smol's reactor uses b-tree to store sorted time points, is it better?
https://docs.rs/async-io/latest/src/async_io/reactor.rs.html#82
https://docs.rs/async-io/latest/src/async_io/reactor.rs.html#204

lherman-cs:
> @algesten @lherman-cs Smol's reactor uses b-tree to store sorted time points, is it better? https://docs.rs/async-io/latest/src/async_io/reactor.rs.html#82 https://docs.rs/async-io/latest/src/async_io/reactor.rs.html#204

@Lanius-collaris A `BTreeMap` would be similar to a `BinaryHeap`; the data structure is mostly orthogonal to the concern here.

The question is whether we can keep the laziness and event-driven benefits of a central scheduler, while avoiding the footgun where a subsystem changes its deadline but forgets to update the scheduler.

There's this effort, https://github.com/algesten/str0m/pull/1011, but I unfortunately don't see CPU improvements in my case.

PR #1010, opened 2026-07-13
fix: streams reset handshake, make stream ids safely reusable

gab8i:
Closing a data channel was local-only, the remote never got `ChannelClose`, and the id stayed occupied on the wire, reusing it hit sctp-proto's `ErrStreamAlreadyExist` / `ErrStreamResetPending`, treated as fatal, so the new channel died instantly (seen in litep2p under churn).

fix:
- `close_stream()` closes the underlying stream, sending the reset, so both sides close and free the id.
- `StreamEvent::Finished` classification:
   - if live stream then the remote-initiated reset
   - otherwise completion of our own reset, which implies the internal emission of `SctpEvent::StreamResetComplete`.
- The allocator holds a closed id until reset complete or a 30s max-hold timer exceeds.
- Transient open errors retry for up to 10s instead of failing, driven by the packet that resolves them or a poll_timeout backstop 

The cooldown approach is being discarded because a stream ID can be re-used once the reset handshake has been completed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,8 +53,7 @@ _str0m_test = { path = "crates/_str0m_test" }
 # Core dependencies
 tracing = "0.1.37"
 fastrand = "2.0.1"
-# sctp-proto = "0.10.1"
-sctp-proto = { path = "../sctp-proto" }
+sctp-proto = "0.10.2"
 combine = "4.6.6"
 subtle = "2.0.0"
 arrayvec = "0.7.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,8 @@ _str0m_test = { path = "crates/_str0m_test" }
 # Core dependencies
 tracing = "0.1.37"
 fastrand = "2.0.1"
-sctp-proto = "0.10.1"
+# sctp-proto = "0.10.1"
+sctp-proto = { path = "../sctp-proto" }
 combine = "4.6.6"
 subtle = "2.0.0"
 arrayvec = "0.7.6"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,7 +53,7 @@ _str0m_test = { path = "crates/_str0m_test" }
 # Core dependencies
 tracing = "0.1.37"
 fastrand = "2.0.1"
-sctp-proto = "0.10.1"
+sctp-proto = "0.10.3"
 combine = "4.6.6"
 subtle = "2.0.0"
 arrayvec = "0.7.6"
--- a/src/channel.rs
+++ b/src/channel.rs
@@ -173,9 +173,18 @@ impl fmt::Debug for ChannelData {
 pub(crate) struct ChannelHandler {
     allocations: Vec<ChannelAllocation>,
     next_channel_id: usize,
-    /// Stream IDs recently closed, with the time they were closed.
-    /// Excluded from allocation until the cooldown expires.
-    closed_stream_ids: Vec<(u16, Instant)>,
+    /// Stream IDs of closed channels, excluded from allocation until the
+    /// reset handshake completes or the max-hold expires.
+    closed_stream_ids: Vec<ClosedStreamId>,
+}
+
+/// A stream ID held back from reallocation after its channel closed.
+#[derive(Debug)]
+struct ClosedStreamId {
+    stream_id: u16,
+    /// When the hold started. Only used for the max-hold fallback,
+    /// normal release is the reset-completion signal from SCTP.
+    held_since: Instant,
 }
 
 #[derive(Debug)]
@@ -190,7 +199,12 @@ struct ChannelAllocation {
     config: Option<ChannelConfig>,
 }
 
-const STREAM_ID_COOLDOWN: Duration = Duration::from_secs(2);
+/// Upper bound on how long a stream ID is held when the reset handshake never
+/// completes. Prevents leaking IDs forever.
+///
+/// The normal release path is the completion signal itself, the SCTP layer
+/// retries transient open errors.
+const STREAM_ID_MAX_HOLD: Duration = Duration::from_secs(30);
 
 impl ChannelHandler {
     pub fn new_channel(&mut self, config: &ChannelConfig) -> ChannelId {
@@ -292,7 +306,7 @@ impl ChannelHandler {
             .allocations
             .iter()
             .filter_map(|a| a.sctp_stream_id)
-            .chain(self.closed_stream_ids.iter().map(|(id, _)| *id))
+            .chain(self.closed_stream_ids.iter().map(|c| c.stream_id))
             .collect();
 
         for a in &mut self.allocations {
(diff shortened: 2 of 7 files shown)


src/sctp/mod.rs
```diff
@@ -809,25 +862,52 @@ impl RtcSctp {
...
+                    ) => {
                         let config = entry.config.as_ref().expect("config if AwaitOpen");
                         let in_band = config.negotiated.is_none();
 
-                        if in_band {
-                            warn!(
-                                "Opening stream {} failed: ErrStreamAlreadyExists with in-band",
-                                entry.id
-                            );
+                        if in_band || matches!(e, ProtoError::ErrStreamResetPending) {
```
algesten: Can we avoid increased indentation with an early return here?

src/sctp/mod.rs
```diff
@@ -809,25 +862,52 @@ impl RtcSctp {
...
+                            let deadline = *entry
+                                .open_deadline
+                                .get_or_insert(self.last_now + STREAM_OPEN_TIMEOUT);
+
+                            if self.last_now < deadline {
+                                debug!("Stream {} open blocked ({:?}), will retry", entry.id, e);
+                                continue;
+                            }
+
+                            warn!("Opening stream {} failed after retries: {:?}", entry.id, e);
```
algesten: Is this really warn? If the remote peer can cause this, it's not something we should warn. `debug!` probably since I tend to get angry users when I use too much `info!`

src/sctp/mod.rs
```diff
@@ -247,6 +260,17 @@ impl StreamEntry {
 /// chunk header + some extra.
 const SCTP_OVERHEAD: usize = 40;
 
+/// How long `open_stream` keeps retrying when blocked by a transient error
+/// which could happen during reset handshake (RFC 6525).
+/// Past this the peer is considered broken and the open gives up.
+const STREAM_OPEN_TIMEOUT: Duration = Duration::from_secs(10);
```
algesten, line 271: Should this be in sctp-proto crate?

algesten (commented):
Hi! Looks good!

Overall I wonder if some of this belong in the sctp-proto crate?

src/sctp/mod.rs
gab8i: Definitely!

src/sctp/mod.rs
gab8i: Yes, debug is also fair, this can be caused by many issue, the common ground is though: nothing arrived after X seconds since the reset handshake has been started. This can be not something done by the peer, but probably also a local issue with the network. 

A `warn` seemed the right thing here because closing a stream, if the connection is healthy, should never fail thus being explicit with the user seemed to make sense. But using `debug` is also valid!

src/sctp/mod.rs
gab8i: Maybe not this part, because this is related to how to deal with opening streams, but definitely issuing another event rather than `StreamEvent::Finished` and extract to meaning based on the stream state could be moved to sctp-proto. I will try to handle and move this into `sctp-proto` than!

algesten: > I will try to handle and move this into `sctp-proto` than!

Excellent! Let's see how much of this shifts into `sctp-proto`

gab8i: Ok, I've just open an `sctp-proto` pr and commit here the adaptation to the new added event. Modifying sctp-proto was the right thing to do, it also unveiled some hidden bugs within the reset procedure! 

Last commit doesn't compile, waiting for a new `sctp-proto` release now!

gab8i: Hello! Did you find some time to review https://github.com/algesten/sctp-proto/pull/53 ? Or the fix is too complicated which makes it difficult to understand? Let me know if I should refactor some things!

algesten: @gab8i there was a CI error, so I parked it, but I guess it was just formatting errors. I'll take a look.

gab8i:
Waiting for a new `sctp-proto` release so I can push updated version here!

algesten:
@gab8i 0.10.2 is on crates.io now.

gab8i:
hello! Any update @algesten ? Anything that you didn't like about the implementation? I think I've tackled most of the issue of the first iteration!

algesten:
I found some gaps in sctp-proto and here. Illustrated as failing tests. https://github.com/algesten/sctp-proto/pull/54

gab8i:
Thanks for having added those tests! Solving them gave the opportunity to notice other problems, I've described them in the commit description!

algesten:
@gab8i I made additional fixes in https://github.com/algesten/sctp-proto/pull/54 - should I land that and release sctp-proto patch?

gab8i:
Ok great, land that so I can re-work this one if needed! It would be great if you can give some feedback over the last commit in the mean while!

bkchr:
> @gab8i I made additional fixes in [algesten/sctp-proto#54](https://github.com/algesten/sctp-proto/pull/54) - should I land that and release sctp-proto patch?

@algesten could you move ahead with this?

algesten:
@bkchr @gab8i the sctp-proto 0.10.3 is released. Was more work to finish than I thought.

lexnv:
@algesten Thanks a lot for tackling that! Could you have another look over this PR? Ideally we can merge it + sctp-proto bump in a new release? 🙏

algesten:
Can you bump to 0.10.3 in this PR please so we can consider them in tandem.

algesten:
I found a couple of problems.

src/channel.rs
lexnv:
```suggestion
// Stream ID must be known before the config is taken. 
let Some(sctp_stream_id) = a.sctp_stream_id else {
continue;
};

let Some(config) = a.config.take() else {
continue;
};

```

src/channel.rs
```diff
@@ -365,21 +395,30 @@ impl ChannelHandler {
...
     }
 
-    /// Remove stream IDs from the cooldown list that have expired.
-    pub fn expire_closed_stream_ids(&mut self, now: Instant) {
-        self.closed_stream_ids
-            .retain(|(_, closed_at)| now.duration_since(*closed_at) < STREAM_ID_COOLDOWN);
+    /// The reset handshake for a held stream ID completed, release it for
+    /// reallocation immediately.
+    pub fn stream_reset_complete(&mut self, stream_id: u16) {
+        self.closed_stream_ids.retain(|c| *c != stream_id);
```
lexnv:
```suggestion
        if let Some(pos) = self.closed_stream_ids.iter().position(|c| *c == stream_id) {
            self.closed_stream_ids.remove(pos);
        }
```

src/channel.rs
```diff
@@ -419,39 +456,173 @@ mod tests {
...
+                .closed_stream_ids
+                .iter()
+                .any(|closed| allocation.sctp_stream_id == Some(*closed))
+        });
+
+        assert!(
+            !overlaps_held_generation,
+            "a negotiated ID must remain unavailable until its old reset completes"
+        );
+    }
```
lexnv, line 535: nit: No longer needed

```suggestion
```

Cargo.toml
```diff
@@ -53,7 +53,7 @@ _str0m_test = { path = "crates/_str0m_test" }
 # Core dependencies
 tracing = "0.1.37"
 fastrand = "2.0.1"
-sctp-proto = "0.10.1"
+sctp-proto = "0.10.2"
```
lexnv:
```suggestion
sctp-proto = "0.10.3"
```

lexnv:
@algesten Thanks a lot for the regression test! I've pushed my changes https://github.com/gab8i/str0m/pull/1 to Gab's branch since he's on holiday 

If they make sense could you either apply the git-diff suggestions or cherry pick them? 🙏 Thanks again!

algesten:
Thank you all!

lexnv:
Thanks a lot @algesten for tackling this and for the release🙏

PR #1011, opened 2026-07-14
Add readiness gate to make idle poll_output O(1)

k0nserv:
Everything below Rtc forms a tree; events and timeouts bubble up. On main every poll_output re-walks the whole tree (the .soonest() timeout fold plus poll_event over all medias/streams), so an idle poll that only returns Output::Timeout is O(number of m-lines). An SFU calls poll_output thousands of times per second per connection, paying that scan O(N) times.

Introduce a Wake guard threaded down the tree (src/poll.rs). It defaults to re-polling everything and is narrowed locally by the component that mutated (no_events / no_timeout), folding into an Rtc-owned Readiness on drop. poll_output gates on it: skip the event pipeline unless something queued output, skip the timeout recompute unless a timer moved, otherwise return the cached timeout. Forgetting or over-arming costs a redundant poll, never dropped output.

The mutation and handle_input paths pass their own guard to each subsystem so the invalidation decision stays local (session, chan, sctp, Writer, Bwe, Channel). External crates (ice, sctp-proto) can't take a Wake, so those boundaries arm conservatively; SdpApi/DirectApi still arm at the accessor pending threading.

tests/poll-scaling.rs measures it: idle poll_output is flat ~27ns from 2 to 50 m-lines here, versus 135ns->1038ns (7.7x) on main.

### DIFF
--- a/src/bwe/api.rs
+++ b/src/bwe/api.rs
@@ -1,5 +1,6 @@
 //! Bandwidth estimation.
 
+use crate::poll::RtcMut;
 use crate::{Rtc, rtp_::Mid};
 
 pub use crate::rtp_::Bitrate;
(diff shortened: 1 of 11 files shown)


k0nserv:
@copilot Review

PR #1013, opened 2026-07-18
Add support for Comfort Noise

### LINKED ISSUE #993: Add support for Comfort Noise
https://datatracker.ietf.org/doc/html/rfc3389

terminalchai:
Adds RFC 3389 Comfort Noise as a first-class audio codec. The opt-in helper configures static PT 13 at 8 kHz, while callers can configure 16/24/32/48 kHz with dynamic payload types through `CodecConfig::add_config`.

CN payloads pass through one per RTP packet and never set the marker bit. Coverage includes codec identity, configuration, payload round trips, and SDP negotiation at all five requested clock rates.

Closes #993

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
-  * Support RFC 3389 Comfort Noise payloads #993
+  * Support RFC 3389 Comfort Noise payloads #1013
   * Apply reliability parameters to a DCEP-receiving in-band channel #1004
 
 # 0.21.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
+  * Support RFC 3389 Comfort Noise payloads #993
   * Apply reliability parameters to a DCEP-receiving in-band channel #1004
 
 # 0.21.0
--- a/src/format/codec.rs
+++ b/src/format/codec.rs
@@ -53,6 +53,8 @@ pub enum Codec {
     PCMU,
     PCMA,
     G722,
+    /// Comfort Noise payload, per RFC 3389.
+    ComfortNoise,
     H264,
     // TODO show this when we support h265.
     #[doc(hidden)]
@@ -82,7 +84,7 @@ impl Codec {
     /// Tells if codec is audio.
     pub fn is_audio(&self) -> bool {
         use Codec::*;
-        matches!(self, Opus | PCMU | PCMA | G722)
+        matches!(self, Opus | PCMU | PCMA | G722 | ComfortNoise)
     }
 
     /// Tells if codec is video.
@@ -109,6 +111,7 @@ impl<'a> From<&'a str> for Codec {
             "pcmu" => Codec::PCMU,
             "pcma" => Codec::PCMA,
             "g722" => Codec::G722,
+            "cn" => Codec::ComfortNoise,
             "h264" => Codec::H264,
             "h265" => Codec::H265,
             "h266" => Codec::H266,
@@ -128,6 +131,7 @@ impl fmt::Display for Codec {
             Codec::PCMU => write!(f, "PCMU"),
             Codec::PCMA => write!(f, "PCMA"),
             Codec::G722 => write!(f, "G722"),
+            Codec::ComfortNoise => write!(f, "CN"),
             Codec::H264 => write!(f, "H264"),
             Codec::H265 => write!(f, "H265"),
             Codec::H266 => write!(f, "H266"),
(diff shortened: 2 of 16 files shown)


src/format/codec_config.rs
```diff
@@ -578,6 +601,34 @@ mod test {
     use super::*;
     use crate::format::{CodecSpec, FormatParams};
 
+    #[test]
```
xnorpx, line 699: add test for the other 4 as well

xnorpx: 16,24,32,48

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
+  * Support RFC 3389 Comfort Noise payloads #993
```
xnorpx: should be the pr here so 1013

tests/sdp-negotiation.rs
```diff
@@ -1413,6 +1447,19 @@ fn g722(pt: u8) -> PayloadParams {
     )
 }
 
+fn comfort_noise(pt: u8, clock_rate: u32) -> PayloadParams {
```
xnorpx: also add integration test both with sample/frame and rtp with all 5 CN

src/format/codec_config.rs
terminalchai: Added `comfort_noise_supports_dynamic_payload_types_at_higher_clock_rates`, covering dynamic PTs at 16, 24, 32, and 48 kHz in acd1877e.

CHANGELOG.md
terminalchai: Updated the changelog reference to this PR, `#1013`, in acd1877e.

tests/sdp-negotiation.rs
terminalchai: Added integration coverage for both paths in acd1877e: frame mode and raw RTP mode each negotiate and round-trip all five CN clock rates (8, 16, 24, 32, and 48 kHz).

tests/sdp-negotiation.rs
```diff
@@ -108,6 +108,117 @@ pub fn g722_offer_accepts_explicit_mono_channel() {
...
+                (97, 24_000),
+                (98, 32_000),
+                (99, 48_000),
+            ]
+        );
+    }
+}
+
+#[test]
+pub fn comfort_noise_frame_mode_round_trips_all_supported_clock_rates() -> Result<(), RtcError> {
```
xnorpx: Create a new file for these comfort_noise tests

terminalchai: Moved all three Comfort Noise integration tests and their helpers into tests/comfort-noise.rs in e10357d9. The SDP negotiation test file is back to its upstream contents.

xnorpx:
@algesten this looks ok to me, I let you take a pass as well.

terminalchai:
Implemented the regression cases added in 725f3241: CN payload validation, clock-rate-aware PT 13 handling, bare static CN SDP resolution, marker suppression, and sequence continuity across primary audio/CN payload switches. cargo test --all-targets and cargo clippy --lib -- -D warnings pass locally.

algesten:
@xnorpx @terminalchai My codex found one more, but I'm not convinced it is actually a problem. I can back it out depending on what you think.

terminalchai:
I think this is a real issue. A sustained switch to another payload type adds one padding entry per packet to the inactive depayloader, while pop() cannot drain a padding-only queue. The queue therefore grows until the original PT resumes. I'd keep the regression test. I'll implement bounded cleanup while preserving RTP sequence continuity and the configured reordering window.

terminalchai:
Fixed in 1eec66d6. Padding-only queues now retain only the configured reordering window, advance sequence state only across contiguous discarded padding, and preserve a discontinuity when a sequence number was lost. I also extended the regression coverage to verify both loss-free PT resumption and the loss case. cargo test --all-targets, cargo clippy --lib -- -D warnings, and cargo fmt --all -- --check pass locally.

algesten:
Found one more problem, with an unbounded situation. Failing test case pushed.

terminalchai:
Fixed in 3ce8971c. When an incomplete media fragment blocks the queue head, synthetic padding behind it is now trimmed to the configured hold-back window without dropping the fragment or advancing sequence state across it. The new regression passes, along with cargo test --all-targets, cargo clippy --lib -- -D warnings, and cargo fmt --all -- --check.

src/format/codec.rs
```diff
@@ -53,6 +53,8 @@ pub enum Codec {
     PCMU,
     PCMA,
     G722,
+    /// Comfort Noise payload, per RFC 3389.
+    ComfortNoise,
```
algesten: Rename this CN to follow pattern

src/format/codec_config.rs
```diff
@@ -569,6 +592,14 @@ impl Codec {
     }
 }
 
+impl CodecSpec {
+    pub(crate) fn default_pt(&self) -> Option<Pt> {
+        self.codec.default_pt().filter(|_| {
+            self.codec != Codec::ComfortNoise || self.clock_rate == Frequency::EIGHT_KHZ
+        })
+    }
+}
```
algesten, line 586: This seems very wrong. It overloads the `default_pt()` call with filter logic that I think is local to payload_params.rs. Separation of concerns.

algesten:
@terminalchai @xnorpx I'm going to add some review comments that are for myself. I'll land this PR from here.

algesten (commented):
(Comments for myself to fix)

algesten (approved):

Issue #1015, opened 2026-07-20
H264 negotiation with Safari/WebKit fails: higher offered levels reject the codec, and Constrained High is not recognized


dragona79:
## Summary

When `str0m` answers an SDP offer from a recent Safari browser (macOS/iOS, WebKit) using the default `enable_h264(true)` configuration, H264 is completely dropped from the negotiation. This results in an answer containing no H264 payloads at all. 

This behavior is caused by two independent bugs in the H264 matching logic, both of which are present in `0.21.0` and on the current `main` branch.

## Environment

* **Version:** str0m 0.21.0 (also verified against the latest `main` branch)
* **Role:** SDP Answerer (`accept_offer`), acting as a gateway/SFU that relays pre-encoded H264 without transcoding.

## Reproduction & Context

Safari's offer contains the following H264 `fmtp` lines:

```text
a=fmtp:96 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=640c34
a=fmtp:98 level-asymmetry-allowed=1;packetization-mode=1;profile-level-id=42e034
```

With the default `enable_h264(true)` registration, all local profiles are set to level 3.1 (e.g., `42e01f`). Because of the matching issues detailed below, neither of Safari's lines matches our local parameters, causing H264 to disappear from the answer entirely.

---

## Problem 1: Codec is rejected if the offered level is higher (Violates RFC 6184)

In `src/format/payload_params.rs`, the `match_h264_score` function contains the following logic:

```rust
// Reject if the offered level exceeds our configured capability
if c1_profile_level.level() > c0_profile_level.level() {
    return None;
}
```

Returning `None` here drops the codec from the negotiation entirely. However, under **RFC 6184 Section 8.2.2**, the level byte in `profile-level-id` is a **negotiable value**, not a hard matching criterion. This applies to both offer/answer branches:

* **When `level-asymmetry-allowed=1` is signaled by both sides** (which all major browsers do, and `str0m`'s own `add_h264` declares by default): Levels are per-direction. The answerer simply responds with its own receive level capability; a higher offered level is irrelevant to whether the codec matches.
* **When asymmetry is not negotiated:** The level to use for both directions falls back to the lower of the two default levels. The answerer responds with its own (lower) level, which satisfies the "no level upgrade" constraint.

[...]

PR #1016, opened 2026-07-21
Fix H264 negotiation: negotiable level and Constrained High profile

dragona79:
Two independent fixes for #1015:

* match_h264_score no longer rejects an offer whose level differs from the locally configured level. RFC 6184 section 8.2.2 makes the level a negotiable (downgradable) parameter, so the level difference now only penalizes the match score (abs_diff), mirroring how match_h265_score already works. This makes e.g. Safari's Constrained Baseline 5.2 (42e034) match a local 42e01f.

* Recognize the Constrained High profile (profile_idc 0x64 with constraint_set4_flag and constraint_set5_flag, profile-iop 0x0C). WebKit offers profile-level-id=640c34, which previously failed to parse and unconditionally removed the payload from negotiation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix H.264 negotiation: accept a negotiable (downgradable) level and recognize the Constrained High profile #1015
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Unreleased
 
-  * Fix H.264 negotiation: accept a negotiable (downgradable) level and recognize the Constrained High profile #1015
+  * Fix H.264 negotiation: accept a negotiable (downgradable) level and recognize the Constrained High profile #1016
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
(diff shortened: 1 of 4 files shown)


algesten:
Thanks for the PR!

The Safari case looks fixed. Two things before merge:

- Add an end-to-end SDP test using `enable_h264(true)` and Safari's exact `640c34`/`42e034` offer. The current tests only hit `match_h264_score()` and the parser, so they don't prove the original `accept_offer()` failure is fixed.
- Add a changelog row.

One adjacent issue. `match_h264_score()` now says any level combination is allowed, but `update_param()` only narrows the level for H265/H266. If local is 5.2 and the offer is 3.1 without asymmetry, we still answer 5.2. [RFC 6184 section 8.2.2](https://www.rfc-editor.org/rfc/rfc6184.html#section-8.2.2) says the answer must use the lower level in that case.

That bug predates this PR, so I don't think it should block the Safari fix. I'd either fix it here or narrow the comment and track it separately.

dragona79:
Added h264_offer_with_higher_level_negotiates — runs Safari's exact 640c34/42e034 offer through accept_offer() and asserts the m-line negotiates instead of being rejected. Confirmed it fails against the old reject logic.
Changelog row added.
Went with narrowing the comments: they now note the level is negotiable at match time only, and that update_param doesn't narrow the answered H.264 level like H.265/H.266 does

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix H.264 negotiation: accept a negotiable (downgradable) level and recognize the Constrained High profile #1015
```
algesten: Should say 1016

algesten (approved):

algesten:
Looks great, thanks! One tiny tweak and we merge.

CHANGELOG.md
victor-yjhong: 1015 is the issue number. Should I change it to the PR number?

algesten: Yes PR

algesten:
Thanks!

PR #1018, opened 2026-07-26
Expose bwe overusing flag

lherman-cs:
Distinguishes network congestion from ALR.

SFU and BWE bitrate calculations can diverge during VBR/screenshare. Without this flag, BWE misinterprets congestion as ALR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Expose bwe overusing flag #1018
   * Expose the latest send queue information on `StreamTx` #1005
   * Fix VP9 non-flexible SS advertising a bogus R=0 GOF for single-layer streams #999
   * Support G722 audio codec (16 kHz codec with 8 kHz RTP clock per RFC 3551) #992
--- a/src/bwe/api.rs
+++ b/src/bwe/api.rs
@@ -54,4 +54,15 @@ impl<'a> Bwe<'a> {
     pub fn reset(&mut self, init_bitrate: Bitrate) {
         self.0.session.reset_bwe(init_bitrate);
     }
+
+    /// Whether the delay-based controller currently signals overuse.
+    ///
+    /// This is `true` while the congestion detector observes a rising delay trend, i.e. the
+    /// link is genuinely congested rather than merely application-limited. It lets a bitrate
+    /// allocator tell apart an estimate that fell because the link shrank (must react) from
+    /// one that fell only because less was being sent (safe to hold), since both look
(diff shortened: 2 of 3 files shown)


algesten:
Spontaneously this sounds like a too low level detail to expose. The BWE is a series of interconnected subsignals that interact and counteract with each other. https://github.com/algesten/str0m/blob/main/docs/BWE.md

Does libwebrtc expose this?

lherman-cs:
> Spontaneously this sounds like a too low level detail to expose. The BWE is a series of interconnected subsignals that interact and counteract with each other. https://github.com/algesten/str0m/blob/main/docs/BWE.md
> 
> Does libwebrtc expose this?

It's not exposed. This is trying to solve a chicken-and-egg problem between allocation and BWE, so the allocator doesn't get stuck. I typically see these 2 cases:

1. Bootstrap: BWE says 500Kbps, but we want 1Mbps. So, the desired_bitrate is set to 1Mbps. But BWE never expands the available bandwidth. There's no RTX cache yet.
2. VBR: mismatch between BWE internal bitrate estimate and the allocator bitrate estimate. The allocator aims to be a conservatively higher and more stable bitrate. The BWE then thinks we're in ALR and can't grow the BWE available bandwidth to what the allocator wants. The allocator ends up downgrading, and eventually gets stuck in "paused".

So, I'm using this flag to basically always keep feeding the BWE with enough data to hit the desired bitrate. I'm experimenting with sending paddings to ssrc=0. It feels like if we can get the BWE available to the desired bandwidth (excluding congestion), this should be good.

lherman-cs:
For case 2, should we bring back the `set_current_bitrate`? It would be useful if the stream's bitrate estimate were derived from the same calculation used for the desired bitrate.

lherman-cs:
I don't think the second case is a problem after all. I just found out that the publisher still adjusts the video resolution despite being told to "maintain-resolution", and signaled via vla. I have a logic bug where it basically cuts off this stream completely instead. And, `estimate_recovers_after_capacity_restoration` is a case I was thinking of here. 

The first case is real but minor. In either case, is_overusing is not needed. The real fix is probing at ssrc=0 or simply allocating anyway on the very first allocation.

algesten:
> 1. Bootstrap: BWE says 500Kbps, but we want 1Mbps. So, the desired_bitrate is set to 1Mbps. But BWE never expands the available bandwidth. There's no RTX cache yet.

Before there is any RTX, I think libWebRTC sends dummy data on SSRC 0 for this purpose. We could look into doing that in str0m.

PR #1019, opened 2026-08-05
Add support for DTLS over ICE

pthatcher:


tests/dtls_over_ice.rs
```diff
@@ -0,0 +1,108 @@
+//! DTLS handshake run over ICE, inside STUN connectivity checks.
```
xnorpx, line 1: I would test all combination of this in 

https://github.com/algesten/str0m/blob/main/tests/handshake-direct.rs

as well.

so dtls over ice
true, true
true, false
false, true
false, false

xnorpx:
@pthatcher I pushed 2 extra commits that was needed.

algesten:
@pthatcher is this one ready for review or is it still Draft?

xnorpx:
> @pthatcher is this one ready for review or is it still Draft?

@algesten It seems to work but we need to sync with fippo to ensure the different implementations are aligned. (libwebrtc/pion etc)

PR #1020, opened 2026-08-06
Fix small padding invisible to the pacer

lherman-cs:
Padding smaller than the 800-byte average size divided to zero fake packets, so queue_state_padding() returned None and the pacer never scheduled it. Round up so any nonzero padding stays visible to the pacer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+  * Fix small padding invisible to the pacer #1020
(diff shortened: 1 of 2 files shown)


algesten (approved):

algesten:
@lherman-cs if you add a CHANGELOG entry we can merge.

lherman-cs:
@algesten done

algesten (approved):

algesten:
thanks!

PR #1022, opened 2026-08-08
Respond to RRTR with DLRR

kvasilye:
Send DLRR in response to RRTR

Usefulness: subscribers can measure their RTT

https://datatracker.ietf.org/doc/html/rfc3611#section-4.4

https://datatracker.ietf.org/doc/html/rfc3611#section-4.5

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -737,7 +737,13 @@ impl Session {
                 // DLRR responder (RFC 3611 §4.5): store the remote's RRTR so we can
                 // reply with a DLRR, letting the remote compute RTT.
                 let lrr = (rrtr.ntp_time.as_ntp_64() >> 16) as u32;
-                self.pending_rrtrs.insert(ssrc, LastRrtr { lrr, received_at: now });
+                self.pending_rrtrs.insert(
+                    ssrc,
+                    LastRrtr {
+                        lrr,
+                        received_at: now,
+                    },
+                );
                 continue;
             }
 
--- a/src/session.rs
+++ b/src/session.rs
@@ -744,7 +744,10 @@ impl Session {
                 // DLRR responder (RFC 3611 §4.5): store the remote's RRTR so we can
                 // reply with a DLRR, letting the remote compute RTT.
                 let lrr = (rrtr.ntp_time.as_ntp_64() >> 16) as u32;
-                let entry = LastRrtr { lrr, received_at: now };
+                let entry = LastRrtr {
+                    lrr,
+                    received_at: now,
+                };
 
                 // Update in place if this SSRC is already pending (preserves queue order).
                 if let Some(existing) = self.pending_rrtrs.iter_mut().find(|(s, _)| *s == ssrc) {
--- a/src/session.rs
+++ b/src/session.rs
@@ -288,12 +288,15 @@ impl Session {
             &mut self.feedback_tx,
         );
 
-        if do_nack {
-            self.last_nack = now;
-        }
-
-        // Flush any pending RRTRs as a single DLRR ExtendedReport.
-        if !self.pending_rrtrs.is_empty() {
+        // Flush any pending RRTRs as a DLRR alongside the Sender Report compound packet.
+        // We piggyback on the SR cadence so the DLRR doesn't generate extra standalone
+        // RTCP traffic that could perturb bandwidth estimation.
+        if !self.pending_rrtrs.is_empty()
+            && self
+                .feedback_tx
+                .iter()
+                .any(|r| matches!(r, Rtcp::SenderReport(_)))
+        {
             let items: Vec<DlrrItem> = self
                 .pending_rrtrs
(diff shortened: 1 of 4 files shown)


algesten:
Hey @kvasilye thanks for look at this.

I think there's a problem which was already there before this PR. The `ssrc()` in the RRTR is a remote SSRC, and it's already routed as if it is local. That in turn affects this PR shape. It should be stored on session and we send a response as soon as we can.

kvasilye:
I think you mean that the RRTR's SSRC may not necessarily be one of ours that we're publishing with, correct?

Let me work on fixing this.

kvasilye:
Fixed - there is now a map of SSRC -> most recently received RRTR where the SSRC values don't have to match any of our publish SSRC values.

This map is flushed in the timeout method, but only when we have just queued up a Sender Report. This allows the DLRR to be combined with the sender report, so there is still only one outgoing packet.

Without checking for the pending Sender Report - meaning the DLRR was sent as its own RTCP packet - the bwe test started to under-estimate and fail.

src/session.rs
```diff
@@ -44,6 +45,17 @@ const NACK_MIN_INTERVAL: Duration = Duration::from_millis(33);
...
+/// Stored at session level so we can respond with a DLRR regardless of which local stream
+/// the remote is observing.
+#[derive(Debug, Clone, Copy)]
+struct LastRrtr {
+    /// Middle 32 bits of the RRTR NTP timestamp (compact NTP / LRR).
+    lrr: u32,
+    /// When the RRTR arrived locally.
+    received_at: Instant,
+}
+
```
algesten, line 53: This should be after the `Session` struct.

src/session.rs
```diff
@@ -117,6 +129,10 @@ pub(crate) struct Session {
     feedback_rx: VecDeque<Rtcp>,
     rtp_closing: bool,
 
+    /// Pending RRTRs from remote SSRCs, keyed by the sender's SSRC.
+    /// Flushed as a single DLRR ExtendedReport on the next handle_timeout.
+    pending_rrtrs: HashMap<Ssrc, LastRrtr>,
```
algesten: Looking at libwebrtc, it keeps pending RRTRs in FIFO order, deduplicated by SSRC. A new RRTR for an already pending SSRC updates its value without changing its queue position. It caps the pending queue at 300 entries and consumes at most 50 per DLRR report, leaving the rest for subsequent reports.

We should use a `VecDeque` plus the 300 and 50 limit. That would bound memory and DLRR packet size.

algesten (commented):
A couple of things.

Also, a pre-existing outbound RRTR issue in `StreamRx::create_extended_receiver_report()` sets the XR originator SSRC to self.ssrc, which is the remote media source SSRC. That should be local I think.

kvasilye:
Updated:

  1. Moved LastRrtr struct after the Session struct
  2. Changed pending_rrtrs from HashMap to VecDeque — FIFO order, deduplicates by SSRC in place, capped at 300 entries, consumes at most 50 per DLRR report (matching libwebrtc's approach)
  3. Moved constants to the top alongside the other module-level constants

kvasilye:
> Also, a pre-existing outbound RRTR issue in `StreamRx::create_extended_receiver_report()` sets the XR originator SSRC to self.ssrc, which is the remote media source SSRC. That should be local I think.

In this pull request, `str0m` is the publisher - and the remote SFU is the subscriber.

Since str0m is the publisher, it sends DLRR in response to the SFU's RRTR, and thus the SSRC should be str0m's no?

Sending RRTR (as your comment says) would be the opposite - 1) str0m is the subscriber and wants to measure the RTT by sending RRTR 2) the SFU is the publisher and will reply with DLRR.

I'm not concerned about the "str0m is the subscriber" case in this pull request but could take a look in a separate follow-up pull request if you wanted a complete RTRR / DLRR implementation for both modes.

algesten:
Yeah. That's what I mean. The RRTR sender side is the opposite direction and can be a separate fix.

It does affect the test in this PR though. The test uses str0m to create the RRTR, so it gets the remote SSRC from that existing bug. That means it doesn't actually test that an RRTR with an unrelated remote SSRC is handled.

But yeah, whatevs, we can do another PR for that.

algesten (approved):

kvasilye:
@algesten what's the plan now?

Any plans to merge this given that you already approved?

kvasilye:
Thank you Martin.

I'll submit a new, separate PR to fix the sending SSRC.

kvasilye:
Oh wait looks like you already fixed that:

https://github.com/algesten/str0m/commit/8855f4ee7740a782684cecf0fe7d2947ae1c67a4

When can we expect a new release tag?

algesten:
@kvasilye done. 0.23.0 on crates.io now.

kvasilye:
Appreciate it!

PR #1023, opened 2026-08-08
send media packets unpadded

### LINKED ISSUE #1014: media packets are always padded to the SRTP block size, which breaks SFU forwarding
Every outgoing media packet is padded to the 16-byte SRTP block with the RTP padding bit set (`streams/send.rs`, "pad for SRTP"; present in 0.20 and 0.21). 

This is legal RTP and direct str0m or pion receivers strip it fine, but an SFU that forwards packets near-verbatim can mishandle it: LiveKit strips the padding bytes on ingest yet forwards downstream with the P bit still set, so a Chrome subscriber reads payload tail bytes as a padding length and discards every frame (framesReceived stays 0). 

Filed the LiveKit side as livekit/livekit issue too.

AES-CTR does not need block-aligned input, so the padding appears to exist only to satisfy the aligned-input assert in `rtp/srtp.rs`. Could media padding be dropped or made optional?

Packet captures and stats available on request.

boxerab:
fixes #1014

Every outgoing media packet was padded to the 16-byte srtp block with the P bit set. `aes-ctr` is a stream cipher and doesn't need aligned input, so the padding only existed to satisfy the assert in `protect_rtp`. This drops the `pad_packet` call from the media send path and the assert with it. Pacer padding packets are untouched.

On the history question from the issue: the `assert` and `SRTP_BLOCK_SIZE` came in `1c968a30` (sep 2022), when the cipher was already `openssl aes_128_ctr`. The `pad_packet` call followed a week later in `a73562de` to satisfy the `assert`. None of the backends since (`rust-crypto`, `aws-lc-rs`, the `gcm` profiles) need block-aligned input either, so the padding never had a crypto reason.

I have verified against livekit: the publisher from the issue went from 31034 packets / 0 frames received in Chrome to normal playback with this patch. lib tests pass unchanged.

### DIFF
--- a/src/rtp/header.rs
+++ b/src/rtp/header.rs
@@ -91,6 +91,7 @@ impl RtpHeader {
         buf[from + pad - 1] = pad as u8;
     }
 
+    #[allow(dead_code)] // unused since media packets are sent unpadded
     pub(crate) fn pad_packet(
         buf: &mut [u8],
         header_len: usize,
--- a/src/rtp/header.rs
+++ b/src/rtp/header.rs
@@ -91,23 +91,6 @@ impl RtpHeader {
         buf[from + pad - 1] = pad as u8;
     }
 
-    #[allow(dead_code)] // unused since media packets are sent unpadded
-    pub(crate) fn pad_packet(
-        buf: &mut [u8],
-        header_len: usize,
-        body_len: usize,
-        block_size: usize,
-    ) -> usize {
-        let pad_len = block_size - body_len % block_size;
-        if pad_len == block_size {
-            return 0;
-        }
-
-        Self::do_pad(buf, header_len + body_len, pad_len);
-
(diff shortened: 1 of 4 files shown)


src/rtp/header.rs
```diff
@@ -91,6 +91,7 @@ impl RtpHeader {
         buf[from + pad - 1] = pad as u8;
     }
 
+    #[allow(dead_code)] // unused since media packets are sent unpadded
     pub(crate) fn pad_packet(
```
algesten: Let's delete the code

algesten (approved):
Let's go!

algesten:
Thanks!

boxerab:
Thank you for quick reply, and also for building this wonderful software - exactly what I needed for my project. :pray:

PR #1025, opened 2026-08-10
Add DirectApi::send_pli_feedback

ekaratzas:
Send a Picture Loss Indication (PLI, PT=206 FMT=1) for a media SSRC with a caller-chosen RTCP sender SSRC. Unlike requesting a keyframe on a local receive stream, this needs no receive stream for the media SSRC and lets the caller set the feedback's sender SSRC, so it can relay a keyframe request for a stream sourced by the remote peer.


algesten (approved):

algesten:
LGTM!

PR #1028, opened 2026-08-19
Allow configuration of RTCP sender SSRCs

pthatcher:
And make sure RTCP compound packets share a single RTCP sender SSRC.


algesten:
@pthatcher is this one ready for review or is it still Draft?

PR #1029, opened 2026-08-21
Guard SRTP/SRTCP and RTCP parsing against malformed input

algesten:
Guards and bounds on parsing paths that could be reached with input that doesn't
match what the code assumed:

- `src/rtp/srtp.rs` - check the length covers the RTP header plus the auth tag
  before slicing, and the SRTCP header plus index plus tag for RTCP.
- `src/rtp/rtcp/remb.rs` - bound the SSRC count against the buffer.
- `src/rtp/rtcp/twcc.rs` - a `VectorDouble` chunk can carry `0b11` for any of its
  statuses, which the parser maps to `PacketStatus::Unknown`. `TwccIter::next` had
  no arm for that.
- `src/rtp/rtcp/xr.rs` - cap the `Dlrr` pre-allocation by what the buffer can hold.
- `src/rtp/ext.rs` - `Instant - Duration` panics on underflow, so saturate.
- `src/channel.rs` - stop the SCTP stream id walk from running off the end of `u16`.
- `src/sctp/mod.rs` - `entries` moved to a `BTreeMap`, plus a stream cap, a
  recursion depth cap, `saturating_sub` in `available()`, and a graceful close when
  an `AwaitConfig` stream is already gone.

Tests in `tests/malformed-packet-injection.rs` push hand-crafted datagrams through
`Rtc::handle_input`. Two of them sweep a length range and collect every length that
fails instead of stopping at the first. Two more sit next to the code in
`src/channel.rs` and `src/sctp/mod.rs`.

Found by @gab8i and @dmitry-markin. The bulk of this is theirs, I added tests and
one extra fix on top.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+  * Guard SRTP/SRTCP and RTCP parsing against malformed input #1029
+
 # 0.23.0
 
   * Respond to RRTR with DLRR #1022
--- a/crypto/apple-crypto/src/srtp.rs
+++ b/crypto/apple-crypto/src/srtp.rs
@@ -69,7 +69,11 @@ impl AeadAes128GcmCipher for AppleCryptoAeadAes128GcmCipher {
         input: &[u8],
         output: &mut [u8],
     ) -> Result<usize, CryptoError> {
-        assert!(input.len() >= AeadAes128Gcm::TAG_LEN);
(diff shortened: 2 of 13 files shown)


xnorpx:
Bttreenap i tend to avoid due to heap allocations but I have not looked at the code how frequent there is add and removes

xnorpx (approved):

PR #1030, opened 2026-08-23
Refactor codec- and RED-specific logic out of Payloader

algesten:
## What

Follow-up to #982: pull the codec- and feature-specific logic out of `Payloader` so it is a generic packetization pipe.

`Payloader` had accumulated baggage that isn't its job:

- the whole RFC 2198 RED send state (`RedState` + `wrap`/history), plus a per-payload `set_red`;
- per-codec branches: H265/H266 DONL setup (`if let CodecPacketizer::H265 …`), a `Codec::CN` talkspurt-marker carve-out, and the audio/video `nackable` rule.

After this it is just `{ pack, clock_rate }` and names no `Codec` or `CodecPacketizer` variant.

## RED moves out to a sink

- `RedState` becomes `RedSender` in `packet::red` — the send-side companion to `RedDecoder`, so all RFC 2198 byte-work lives in one module. Its state (history, distances) is owned by `Media`, bundled with the payloader in a single per-`(pt, rid)` map (a `PayloaderEntry` value) rather than a parallel map on the same key.
- `Payloader::push_sample` emits an `OutgoingPacket` through a `PacketSink`. `StreamTx` is the plain sink; `RedSink` decorates the stream, wraps the payload and swaps to the RED PT. `Media` picks which sink to hand in.
- Dispatch is **static** (`push_sample<S: PacketSink>`), so the non-RED path compiles with no RED branch at all — better than the previous per-packet `if self.red.is_some()`.
- No extra allocation: the sink receives the primary as an owned `Vec` and RED wraps it **before** the single `Vec → Arc` conversion, matching the previous allocation profile exactly.

## Codec quirks move onto the packetizer

- `Packetizer` gains `marks_talkspurt()` (default false; audio overrides true, comfort noise stays false) and `nackable()` (default true; audio overrides false). `push_sample` no longer takes an `is_audio` argument.
- `CodecPacketizer::configure_for(&spec)` owns the DONL setup, so the H265/H266 knowledge stays inside the codec-aware type.

## Notes

- Behaviour is unchanged. The RED wrap unit tests move to `packet::red` alongside `RedSender`; the full suite (incl. `tests/red.rs`) is green.
- Not addressed here (pre-existing, tracked separately): the receive-side RTX-of-RED composition — `is_repair`/`is_red` in `session.rs` is still mutually exclusive, so an RTX carrying a RED payload from a non-conformant peer isn't unwound. Only reachable with a non-conformant sender; noted as a known limitation.

### DIFF
--- a/src/media/mod.rs
+++ b/src/media/mod.rs
@@ -8,7 +8,8 @@ use crate::RtcError;
 use crate::change::AddMedia;
 use crate::format::CodecConfig;
 
-use crate::packet::{CodecDepacketizer, DepacketizingBuffer, Payloader, RtpMeta};
+use crate::packet::{CodecDepacketizer, DepacketizingBuffer, Payloader};
+use crate::packet::{RedSender, RedSink, RtpMeta};
 use crate::rtp_::ExtensionMap;
 use crate::rtp_::MidRid;
 use crate::rtp_::SRTP_BLOCK_SIZE;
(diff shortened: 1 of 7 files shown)


farit2000:
Looks good. Went through the RED side and the behaviour's preserved. Nice cleanup.

xnorpx (approved):

PR #1032, opened 2026-08-28
fix(rtp): keep the voice activity bit out of the audio level's sign

AlexanderSerbul:
## What

When serializing the RFC 6464 `ssrc-audio-level` extension, the voice activity
(V) bit is forced to `1` for every non-zero level. A header written with
`voice_activity: Some(false)` is parsed back as `Some(true)`.

The audio level itself is always written correctly — only the flag is lost.

## Why

```rust
buf[0] = if v2 { 0x80 } else { 0 } | (-(0x7f & v1) as u8);
```

`audio_level` is negative, so masking it before negating leaves the sign in
place. For `v1 = -37`:

| step | value |
|---|---|
| `v1` as bits | `0xDB` |
| `0x7f & v1` | `0x5B` = 128 − 37 |
| `-(0x5B)` as `u8` | `0xA5` = 128 + 37 |

Bit 7 is already set before the flag term is OR-ed in, and OR cannot clear it.
The low seven bits come out as the magnitude, which is why the level survived
and only the flag was lost — and why this went unnoticed: with
`voice_activity: Some(true)` everything matches.

## Fix

```rust
buf[0] = if v2 { 0x80 } else { 0 } | (v1.unsigned_abs() & 0x7f);
```

Checked exhaustively over the scale: of the 256 `(level, voice_activity)` pairs,
127 used to round-trip with the wrong flag — all with the flag `false` and a
non-zero level — and all 256 round-trip correctly now. The added test
`audio_level_round_trip` walks the whole scale and fails on `main`
(`left: Some(true), right: Some(false)`).

## Test changes

The two golden-byte tests in `rtp::header` encoded the old behaviour: they build
headers with `voice_activity: Some(false)` and expect `170` (`0xAA`) where the
correct byte is `42` (`0x2A`). Updated with a note explaining why, so the change
does not read as a loosened expectation.

The parsing tests are deliberately untouched: `0xAA` on the wire really does
mean "voice, level 42", and the parse side needs no change.

`cargo test --lib` is green (707 passed), `cargo fmt --check` clean.

## How it was found

Building an SFU on str0m. Our own test asserted that a declared "not speech"
flag survives forwarding, and it came back as speech every time. For an SFU this
matters: the point of RFC 6464 is to make forwarding decisions without
decrypting or decoding, and speaker detection that trusts the V bit cannot tell
speech from steady background noise.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
   * Add RFC 2198 RED (redundant audio) support for all audio codecs #982
+  * Preserve the RFC 6464 audio-level voice activity bit when serializing #1032
 
 # 0.23.1
 
--- a/src/rtp/ext.rs
+++ b/src/rtp/ext.rs
@@ -664,7 +664,11 @@ impl Extension {
             AudioLevel => {
                 let v1 = ev.audio_level?;
                 let v2 = ev.voice_activity?;
-                buf[0] = if v2 { 0x80 } else { 0 } | (-(0x7f & v1) as u8);
+                // `v1` is negative, so masking it before negating leaves the
+                // sign bit in place: for -37 the expression `-(0x7f & v1)`
+                // yields 0xA5, which already has bit 7 set. Taking the
+                // magnitude first keeps the sign out of the V bit.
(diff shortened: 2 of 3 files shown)


algesten (commented):
Excellent find! Thanks!

algesten (approved):

PR #1034, opened 2026-09-05
Fix Clippy

xnorpx:
## Summary

Clean up the clippy warnings reported by `cargo clippy --all-targets`. All changes are in test code (unit tests under `src/` and integration tests under `tests/`); no production behavior changes.

- **`src/packet/vp9.rs`**: allow `clippy::erasing_op` and `clippy::identity_op` on `test_vp9_packet_unmarshal` so the fixture keeps its explicit bit-layout expressions (`N:0` etc.). Replace the hand-rolled `r0` counter loops with a range, and drop a redundant `& 0xFF` on a byte comparison.
- **`src/config.rs`**: construct the deliberately inverted MTU range with `RangeInclusive::new(1300, 1299)` to satisfy `reversed_empty_ranges` while keeping the expected-panic test.
- **`src/util/value_history.rs`**: the `unchecked_duration_subtraction` lint was renamed upstream; update the `allow` to `unchecked_time_subtraction`.
- **`tests/handshake-direct.rs`**: group the event-loop I/O handles into an `RtcLoopIo` struct and pass `IceCreds` directly to cut the argument count; use `if let` in `write_pcap`; collapse the ICE `Checking` arm into a match guard.
- **Elsewhere**: `repeat(x).take(n)` -> `repeat_n(x, n)`, `.last()` -> `.next_back()`, `&[x.clone()]` -> `slice::from_ref(&x)`, arrays instead of `vec!` for fixtures, and assorted needless borrows, casts, clones, `unwrap_or_else` -> `unwrap_or`, `assert_eq!(.., false)` -> `assert!(!..)`, a `Default` field reassign, a boxed `Fn` type alias, and a nested `if` folded into a match guard.

## Validation

- `cargo clippy --all-targets -- -D warnings` is clean on stable.
- `cargo test` passes.
- MSRV: `cargo msrv verify` and `cargo msrv verify -- cargo check --all-targets` pass on 1.85.0, and `cargo +1.85.0 test --tests` passes.
- `cargo fmt --all -- --check`

### DIFF
--- a/src/packet/vp9.rs
+++ b/src/packet/vp9.rs
@@ -1086,13 +1086,13 @@ mod test {
             (
                 "ParseRefIdx",
                 &[
-                    0xD8,                              /* I:1 P:1 L:0 F:1
-                                                        * B:1 E:0 V:0 Z:0 */
-                    (0x80 | ((17 >> 8) & 0x7F)) as u8, // Two byte pictureID.
-                    17,                                // TL0PICIDX
-                    (17 << 1) | 1,                     // P_DIFF N:1
-                    (18 << 1) | 1,                     // P_DIFF N:1
-                    (127 << 1) | 0,                    // P_DIFF N:0
+                    0xD8,          /* I:1 P:1 L:0 F:1
+                                    * B:1 E:0 V:0 Z:0 */
+                    0x80,          // Two byte pictureID.
+                    17,            // TL0PICIDX
+                    (17 << 1) | 1, // P_DIFF N:1
+                    (18 << 1) | 1, // P_DIFF N:1
+                    127 << 1,      // P_DIFF N:0
                 ],
                 Vp9Depacketizer {
                     i: true,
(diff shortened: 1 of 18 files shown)


src/packet/vp9.rs
```diff
@@ -1086,13 +1086,13 @@ mod test {
             (
                 "ParseRefIdx",
                 &[
-                    0xD8,                              /* I:1 P:1 L:0 F:1
-                                                        * B:1 E:0 V:0 Z:0 */
-                    (0x80 | ((17 >> 8) & 0x7F)) as u8, // Two byte pictureID.
```
xnorpx, line 1091: I guess we can remove  the as u8 but leave the rest?

src/packet/vp9.rs
```diff
@@ -1086,13 +1086,13 @@ mod test {
...
-                    17,                                // TL0PICIDX
-                    (17 << 1) | 1,                     // P_DIFF N:1
-                    (18 << 1) | 1,                     // P_DIFF N:1
-                    (127 << 1) | 0,                    // P_DIFF N:0
```
xnorpx, line 1095: leave N:0 here

xnorpx: add allow if needed

xnorpx:
@algesten this should be ready

algesten (approved):

PR #1035, opened 2026-09-07
fix: is, candidate invalidation with protocol

Minaakaa:
this can be an issue if we have both tdp and udp, we need to differentiete the protocols

note: i am not good at english i will try to write a bigger comment if it is ok


algesten:
This looks like a correct change. Seems to be a mistake it was omitted.

Thanks!

algesten (approved):

PR #1038, opened 2026-09-09
Remove SChannel; use dimpl + CNG for Windows DTLS

algesten:
The `wincrypto` backend now always uses dimpl for DTLS 1.2/1.3 and Windows CNG for cryptographic operations. Remove the native SChannel implementation, its unused legacy provider, and SChannel-only dependencies. Keep `wincrypto-dimpl` and the backend crate's `dimpl` feature as compatibility aliases.

Update the backend documentation and enable the existing DTLS 1.3/Auto, shutdown, and retransmission tests for `wincrypto`. Both Windows feature configurations remain covered by CI.

Validation:
- All 19 focused handshake, shutdown, and retransmission tests passed with rust-crypto on macOS.
- Windows cross-checks passed for str0m-wincrypto including tests, and str0m with both wincrypto and wincrypto-dimpl.
- Windows-target Clippy with all backend features and warnings denied passed; formatting and diff checks passed.
- Native Windows runtime tests require the Windows CI runners.


tests/common.rs
```diff
@@ -466,12 +466,7 @@ fn get_crypto_provider_by_name(name: &str) -> CryptoProvider {
         #[cfg(all(feature = "wincrypto-dimpl", target_os = "windows"))]
```
efer-ms: why is this still here? shouldn't wincrypto-dimpl .. simply turn on "wincrypto" in the cargo.toml. Then the code doesn't need to use wincrypto-dimpl anywhere?

tests/common.rs
```diff
@@ -488,12 +483,7 @@ fn get_crypto_provider_by_name(name: &str) -> CryptoProvider {
...
             available.push("openssl-dimpl");
-            // If both wincrypto and wincrypto-dimpl are declared, only wincrypto-dimpl is available.
-            #[cfg(all(
-                feature = "wincrypto",
-                not(feature = "wincrypto-dimpl"),
-                target_os = "windows"
-            ))]
+            #[cfg(all(feature = "wincrypto", target_os = "windows"))]
             available.push("wincrypto");
             #[cfg(all(feature = "wincrypto-dimpl", target_os = "windows"))]
```
efer-ms: ditto here. like I think it's fine to maintain the alias (although I"m not sure it's worth it, how many people are using wincrypto, and migrating from wincrypto-dimpl -> wincrypto is just a small cargo file update)

tests/dtls-close.rs
```diff
@@ -4,7 +4,7 @@
     feature = "rust-crypto",
     feature = "openssl",
     feature = "openssl-dimpl",
-    feature = "wincrypto-dimpl",
+    feature = "wincrypto",
```
efer-ms, line 7: I wonder if we should take to opportunity to rename to match the other styles... so 'win-crypto' to go with 'rust-crypto' and 'apple-crypto'

efer-ms: again, we could alias it but I'd wager there are not many people impacted by this getting renamed

algesten:
I think I'll keep `wincrypto`. The `wincrypt.h` naming is part of why it reads naturally to me. Matching `rust-crypto` and `apple-crypto` isn't enough reason to rename the existing feature and crate.

The redundant `wincrypto-dimpl` checks are gone now. It's just a Cargo alias for `wincrypto`.

xnorpx (approved):

PR #1039, opened 2026-09-10
Let Rtc::accepts see an ice-lite agent's validated remote

mishan:
# Let `Rtc::accepts` see an ice-lite agent's validated remote

## The problem

`IceAgent::has_viable_remote_candidate` (`crates/is/src/agent.rs:514`) only considers
candidate pairs in `CheckState::Succeeded`, and a pair reaches that state exclusively in
`CandidatePair::record_binding_response` (`crates/is/src/pair.rs:352`) — on a response to a
check *we* sent.

An ice-lite agent sends no checks: `handle_timeout` returns before
`stun_client_binding_request` with `"Stop timeout since ice-lite do no checks"`
(`agent.rs:1251`), and `poll_timeout` returns `None` for the pair schedule (`agent.rs:1316`).
So a lite agent's pairs stay in `CheckState::Waiting` for the life of the connection, and
`has_viable_remote_candidate` always answers `false` no matter how many binding requests the
agent has answered from that address.

That leaves `Rtc::accepts` (`src/lib.rs:1880`) with two working paths out of three for a lite
`Rtc`: the `send_addr` fast path and STUN by ICE credentials. The slow path is dead.

Two consequences, one transient and one permanent:

- **Transient.** A user demultiplexing many `Rtc` over one socket the documented way — offer
  each datagram to each `Rtc`, hand it to the first that accepts, drop the rest — refuses the
  peer's first DTLS record. It arrives after we answered the peer's binding request but before
  the nominating request has been polled out into `send_addr`, so no path matches. The
  handshake then waits out a DTLS retransmit, about a second. `examples/chat.rs` is one line
  from this: it demultiplexes with `accepts()` at line 154, drops what nobody accepts at line
  160, and has `.set_ice_lite(true)` commented out in its builder at line 88.
- **Permanent.** Traffic from a validated remote address of the same peer that isn't the
  currently nominated one — the exact case the slow path's own comment describes — is refused
  for as long as the connection lives.

`accepts()`'s rustdoc describes that path as "all remote ICE candidates are checked for a
match", while the code checks only `Succeeded` *pairs*, which is part of why this reads like an
oversight rather than policy.

The test suite doesn't catch it because it never calls `accepts()` — `tests/common.rs::netem_to_rtc`
hands every datagram straight to `handle_input`.

Reproduced on 0.23.1 and on `main` (b616675f); `accepts()` and `has_viable_remote_candidate` are
identical on both.

## The change

Record the candidate pair as succeeded when a **controlled** ice-lite agent answers a binding
request that passed message integrity, in `IceAgent::handle_packet`.

To be explicit about what that is and isn't: a binding *response* proves return routability, and
an answered *request* does not — it proves only that the sender knows our password, and an
off-path spoofer who has the SDP can forge one from any source address. The argument isn't that
these are equivalent. It's that:

- an ice-lite agent cannot obtain return routability at all, by construction, so requiring
  `Succeeded` to imply it means requiring something the mode is defined not to provide — and what
  that buys is a permanently dead path, not a safer one; and
- for a **controlled** lite agent this grants an attacker strictly less than they already have. A
  spoofer who knows the password can today send a forged USE-CANDIDATE from any address and reach
  `pair.nominate()` (`agent.rs:1618`) → `NominatedSend` → `Rtc::send_addr`, which redirects
  **egress**. This adds only **ingress** acceptance of datagrams that must still pass DTLS/SRTP.

A **controlling** ice-lite agent is deliberately excluded. There `CheckState::Succeeded` feeds
`evaluate_nomination()` (`agent.rs:1622`), whose controlling branch picks `max_by_key(prio)` over
succeeded pairs — and a peer-reflexive pair's priority comes from the sender's own PRIORITY
attribute, so promoting there would let one forged request choose the send address. That
configuration is reachable only through the direct API (`set_ice_lite(true)` +
`set_ice_controlling(true)`; the SDP path forces `controlling = !ice_lite` at
`src/change/sdp.rs:698`) and is left exactly as it is today.

## Tests

Three, each verified failing before the change and passing after:

- `is::agent::test::ice_lite_remote_candidate_viable_after_binding_request` —
  `has_viable_remote_candidate` is true after an answered request.
- `is::agent::test::ice_lite_controlling_does_not_nominate_on_a_binding_request` — the guard
  above.
- `tests/ice-lite-accepts.rs` — an ice-lite `Rtc` accepts a DTLS record from a peer whose binding
  request it just answered.

`cargo test -p str0m -p is` is green: 1118 tests, no failures.

## An alternative shape

If you'd rather not overload `CheckState`, the same fix works as a dedicated
`remote_validated: bool` on `CandidatePair`, set on the same event and read by
`has_viable_remote_candidate`. That sits closer to how the existing ice-lite work treats lite
pairs (`remote_binding_request_time` rather than `CheckState`) and needs no controlling guard at
all, since nothing else would read it. Happy to send that instead — say the word.

One note if you go that way: the existing `remote_binding_request_time` /
`remote_binding_requests` fields can't be reused as-is. `agent.rs:1251` stamps them on every pair
on each lite `handle_timeout`, so both would report "validated" for pairs that never received
anything.

<details>
<summary>Standalone reproduction, without this repo's test harness</summary>

str0m 0.23.1, `default-features = false, features = ["rust-crypto"]`, edition 2024.

```toml
[dependencies]
str0m = { version = "0.23.1", default-features = false, features = ["rust-crypto"] }
```

```rust
use std::net::SocketAddr;

use str0m::net::{Protocol, Receive};
use str0m::{Candidate, Input, Output, RtcConfig};

/// Set to true to make B an ice-lite agent.
const ICE_LITE: bool = true;

fn main() {
    str0m::crypto::from_feature_flags().install_process_default();

    let start = Instant::now();
    let mut now = start;

    let a_addr: SocketAddr = "1.1.1.1:1000".parse().unwrap();
    let b_addr: SocketAddr = "2.2.2.2:2000".parse().unwrap();

    let mut a = RtcConfig::new().build(now);
    let mut b = RtcConfig::new().set_ice_lite(ICE_LITE).build(now);

    a.add_local_candidate(Candidate::host(a_addr, "udp").unwrap());
    b.add_local_candidate(Candidate::host(b_addr, "udp").unwrap());

    let fa = a.direct_api().local_dtls_fingerprint().clone();
    let fb = b.direct_api().local_dtls_fingerprint().clone();
    a.direct_api().set_remote_fingerprint(fb);
    b.direct_api().set_remote_fingerprint(fa);

    let ca = a.direct_api().local_ice_credentials();
    let cb = b.direct_api().local_ice_credentials();
    a.direct_api().set_remote_ice_credentials(cb);
    b.direct_api().set_remote_ice_credentials(ca);

    a.direct_api().set_ice_controlling(true);
    b.direct_api().set_ice_controlling(false);

    a.add_remote_candidate(Candidate::host(b_addr, "udp").unwrap());
    b.add_remote_candidate(Candidate::host(a_addr, "udp").unwrap());

    a.direct_api().start_dtls(true).unwrap();
    b.direct_api().start_dtls(false).unwrap();

    for _ in 0..400 {
        for from_a in [true, false] {
            let (src, dst, src_addr, dst_addr) = if from_a {
                (&mut a, &mut b, a_addr, b_addr)
            } else {
                (&mut b, &mut a, b_addr, a_addr)
            };

            src.handle_input(Input::Timeout(now)).unwrap();

            let mut out = Vec::new();
            loop {
                match src.poll_output().unwrap() {
                    Output::Timeout(_) => break,
                    Output::Transmit(t) => out.push(t.contents.to_vec()),
                    Output::Event(e) => {
                        println!("[{:?}] {} {:?}", now - start, side(from_a), e)
                    }
                }
            }

            for data in out {
                let input = Input::Receive(
                    now,
                    Receive {
                        proto: Protocol::Udp,
                        source: src_addr,
                        destination: dst_addr,
                        contents: data.as_slice().try_into().unwrap(),
                    },
                );

                // The documented "one socket, many Rtc" pattern: offer the
                // datagram to each Rtc, give it to the first that accepts,
                // drop what nobody accepts.
                if !dst.accepts(&input) {
                    println!(
                        "[{:?}] {} REFUSED a datagram starting with byte {:?}",
                        now - start,
                        side(!from_a),
                        data.first()
                    );
                    continue;
                }

                dst.handle_input(input).unwrap();
            }
        }

        now += Duration::from_millis(5);
    }

    println!(
        "ice_lite={ICE_LITE} a_connected={} b_connected={}",
        a.is_connected(),
        b.is_connected()
    );
}

fn side(is_a: bool) -> &'static str {
    if is_a {
        "A"
    } else if ICE_LITE {
        "B (ice-lite)"
    } else {
        "B"
    }
}
```

`ICE_LITE = true`:

```
[0ns] A IceConnectionStateChange(Checking)
[0ns] B (ice-lite) IceConnectionStateChange(Checking)
[5ms] A IceConnectionStateChange(Completed)
[5ms] B (ice-lite) REFUSED a datagram starting with byte Some(22)
[5ms] B (ice-lite) IceConnectionStateChange(Completed)
[830ms] B (ice-lite) Connected
[835ms] A Connected
ice_lite=true a_connected=true b_connected=true
```

Byte 22 is a DTLS handshake record. The refusal is deterministic at 5 ms; the connect time
varies with the DTLS retransmit timer, 780 ms – 1.02 s across runs.

`ICE_LITE = false`, everything else unchanged:

```
[0ns] A IceConnectionStateChange(Checking)
[0ns] B IceConnectionStateChange(Checking)
[5ms] A IceConnectionStateChange(Completed)
[5ms] B IceConnectionStateChange(Completed)
[10ms] B Connected
[15ms] A Connected
ice_lite=false a_connected=true b_connected=true
```

Nothing is refused and the connection is up in one round trip.

</details>


algesten (approved):
Looks good!

Issue #1037, opened 2026-09-08
Configure Dimpl DTLS retransmission timeouts through str0m-wincrypto


Ramil980:
Hi @algesten 

We use Dimpl as our DTLS implementation and str0m-wincrypto as its Windows CNG backend. We need to configure the initial RTO, retry count, and overall handshake timeout.

Currently, we can configure these settings by constructing Dimpl directly, but  str0m-wincrypto  does not publicly expose its Dimpl CNG provider. Alternatively, we can create Dimpl DTLS through DtlsProvider::new_dtls() , but that API does not expose Dimpl’s retransmission settings

Prototyped 2 quick solution with the help of agents

1. Expose the Dimpl CNG crypto provider: https://github.com/Ramil980/str0m/commit/ef00dd2a307a01cf066532bd09d399f6dce6bced
2. Add configurable options to DtlsProvider: https://github.com/algesten/str0m/compare/main...Ramil980:str0m:configurable-dtls-options

What do you think? Would you prefer either of these approaches, or would you suggest something else?

PR #1040, opened 2026-09-11
Expose Dimpl CNG crypto provider

Ramil980:
Exposes the existing Windows CNG crypto provider so standalone Dimpl consumers can construct their own dimpl::Config , including custom retransmission and handshake timeout settings.

This does not change existing provider behavior or defaults

#1037


algesten (approved):
Let's roll with it.

Issue #1033, opened 2026-09-05
Video reorder queue waits forever for unrecovered frames


efer-ms:
reordering_size_video controls the size of the frame reordering queue. If a packet gets lost and is never recovered, the video queue will grow to reordering_size_video before skipping the frame that went unrecovered. This is a larger problem if video streams have different frame rates. Where a low FPS (2 FPS) feed would hold up to 15s of video if the reorder queue was set to 30, because you are supporting 30fps video elsewhere where the depth is only ~1s.

Proposal is to have an option that can declare a frame is unrecoverable after a timeout. The timeout should be based around measured RTT times, and jitter, as well as if RTX is available or not.

For example,  if RTX is not enabled, then perhaps you just wait for a jitter buffer delay before declaring a frame lost. if RTX is enabled you wait based upon the total # of RTX rounds sent and the interval, plus RTT, and jitter.

This timeout would be set as soon as the first packet is received for a following frame, or the last packet (marker bit set) is received for the frame.

Something like:
  frame_timeout = ((has_rtx ? (current_rtt + rtx_retry) : 0) + current_jitter) * fudge_factor;

Each frame would have it's own timeout, so if frame 0 and frame 1 have loss. Frame 0 is declared lost independently of frame 1.

algesten:
Nice finding! We could do with checking more closely how libwebrtc implements jitter buffers.

efer-ms:
So my investigation libwebrtc is that it works differently. libwebrtc emits completed frames as they complete. So if there are lost packets, the subsequent frames aren't held up. Reordering of the frames occurs after depacketization. Part of the information emitted after depacketization is the frame dependency information extracted from things like DD, or inside the bitstreams.

So say you have frame A (t=0), B (t=1), C (t=2), and D (t=3)... say you some packets in B arrive after D complete, and C has packets that are never recovered. For this example B and C depend on A and D depends on B.

So things get emitted as A(t=0), D(t=3), B(t=1) ... the render queue received the frames in this order, and so if t=3 before B finishes arriving, D will fail to render, and you'll likely see a KF request for out. If B arrives at say t=2 ... then the render is able to render D on time, since it received B.

It doesn't have something quite as similar to what str0m currently has. We could go down the path of having str0m emit the frames as soon as they are complete, but without the additional dependency info it'd be up to the layer above to determine if there is a missing required frame it should wait for, like we can signal that there was a gap, but the gap could be recovered, so I'm not sure that's ideal.

[...]

PR #1041, opened 2026-09-17
Add configurable video reordering timeout

attila-szenczi:
Allow blocked video frames to advance after a configurable fixed duration. Configure it through `RtcConfig::set_reordering_timeout_video(Option<Duration>)` before building the connection. `None` preserves count-based waiting by default; `Some(Duration::ZERO)` skips the wait once a complete frame is available.

Deadlines use each held frame's earliest packet receipt time and feed into the existing timeout scheduling, including when packets stop arriving. Frame-count limits and codec dependency checks still apply. Audio and RTP-mode delivery are unchanged. The setter is infallible; unrepresentable deadlines are handled safely without imposing an arbitrary duration cap.

Refs #1033.

Most of the diff is regression coverage. Suggested review order for the implementation:

- `buffer_rx.rs`: waiting, timeout release, codec-rejection loop, and deadline calculation.
- `media/mod.rs`: video-only application and deadline aggregation across receive buffers.
- `session.rs`: configuration and integration into overall timeout scheduling.

Validation: full `cargo test`, `cargo fmt --all -- --check`, `cargo clippy -- -D warnings`, and `RUSTDOCFLAGS=-Dwarnings cargo doc --no-deps`.


src/packet/buffer_rx.rs
```diff
@@ -222,75 +222,104 @@ impl DepacketizingBuffer {
...
     }
 
-    pub fn pop(&mut self) -> Option<Result<Depacketized, PacketError>> {
-        self.update_segments();
+    pub fn pop(
+        &mut self,
+        now: Instant,
+        reordering_timeout: Option<Duration>,
+    ) -> Option<Result<Depacketized, PacketError>> {
+        loop {
```
attila-szenczi, line 230: Added a loop to avoid getting stuck waiting for polls when a frame is ready to be returned.

Note that the majority of the diffs in this function are due to indentation, review it by ignoring whitespaces.

https://github.com/algesten/str0m/pull/1041/changes/ce5b28c072d2ee64215057814d93e5d0edc57ade?w=1

src/packet/buffer_rx.rs
```diff
@@ -222,75 +222,104 @@ impl DepacketizingBuffer {
...
+                .expect("entry for stop index")
+                .meta
+                .seq_no;
+
+            // We're not going to emit frames in the incorrect order, there's no point in keeping
+            // stuff before the emitted range.
+            self.queue.drain(0..=stop);
+
+            if !can_emit {
+                reordering_timeout?;
```
attila-szenczi: Return when None to preseve the exact original behavior when the field is unset.

src/media/mod.rs
```diff
@@ -337,9 +337,17 @@ impl Media {
     pub(crate) fn poll_sample(
         &mut self,
         params: &[PayloadParams],
+        now: Instant,
+        reordering_timeout: Option<Duration>,
     ) -> Result<Option<MediaData>, RtcError> {
+        let reordering_timeout = if self.kind.is_video() {
+            reordering_timeout
```
efer-ms, line 344: why does this check is_video? won't we pass None for reorderoing_timeout for Audio today anyways? Like it's weird to have an argument that's just ignored

src/media/mod.rs
```diff
@@ -508,6 +516,20 @@ impl Media {
         }
     }
 
+    pub(crate) fn poll_receive_timeout(
+        &mut self,
+        reordering_timeout: Option<Duration>,
+    ) -> Option<Instant> {
+        if !self.kind.is_video() {
```
efer-ms, line 523: same quesiton regarding why this check is_video

src/session.rs
```diff
@@ -1161,6 +1178,13 @@ impl Session {
         let twcc_at = self.twcc_at();
         let pacing_at = self.pacer.poll_timeout();
         let packetize_at = self.medias.iter().flat_map(|m| m.poll_timeout()).next();
+        let video_timeout = self.reordering_timeout_video.filter(|_| !self.rtp_mode);
```
efer-ms, line 1168: more curious than anything else, does this matter? in rtp_mode the depacketizers are not used right? and this value only applies if we're buffering RTP in some manner.

src/packet/buffer_rx.rs
```diff
@@ -222,75 +222,104 @@ impl DepacketizingBuffer {
...
-            last.meta.seq_no
-        };
+            // depack ahead, even if we may not emit right away
+            let mut dep = match self.depacketize(start, stop, seq) {
+                Ok(d) => d,
+                Err(e) => {
+                    // this segment cannot be decoded correctly
+                    // remove from the queue and return the error
+                    self.last_emitted = Some((seq, CodecExtra::None));
+                    self.queue.drain(0..=stop);
```
efer-ms: rather than a loop, can't this just return an error when a frame fails due to timeout? like it's a similar issue.. we couldn't decode correctly.

efer-ms: actually .. why don't we just check the way we do in wait_for_contiguity. Like shouldn't we just have an additional check there? like segmetns.len() >= hold_back || expired

src/packet/buffer_rx.rs
efer-ms: So I'm a bit lost here. Is there an existing behavior (I'm gonna call it a bug), where if the hold_back limit is hit, we don't return the next frame, until someone polls again?

src/packet/buffer_rx.rs
```diff
@@ -222,75 +222,104 @@ impl DepacketizingBuffer {
...
-                // remove from the queue and return the error
-                self.last_emitted = Some((seq, CodecExtra::None));
-                self.queue.drain(0..=stop);
-                return Some(Err(e));
+            // If we have contiguity of seq numbers we emit right away,
+            // Otherwise, we wait for retransmissions up to `hold_back` frames
+            // and re-evaluate contiguity based on codec specific information
+
+            let more_than_hold_back = self.segments.len() >= self.hold_back;
+            let contiguous_seq = self.is_following_last(start);
```
efer-ms: I think all you need here is a check that says.. I have a frame ready.. I've waiting timeout_time for it. And since I did.. go ahead and don't wait for contiguity.

That will drop all the packets in front of it. So you probably need the timestamp for the first packet received in the segment, and if `first_segment_packet_time + timeout <= now` set wait_for_continuity to false.

src/packet/buffer_rx.rs
```diff
@@ -222,75 +222,104 @@ impl DepacketizingBuffer {
         }
     }
```
efer-ms, line 224: so I think when you push a packet in, you should record when the packet was received. update_segments() can calculate the min time a packet was received in the "first completed segment" .. and we use that to decide if the frame should be emitted or not.

I could come up with some degenerate cases where the timeout wouldn't be exact but I don't feel like they'd be worth the extra effort.

efer-ms: segments would then hold (start, end, time_start_received) . 

I suppose a quesiton for @algesten , why do we calculate all the segments? 

we only ever use the "first" one from what I saw. It seems like all we really need is something the returns the first completed segment, the min_time for that segment, and perhaps how many segments we have for the hold_back logic. That could also just be merged into a single call that returns the "first" segment. And have that not emit the first segment.

So like rather than segments.... fn get_next_segment(now: Instant, timeout: Option<Duration>) -> Option<(start, end)> ...

I'm also wondering about "queue"... as far as I can tell it's unbounded? update_segments only creates segmetns for continguous ranges. So it finds a "start" and it finds an "end".. and no gaps. So if I'm a bad actor, can't I send new packets skipping seq numbers, and grow queue unbounded? I could be missing something though. (AI seemed to agree with my reading, but mayber we're both missing something).

algesten: We use `segments.len()` for `hold_back`, so we need the count. But you're right that we only use the first segment's boundaries. We could keep the first segment and a count. I'd leave that refactor for another PR.

Good catch on the unbounded queue. I reproduced it with incomplete VP8 frames and added a fix in #1044. It caps each depacketizer at 4,096 packets, dropping the oldest when a newer packet needs room. The timeout in this PR doesn't cover that case because it needs a complete frame to proceed.

algesten:
I've pushed a simplification. Removed the runtime getter/setter, so the timeout is configured through `RtcConfig` only. Also removed the 600-second cap and its error plumbing. The builder setter now returns `Self`, like the other settings.

sawy3r:
Per conversation with @algesten on #1042 - the same wait exists on audio. `reordering_size_audio` is `15`, which is about 300ms of continuous 20ms packets, but with DTX (a packet every ~400ms) that becomes several seconds, and if the sender goes quiet right after a loss the frames behind the gap are never released at all.

The buffer side of this PR looks kind-agnostic already. The only things keeping audio out are the two `is_video()` gates in `media/mod.rs` (the ones @efer-ms asked about). My suggestion would be a `set_reordering_timeout_audio(Option<Duration>)` next to the video one, mirroring the `reordering_size_audio`/`reordering_size_video` pair, default `None` so nothing changes unless you opt in. Media then picks the value by kind instead of discarding it. I'd keep them separate rather than one shared knob, since a sensible audio value (tens of ms) and a sensible video value (hundreds) are far apart.

Happy to do that as a small follow-up PR once this lands so it doesn't hold up the review here, or as a patch against this branch if you'd rather it went in together.

attila-szenczi:
> I've pushed a simplification. Removed the runtime getter/setter, so the timeout is configured through `RtcConfig` only. Also removed the 600-second cap and its error plumbing. The builder setter now returns `Self`, like the other settings.

@algesten thanks for the review and commit!
In @efer-ms's original proposal the timeout would be dynamic:
`frame_timeout = ((has_rtx ? (current_rtt + rtx_retry) : 0) + current_jitter) * fudge_factor;`

I added the runtime setter so we can do that on the application layer without forcing a specific logic to everyone. @efer-ms , is it okay for you if it's only a startup setting?

src/media/mod.rs
attila-szenczi: Right, originally I passed the video timeout in both cases, changed it in session.rs and removed this check

src/media/mod.rs
attila-szenczi: Removed

src/session.rs
attila-szenczi: Right, it was an overly defensive filter, removed.

src/packet/buffer_rx.rs
attila-szenczi: Basically that's what it does:

```
let wait_for_contiguity = !contiguous_seq
            && !more_than_hold_back
            && !self.timeout_allows_progress(now, first_received, reordering_timeout); // **= expired**
```

We can return an error and wait for scheduler for another poll, to me it sounds more logical to return a frame if we can. @algesten, what do you think? I see you kept the loop in your refactor, should we keep the loop or return an error?

src/packet/buffer_rx.rs
attila-szenczi: I think that's what the code does:
`        let wait_for_contiguity = !contiguous_seq
            && !more_than_hold_back
            && !self.timeout_allows_progress(now, first_received, reordering_timeout);`
            
!timeout_allows_progress checks if the timeout already passed and if so, we don't wait for contiguity but drop the packets and return the frame in the next iteration of the loop.

src/packet/buffer_rx.rs
attila-szenczi: Added first_received to the segments

attila-szenczi:
@sawy3r I agree, the same thing should work for audio, I'd prefer if you would do it as a follow up PR because this is already big enough.

I am open to add audio case to this PR as well.

algesten (approved):

PR #1042, opened 2026-09-17
Bound how long the depacketizing buffer waits for a lost packet

### LINKED ISSUE #1033: Video reorder queue waits forever for unrecovered frames

sawy3r:
Fixes #1033.

### The problem

`DepacketizingBuffer` holds every frame behind a gap in the sequence numbers until `reordering_size_*` frames have piled up behind it. That wait has no upper bound in time. If a packet is lost and never recovered, and fewer than `hold_back` frames arrive after it, the frames behind the gap are never emitted at all — the receiver just goes silent. On a low frame rate video stream it is seconds; on an audio stream with NACK off it is however long until enough frames happen to pile up, if ever.

### The design

You said you were reluctant to add buffers and reordering machinery, but happy with "give up waiting if condition X". That is all this is:

- **No new buffer, no reorderer.** The only new state on `DepacketizingBuffer` is a `max_wait: Duration`, the latest `Instant` it knows about, and the deadline of the frame it is currently holding back.
- **Giving up reuses the hold-back overflow path exactly.** When the deadline passes, `pop()` stops returning early at `wait_for_contiguity` and falls through to the same code the `more_than_hold_back` case already runs: same contiguity check, same drop of the queue up to the segment, same `contiguous: false` on the first frame past the gap. Downstream it is indistinguishable from `hold_back` frames having piled up, only sooner. Whichever of the two conditions fires first wins.
- **The deadline is measured from the data, not from the poll.** It is the earliest `RtpMeta::received` among the held frame's packets plus the bound, so it does not depend on when the application happens to call in.

### The wake-up

A stream that stops mid-loss delivers no more packets, so the buffer cannot learn that its deadline has passed from ingress alone. The deadline is surfaced through the existing `Session::poll_timeout()` as a new `Reason::ReorderWait`, and `Session::handle_timeout()` drives the buffers forward. Without this the bound would be correct in principle and dead in practice, since nothing would apply it until the next packet arrived — which is the exact case it exists for.

### Configuration

```rust
Rtc::builder().set_reordering_max_wait(Some(Duration::from_millis(300)))
```

`reordering_max_wait: Option<Duration>` sits next to `reordering_size_audio`/`reordering_size_video`, with the same builder-and-getter shape. It defaults to `None`, which derives the bound per receive stream from what that stream knows about the path:

```
(rtt + 3 * jitter) * 2, clamped to 100ms..=1s
```

where the RTT term is the measured RTT if we NACK, an assumed 100ms if we NACK but have measured nothing yet, and zero if we do not NACK at all (nothing is going to retransmit the packet, so only the jitter is worth waiting out). Before there is any estimate, a fixed 200ms. The derived value follows the moving estimates rather than being frozen when the buffer was created.

**The constants are a proposal, not a claim.** 200ms, the assumed 100ms RTT, the 3x jitter and 2x headroom factors and the 100ms..1s clamp are my guesses at a sane shape; they are named constants in `streams/receive.rs` and `packet/buffer_rx.rs`, so please tune them if your intuition differs. I would also happily change the config to a plain `Duration` with a documented default if you would rather not have the derived mode at all.

### Unchanged

`reordering_size_audio`/`reordering_size_video` behave exactly as before, as does a gap that closes on its own (late arrival, RTX, RED). Ordering, the codec contiguity rules, padding handling and the depack cache are untouched. The setting has no effect in RTP mode, where there are no depayloaders. A frame that the codec's contiguity rules refuse to emit is dropped and the buffer immediately looks at what is behind it, rather than waiting for the next packet.

### Tests

`tests/reorder_giveup.rs` feeds the VP8 trace with one packet dropped and never retransmitted, leaving fewer frames behind the gap than `reordering_size_video`, then drives the receiver alone through `poll_output`/`poll_timeout` with no other clock running:

- the frames behind the gap are flushed within the derived bound, in order, first one `contiguous: false` and the next `true`, and the dropped frame's packets never appear;
- they are *not* flushed before the bound, and the same idle receiver flushes them once it passes;
- with `set_reordering_max_wait(Some(600ms))` the emission time follows the configuration rather than the default;
- the getter round-trips the setter.

A lossless run of the same trace is unaffected. The existing suite (1019 tests), `cargo fmt --all -- --check`, `cargo clippy -- -D warnings` and `cargo doc` are green.


algesten:
I think we should focus on 1041. I see it like this:

* Either we do the simplest possible fix #1041
* Or we go all in on libwebrtc algorithm

This PR introduces some other algorithm that is not established.

sawy3r:
No worries mate, makes sense, #1041 is the simpler approach and the derived default here was speculative. Happy to close this one out. The same unbounded wait exists on audio when NACK is off, which #1041 leaves as is. Glad to file that separately once #1041 lands if it's useful?

algesten:
Hm. Maybe we should generalize #1041 to include audio? Let's discuss on PR.

PR #1044, opened 2026-09-18
Bound the depacketizing packet queue

algesten:
Incomplete video frames can accumulate forever without reaching the frame-count limit. Cap each depacketizer at 4,096 packets and discard the oldest packet when a newer one needs room. Eviction clears the depacketization cache and preserves the gap indication.

The cap applies on insertion, even without polling. Frames requiring more than 4,096 packets cannot be assembled.

The regression fails before the fix. All 1,017 tests pass, with 2 ignored. Formatting and Clippy pass too.

Found in #1041.


algesten:
Independent review-cycle round 1 on 57d75410008241ca8971dce47c7b266cd1706e28: no actionable findings. Reviewed the complete cumulative diff and surrounding insertion, segment caching, timeout, padding, codec contiguity, and media depayload paths. The cap is enforced before insertion, duplicate/older arrivals cannot evict retained packets, and eviction invalidates cached depacketization while the existing segment-dirty path rebuilds indices. Existing overflow, recovery, cache/gap, and duplicate regression coverage matches the intended behavior. Validation: cargo test -p str0m packet::buffer_rx::test --lib passed all 34 tests. No new regression tests or commits were needed.

